PR #1, opened 2021-05-05
Add LICENSE/NOTICE/asf.yml files

stevedlawrence:


mbeckerle (approved):
+1

.asf.yaml
```diff
@@ -0,0 +1,30 @@
...
+  description: "Apache Daffodil VSCode"
+  homepage: https://daffodil.apache.org/
+  features:
+    wiki:     false
+    issues:   false
+    projects: false
+  enabled_merge_buttons:
+    squash:   false
+    merge:    false
+    rebase:   true
```
tuxji, line 26: Let me ask a question.  I've tried the squash button on another repository and it works as well as you would expect.  GitHub squashes the commits and shows you a new commit message composed from the original commit messages so you can edit the squashed commit message too.  Since we want commits squashed when there are multiple commits, why do we enable the rebase button instead of enabling the squash button?  I would think you'd still have the ability to do a manual rebase/squash in your checkout and force push the squashed commit with either button, but the squash button allows that manual step to be skipped if all you need to do in your next step to edit the squashed commit's message before merging the squashed commit.  However, I know two good reasons for the manual step might be that if we made the squash too easy, a developer 1) might not develop the necessary skills, or 2)  might forget to edit the composed message, meaning that the squashed commit's message still might have all the original commits' messages in it.  Is your thinking that the squash step should remain manual in order to ensure one's skills are well developed and one edits the message more carefully?

tuxji (approved):
+1

.asf.yaml
stevedlawrence: The main reason I don't like the squash merge is because last I checked, it creates a merge commit if the PR isn't directly on top master. I try to avoid merge commits as much as possible.

My other issue with squash is that whoever created the PR probably has a better idea what a good squashed commit message should be, rather than whoever happens to be merging the PR. I'd rather the author fixup the message first, and the the merger doesn't have to do anything.

So it's not so much about forcing others to learn the workflow, but more I like to avoid merge commits, and the author of the commit knows best what a squash commit should be, rather than the merger.

Issue #7, opened 2021-10-08
Publish to marketplace


jw3:
Automated publishing of releases to VS Code marketplace.

Normally you would configure the publishing through https://marketplace.visualstudio.com/manage but this depends on the ASF procedures for publishing.

Blocked by #3 which updates the publisher and other details for the Apache project.

Original issue https://github.com/jw3/example-daffodil-vscode/issues/26

stevedlawrence:
Here's an INFRA issue about other Apache Netbeans publishing to the marketplace:

https://issues.apache.org/jira/browse/INFRA-21017

So `asf` probably is the correct publisher ID, though it sounds like we'll need to open infra tickets for individual Daffodil devs to get contributor access to publish to the marketplace. Once users get that access, we'll want to document the process to publish packages in the release workflow page.

Shanedell:
@stevedlawrence I have published the extension to this VS Code Marketplace so this should be closed correct? (https://marketplace.visualstudio.com/items?itemName=ASF.apache-daffodil-vscode)

stevedlawrence:
Any idea why the breadcrumb at the top of that page says "Snippets"?

Can you update the [Release Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Release+Workflow) "Promoting to Final Release" section to include detailed instructions on how to get access and upload the vsix artifact.

Maybe we should also add a tab next to "Download" (similar to Daffodil's Maven/SBT/DNF tabs) that describes how to install the extension from withing VS Code? Maybe something similar to what the market place describes in the "Installation" section (i.e. Open VS Code, press Ctrl-P, and run "ext install ASF.apache-daffodil-vscode")

Also, I've added a tweet for this on the @ApacheDaffodil twitter.

Shanedell:
@stevedlawrence Would this be okay to close as I updated the [Release Workflow](https://cwiki.apache.org/confluence/display/DAFFODIL/Release+Workflow) with a new subsection for publishing the VS Code extension.

stevedlawrence:
Yep, I think this is good to close. I've also merge https://github.com/apache/daffodil-site/pull/93 to add marketplace instructions.

PR #22, opened 2021-10-11
Update old references and package.json

### LINKED ISSUE #3: Update package.json
Take a pass over `package.json` and update for the new location.

Updates should include, but may not be limited to, the following
- publisher
- repository.url
- bugs.url

### LINKED ISSUE #4: Update references to the old repository
- [nightly.yml](https://github.com/jw3/example-daffodil-vscode/blob/091ad2314af18a81633ef66be996871e20d68a9a/.github/workflows/nightly.yml#L58)
- [README.md](https://github.com/jw3/example-daffodil-vscode/blob/535863991607daf81e5b9372b0b36dbdacdfb21f/README.md#L32)
- [daffodilDebugger.ts#23](https://github.com/jw3/example-daffodil-vscode/blob/8778f5d1403c365419613317eee1a50d980aa1f6/src/daffodilDebugger.ts#L23)
- [daffodilDebugger.ts#48](https://github.com/jw3/example-daffodil-vscode/blob/091ad2314af18a81633ef66be996871e20d68a9a/src/daffodilDebugger.ts#L48)
- [DAPodil.scala](https://github.com/jw3/example-daffodil-vscode/blob/091ad2314af18a81633ef66be996871e20d68a9a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala#L754)
- #3

jw3:
Initial updates for all jw3 references and some general cleanup on the `package.json`
- update links and code referring to jw3 repo
- update `package.json` for Apache publishing
- general cleanup on `package.json` based on the example of other ASF VS Code extensions
- removes nightly build

Referenced the following Apache projects with VS Code extensions
- Apache Openwhisk - https://github.com/apache/openwhisk-vscode-extension
- Netbeans LSP server - https://github.com/apache/netbeans/tree/master/java/java.lsp.server/vscode

Also reference the VS Code `package.json` extension manifest docs
- https://code.visualstudio.com/api/references/extension-manifest

closes #3
closes #4


jw3:
The `package.json` updates here should set #7 up to move forward.

tuxji (approved):
+1

Found one jw3 left, but it's a comment with an url to an old github issue that probably should stay in that repo.

<https://github.com/apache/daffodil-vscode/blob/091ad2314af18a81633ef66be996871e20d68a9a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala#L754>

README.md
```diff
@@ -26,11 +26,11 @@ This is a VS Code extension which enables the interactive debugging of DFDL sche
...
-Until the extension is available in the [VS Code Extension Marketplace](https://marketplace.visualstudio.com/vscode), please download the latest `.vsix` file from the [releases page](https://github.com/jw3/example-daffodil-vscode/releases). Then install it by either:
+Until the extension is available in the [VS Code Extension Marketplace](https://marketplace.visualstudio.com/vscode), please download the latest `.vsix` file from the [releases page](https://github.com/apache/daffodil-vscode/releases). Then install it by either:
   * using the "Extensions: Install from VSIX" command from within VS Code (open the Command Palette with Shift-Command-P, then type `vsix` to bring up the command and pointing it at the downloaded `vsix` file); or
   * on the command-line via `code --install-extension <path-to-downloaded-vsix>`.
 
 
 ## Usage
 
-Please refer to the Wiki page at https://github.com/jw3/example-daffodil-vscode/wiki
+Please refer to the Wiki page at https://github.com/apache/daffodil-vscode/wiki
```
stevedlawrence, line 36: No where in the README are there build dependencies or instructions. Apache is very source code focused, so instructions should first and foremost be about how to build things from source. It's fine to also provide convenience binaries, whether they be from downloads.apache.org or via the marketplace, but that's importtation information that should be in this README. It doesn't need to be part of this PR, but a new issue should be opened to add that information.

package.json
```diff
@@ -1,13 +1,11 @@
 {
   "name": "daffodil-debugger",
-  "displayName": "Daffodil Debugger",
+  "displayName": "Apache Daffodil Debugger Extension",
+  "description": "VS Code extension for Apache Daffodil Schema Debugger",
```
stevedlawrence, line 4: Seems `VSCode` without the space is the common way to refer to Visual Studio Code? I would also refer it as ` `DFDL Schema Debugger`, just to make it clear that these are Apache Daffodil schemas, they are DFDL Schemas, and the ebugger is powered by Apache Daffodil.

package.json
```diff
@@ -1,13 +1,11 @@
 {
   "name": "daffodil-debugger",
-  "displayName": "Daffodil Debugger",
+  "displayName": "Apache Daffodil Debugger Extension",
```
stevedlawrence, line 3: Could the name be changed to make it more clear this is about debugging schemas? With this name, it sounds like it's used for debugging Apache Daffodil, which isn't the main focus of the tool. Maybe something like "DFDL Schema Debugger Extension" or "Apache Daffodil DFDL Schema Debugger Extension", though that's a bit long. Maybe this is fine and the description can be used to clarify it's for schema debugging?

package.json
```diff
@@ -1,13 +1,11 @@
...
+  "description": "VS Code extension for Apache Daffodil Schema Debugger",
   "version": "1.0.0",
-  "publisher": "jw3",
-  "description": "Daffodil Schema Debugger: debug DFDL schema files using Apache Daffodil.",
+  "publisher": "asf",
+  "author": "Apache Daffodil",
   "license": "Apache-2.0",
```
stevedlawrence, line 8: I usually use `Apache License, Version 2.0`, since this is what https://www.apache.org/licenses/LICENSE-2.0 uses.

stevedlawrence (approved):
+1 We should update the remaining jw3 reference that @tuxji found to refer to a bug in the new issues tracker. That bug can refer to the old bug if needed.

package.json
jw3: > Seems VSCode without the space is the common way to refer to Visual Studio Code

You see it both ways in the wild.  I was looking at Openwhisk json when updating this.  Want it changed?

>  would also refer it as DFDL Schema Debugger`

OK

package.json
stevedlawrence: I wasn't aware of spdx, good to know. Apache-2.0 sounds good then.

package.json
stevedlawrence: You're right, the first couple things I saw were VSCode, but I see both uses, even in official vscode documentation. This is fine.

jw3: Yeah I just did a search and we have a mix in this repo, but definitely more with the space.

I added #24 to get this project to a standardization.

PR #26, opened 2021-10-12
View Launch Options

### LINKED ISSUE #13: Layout preferences
The preferred layout is going to vary by user. Currently we have a sparse initial layout that only has the schema opened automatically. The user can use the command menu to open the

    data stream (hex) view
    infoset view
    infoset diff view

We could add/preserve these in the users configuration to enable their preferred/previous config to be loaded automatically.

Original issue https://github.com/jw3/example-daffodil-vscode/issues/91

Shanedell:
Added three options for opening up the different views automatically on debug start

New Launch Options:
- `openHexView` -- When set to `true` tells the extension to open up the Hex View on debug start
- `openInfosetView` -- When set to `true` tells the extension to open up the Infoset View on debug start 
- `openInfosetDiffView` -- When set to `true` tells the extension to open up the Infoset Diff View on debug start

Closes #13


mbeckerle (approved):
+1

stevedlawrence:
+1, but I'd prefer to see a +1 from @jw3 before this is merged

.gitignore
```diff
@@ -11,6 +11,10 @@ node_modules
 target
 .bsp
 
+# npm/yarn install files
```
jw3, line 14: We want to lock the yarn deps, so you can remove that ignore below.  Should update this comment to indicate that we only use yarn for package management, so the npm lock is ignored here.

The npm lock shouldn't be generated given our dev docs, but just in case someone does we will ignore it.

src/hexView.ts
```diff
@@ -120,6 +121,19 @@ export class DebuggerHexView {
           this.onDisplayHex(e.session, e.body)
           break
       }
+
+      let hexFileOpened = false
+
+      for (var i = 0; i < vscode.window.visibleTextEditors.length; i++) {
```
jw3, line 127: Prefer a for-of loop here

yarn.lock
```diff
@@ -0,0 +1,4249 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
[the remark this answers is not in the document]
jw3: How do we add license header for autogenerated files?

jw3:
> Looks like there's hundreds of dependencies if you include the transitivies.

Well you would probably only update a top level dependency, not a transient, otherwise you might create undefined behavior the whole way down by changing someone else's dep.

> There's an easier way - let dependabot do some of the work. 

Yes, concur.

stevedlawrence:
We try to avoid merge commits in pull requests. They make it difficult to reason about history and make some other things like git bisect a bit more difficult. Please squash this all into a single commit before this can be merged.

Issue #19, opened 2021-10-11
Update Wiki images for the new repo


jw3:
The wiki links image from issues.  The issue that holds the images is in the old repo.  Need to update the wiki in this repo with appropriate links to self hosted images.

PR #31, opened 2021-10-14
Toolbar Icons

### LINKED ISSUE #15: Daffodil toolbar
Add a toolbar with buttons to improve usability by providing an alternative to using the command palette. 

This should contains all of the available Daffodil commands, where the stateful commands are represented using toggle buttons.

Original issue https://github.com/jw3/example-daffodil-vscode/issues/123

Shanedell:
Add toolbar icons for:
- Infoset View
  daffodil-infoset-view-icon

- Infoset Diff View
  daffodil-infoset-diff-view-icon

- Hex View
  daffodil-hex-view-icon

Closes #15

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,23 @@
       }
     ],
     "menus": {
+      "editor/title": [
+        {
+          "command": "hexview.display",
+          "when": "resourceLangId == xml",
+          "group": "navigation@3"
+        },
+        {
+          "command": "infoset.display",
+          "when": "resourceLangId == xml",
+          "group": "navigation@1"
+        },
+        {
+          "command": "infoset.diff",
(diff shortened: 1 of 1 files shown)


Shanedell:
@stevedlawrence 
So the icons are baked in VSCode. Here is the list of them https://code.visualstudio.com/api/references/icons-in-labels

I can definitely make infoset be file-code instead. Maybe look at the icon list above and see what you like to be used for the infoset-diff, in the list there is `plus` which is just `+` if that works?

stevedlawrence:
New icons look good to me :+1:

jw3:
The images in the wiki docs will need updated, but that can be done with #19

PR #35, opened 2021-10-15
Change build settings to ensure repeatable builds

### LINKED ISSUE #34: Remove sbt-git plugin for repeatable builds
The `sbt-git` plugin is used to set `buildInfoKeys` based on the git commit hash. While this is useful information, it means a build from a git repo and a build from a source tarball are different. Repeatable builds should be possible regardless of where the build is done. Removing the sbt-git plugin ensures there is no git-specific information in builds.

stevedlawrence:
- Remove git information, so builds from within a git repo are the same
  as those from a source tarball
- Remove build time from BuildInfo

Closes #34

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,7 +18,14 @@ lazy val daffodilVer = "3.1.0"
 
 lazy val commonSettings = {
   Seq(
-    git.useGitDescribe := true,
+    version := {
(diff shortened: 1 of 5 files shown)


mbeckerle (approved):
+1

PR #37, opened 2021-10-18
Add Build Instructions to README.md

### LINKED ISSUE #23: Add build instructions to the readme
No where in the README are there build dependencies or instructions. Apache is very source code focused, so instructions should first and foremost be about how to build things from source. It's fine to also provide convenience binaries, whether they be from downloads.apache.org or via the marketplace, but that's importtation information that should be in this README. It doesn't need to be part of this PR, but a new issue should be opened to add that information.

_Originally posted by @stevedlawrence in https://github.com/apache/daffodil-vscode/pull/22#discussion_r727022373_

Shanedell:
Changes:
- Added instructions of how to build from source to the `README.md`

To view changes:
- https://github.com/Shanedell/daffodil-vscode/tree/issue_23_build_in_readme

Closes #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,14 +21,76 @@
 
 This is a VS Code extension which enables the interactive debugging of DFDL schema parsing using [Apache Daffodil](https://daffodil.apache.org/).
 
+## Build Requirements
+- Java Development Kit (JDK) 11 or higher
+- SBT 0.13.8 or higher
+- Node 10 or higher
+- Yarn (https://yarnpkg.com/getting-started/install)
+
+## Download - Prebuilt
+
+Until the extension is available in the [VS Code Extension Marketplace](https://marketplace.visualstudio.com/vscode), please download the latest `.vsix` file from the [releases page](https://github.com/apache/daffodil-vscode/releases).
+
+
+## Build VSIX and Debugger
+:exclamation:**NOT necessary if using prebuilt VSIX**:exclamation:
+
+:exclamation:**NOT necessary if running extension via VSCode without VSIX but a `yarn install` will be required**:exclamation:
+
+Run full build
+  ```bash
+  yarn build
+  ```
+
+* This command performs the following tasks:
+  * Create sbt zip package
+  * Install dependencies
+  * Compiles Extension
+  * Packages Extension
+  * Creates Files
+    ```
+    server/core/target/universal/daffodil-debugger-*.zip
+    daffodil-debugger-*.vsix
+    ```
+
+## Running Debug Server
+
+The debug server will automatically be ran by the extension unless `useExistingServer` is to set to `true` inside of `.vscode/launch.json`
+
+If you wish to ran the debug server manually the scripts to do so are at the following locations:
+* Debugging through VSCode with or without VSIX:
+  * Linux = `/home/USERNAME/.local/share/daffodil-dap`
+  * Mac = `/Users/USERNAME/Library/Application\ Support/daffodil-dap`
+  * Windows = `C:\\Users\\USERNAME\\AppData\\Roaming\\daffodil-dap`
+  
+
 
 ## Installation
 
-Until the extension is available in the [VS Code Extension Marketplace](https://marketplace.visualstudio.com/vscode), please download the latest `.vsix` file from the [releases page](https://github.com/apache/daffodil-vscode/releases). Then install it by either:
-  * using the "Extensions: Install from VSIX" command from within VS Code (open the Command Palette with Shift-Command-P, then type `vsix` to bring up the command and pointing it at the downloaded `vsix` file); or
-  * on the command-line via `code --install-extension <path-to-downloaded-vsix>`.
+Once you have either downloaded the VSIX file or created it you can now install it. There are two options to do this:
 
+* Via "Extensions: Install from VSIX" command from within VS Code.
+  * Open the Command Palette with Shift-Command-P (Mac) OR Shift-Ctrl-P (Windows/Linux)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,14 +23,79 @@
...
+  ```
+
+* Package
+  ```bash
+  yarn package
+  ```
+
+  * Creates file `daffodil-debugger-${package.json_version}.vsix`
+
+* Install VSIX file
```
stevedlawrence, line 94: I think this belongs in an Install section.

README.md
```diff
@@ -23,14 +23,79 @@
...
+  ```bash
+  yarn compile
+  ```
+
+* Package
+  ```bash
+  yarn package
+  ```
+
+  * Creates file `daffodil-debugger-${package.json_version}.vsix`
```
stevedlawrence, line 92: This is minor, but the heavy use of lists and nesting I find a bit hard to follow. I would suggest using sections and subsections rather than lists and sublists.

README.md
```diff
@@ -23,14 +23,78 @@
...
+## Installation - via release
 
 Until the extension is available in the [VS Code Extension Marketplace](https://marketplace.visualstudio.com/vscode), please download the latest `.vsix` file from the [releases page](https://github.com/apache/daffodil-vscode/releases). Then install it by either:
   * using the "Extensions: Install from VSIX" command from within VS Code (open the Command Palette with Shift-Command-P, then type `vsix` to bring up the command and pointing it at the downloaded `vsix` file); or
   * on the command-line via `code --install-extension <path-to-downloaded-vsix>`.
 
 
+## Installation - via source
+
+:exclamation: Currently only Daffodil 3.1.0 is support, will be updated in the future :exclamation:
```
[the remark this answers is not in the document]
Shanedell: Probably, will delete

README.md
```diff
@@ -23,14 +23,79 @@
 
 This is a VS Code extension which enables the interactive debugging of DFDL schema parsing using [Apache Daffodil](https://daffodil.apache.org/).
 
+## Build Requirements
+- SDK 8 or higher
+- SBT 0.13.8 or higher
+- Node 10 or higher
```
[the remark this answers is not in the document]
Shanedell: yeah you both are correct I will fix these issues

package.json
```diff
@@ -31,7 +31,9 @@
     "watch2": "tsc -watch -p ./",
     "package": "vsce package",
     "publish": "vsce publish",
-    "package-ext": "webpack --mode production --config ./build/extension.webpack.config.js"
+    "package-ext": "webpack --mode production --config ./build/extension.webpack.config.js",
+    "sbt": "sbt universal:packageBin",
+    "build": "yarn sbt && yarn install && yarn compile && yarn package"
```
stevedlawrence, line 36: Cool!

Just a question about yarn, is this standard to have a yarn script run other yarn scripts like this? Thinking how sbt works, you just specify command dependencies. So you could specify that `yarn package` depends on `yarn compile`, which depends on `yarn install`. That way you can tell people to just run `yarn package` and if they haven't run the previous commands, they would get run because they are dependent commands.

But maybe yarn doesn't work like that, and this `yarn build` is the standard way to do this kind of thing?

Shanedell: From what I saw it seems the only way to do is the way I have implemented. Yarn doesn't allow you to specify other scripts a script depends on so that they will run prior.

I believe the only way it functions this way is if you have something like `build` and `prebuild` then run `yarn run build` I think it will automatically run prebuild first but not completely sure on that.

jw3:
What is the status of this?

README.md
```diff
@@ -21,14 +21,81 @@
...
+## License
+
+Apache Daffodil VSCode is licensed under the [Apache License, v2.0].
+
+[Apache License, v2.0]: https://www.apache.org/licenses/LICENSE-2.0
+[Apache RAT]: https://creadur.apache.org/rat/
+[Daffodil JIRA]: https://issues.apache.org/jira/projects/DAFFODIL/
+[Github Actions]: https://github.com/apache/daffodil-vscode/actions?query=branch%3Amain+
+[Releases]: https://github.com/apache/daffodil-vscode/releases
+[SBT]: https://www.scala-sbt.org/
```
stevedlawrence, line 100: None of these link references are used, except for the Apache license. All except that one can be removed.

README.md
```diff
@@ -21,13 +21,65 @@
...
+* Via "Extensions: Install from VSIX" command from within VS Code.
+  * Open the Command Palette with Shift-Command-P (Mac) OR Shift-Ctrl-P (Windows/Linux)
+  * Type `vsix` to bring up the command and pointing it at the `vsix` file
 
+* Via command line
+  ```bash
+  code --install-extension daffodil-debugger-*.vsix
+  ```
 
 ## Usage
```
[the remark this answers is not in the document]
stevedlawrence: JIRA is used just for Daffodil. We are using the GitHub issues for the vscode extension.

stevedlawrence (approved):
+1 once the links are fixed up

README.md
```diff
@@ -21,14 +21,75 @@
...
 
 Please refer to the Wiki page at https://github.com/apache/daffodil-vscode/wiki
+
+## Getting Help
+
+You can ask questions on the dev@daffodil.apache.org or
+users@daffodil.apache.org mailing lists.  You can report bugs via the
+[Daffodil JIRA].
```
jw3, line 89: This should be GitHub Issues not Daffodil JIRA, with link to the issues page.

PR #49, opened 2021-11-12
Launch Config Wizard

### LINKED ISSUE #38: eliminate launch.json configuration steps
We need a "new..." wizard that creates these launch configurations from a dialog. 

The dialog should have a place for users to enter everything they have to customize in the launch.json today.

Shanedell:
Create launch.json config wizard:

- Create WebView for selecting `launch.config` options
- Create new command called `launch.config`
  - When command is ran the user will go through a selection process via custom webivew.
  - Once the save button is clicked the config will be saved to the `launch.json`, both if creating the file or updating it
  - After the file is created or update it is automatically opened
- Add an icon for running the new command to the toolbar (gear icon).
- Create Global settings that be access and edited via the Settings page in vscode
- Update code to grab global defaults rather than reassigning them multiple times

NOTE: Currently I have made the wizard now allow EXACT duplicate configs into the `launch.json`, however would it be better to just go based of names such as if `Wizard Config` is trying to be saved but it already exists in `launch.json` should it be skipped even if some of the variables do not match?

Here is what the icon in the toolbar looks like

Screen Shot 2021-11-11 at 7 53 16 PM

Here is what the WebView looks like

Screen Shot 2021-11-22 at 1 22 00 PM

Closes #38

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -19,6 +19,8 @@ import { getDebugger, getDataFileFromFolder } from '../daffodilDebugger'
 import { FileAccessor } from './daffodilRuntime'
 import * as fs from 'fs'
 import * as infoset from '../infoset'
+import { getConfig } from '../utils'
+import * as launchWizard from '../launchWizard/launchWizard'
 
 // Function for setting up the commands for Run and Debug file
 function createDebugRunFileConfigs(resource: vscode.Uri, runOrDebug: String) {
@@ -32,21 +34,19 @@ function createDebugRunFileConfigs(resource: vscode.Uri, runOrDebug: String) {
     let infosetFile = `${
       path.basename(targetResource.fsPath).split('.')[0]
     }-infoset.xml`
+    vscode.window.showInformationMessage(infosetFile)
 
     vscode.debug.startDebugging(
       undefined,
-      {
-        type: 'dfdl',
-        name: 'Run File',
-        request: 'launch',
-        program: targetResource.fsPath,
-        data: '${command:AskForDataName}',
-        debugServer: 4711,
-        infosetOutput: {
-          type: 'file',
-          path: infosetFile,
-        },
-      },
+      getConfig(
+        'Run File',
+        'launch',
+        'dfdl',
+        targetResource.fsPath,
+        false,
+        false,
+        { type: 'file', path: '${workspaceFolder}/' + infosetFile }
+      ),
       { noDebug: noDebug }
     )
 
@@ -160,18 +160,15 @@ export function activateDaffodilDebug(
         ): ProviderResult<DebugConfiguration[]> {
           if (!vscode.workspace.workspaceFolders) {
             return [
-              {
-                name: 'Daffodil Launch',
-                request: 'launch',
-                type: 'dfdl',
-                program: '${file}',
-                data: '${command:AskForDataName}',
-                debugServer: 4711,
-                infosetOutput: {
-                  type: 'file',
-                  path: '${file}-infoset.xml',
-                },
-              },
+              getConfig(
+                'Daffodil Launch',
+                'launch',
+                'dfdl',
+                '${file}',
+                false,
+                false,
+                { type: 'file', path: '${file}-infoset.xml' }
+              ),
             ]
           }
 
@@ -183,18 +180,15 @@ export function activateDaffodilDebug(
           }-infoset.xml`
 
           return [
-            {
-              name: 'Daffodil Launch',
-              request: 'launch',
-              type: 'dfdl',
-              program: '${file}',
-              data: '${command:AskForDataName}',
-              debugServer: 4711,
-              infosetOutput: {
-                type: 'file',
-                path: infosetFile,
-              },
-            },
+            getConfig(
+              'Daffodil Launch',
+              'launch',
+              'dfdl',
+              '${file}',
+              false,
+              false,
+              { type: 'file', path: '${workspaceFolder}/' + infosetFile }
+            ),
           ]
         },
       },
@@ -269,6 +263,7 @@ export function activateDaffodilDebug(
   )
 
   infoset.activate(context)
+  launchWizard.activate(context)
 }
 
 class DaffodilConfigurationProvider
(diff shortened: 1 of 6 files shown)


jw3:
> ... and text fields

Nothing special in mind here, just a file picker with path display.

Shanedell:
@jw3 Is there a specific variable you are wanting this for?

As the `path` and `data` variables are already files pickers, I believe these do not show you the full path since you use the operating systems file explorer to chose the file. 

I might not be understanding what you may be asking to be done

jw3:
Usually when you click on a settings button you get a panel with settings, then you click ok when done.

The way it is implemented now will probably be surprising to most users.

Shanedell:
Oh I think I see what you are saying.

So a panel that will:
- Show all the different variables that can be set, with their default value in parenthesis like `debugServer (default: 4711)`
- When the user clicks on one of the panel options it will bring up the proper input type to select the item
- Once input is retrieved the panel will be updated with that value
- Once the user clicks `OK` or `Save` on the panel it will create the `.vscode/launch.json`

Was this similar to what you were thinking @jw3 ?

jw3:
> When the user clicks on one of the panel options it will bring up the proper input type to select the item

I'm not sure about this statement.  I am just thinking of standard UI components to configure each setting.

For an example in VS Code go to File -> Preferences -> Settings

Shanedell:
Oh okay so you mean making a new item in the `Extensions` dropdown for `Daffodil Debugger` where you can set all the properties and it doesn't need to create a new folder and file inside of the workspace it just uses the global settings file?

jw3:
Yeah I think that is pretty close.

Shanedell:
@mbeckerle @stevedlawrence @jw3 

So for this PR I am thinking we keep it where is at now which is making a entry inside of the Settings of Visual Studio Code and pull down the default values from that config if the file `.vscode/launch.json` is not available. I think this will work better instead of remaking those default values inside of the code in different areas.

Then after this one is merged I have an idea of how to GUI wizard for making the launch.json using a WebView panel. 

So if this works for you guys I can remove the closing of 38 from the description and work on GUI wizard in a different PR.

jw3:
> making a entry inside of the Settings of Visual Studio Code and pull down the default values from that config if the file .vscode/launch.json is not available

That sounds good. So we have these options

1. A `launch.json` is not required, the default launch config is available from the settings page
2. A `launch.json` can be created in which case it will be detected and used

This seems to take care of the simple case where a user just wants to run something the simplest way possible while allowing for the alternative case where someone wants multiple launch profiles with different options.

> Then after this one is merged I have an idea of how to make a GUI wizard for making the launch.json using a WebView panel.

Sounds good.  I think it will be useful to see a UI tailored for this app and determine if that is a better fit than the settings page.

Shanedell:
@jw3 I do have a rough draft of a UI for the config wizard that I made based on this branch, I will post it here for comments about some items that may need changed
Screen Shot 2021-11-19 at 1 56 56 PM

jw3:
Nice :+1:   So what's it look like when you have multiple launch configurations?

Shanedell:
So basically my idea was that if the save button is clicked if the `.vscode/lauch.json` does not exist it will be created. But if it does exist we will read in the file then append the new configuration to the list of configs. Once the list of configs is updated write the update back to the file, does that make sense?

jw3:
Yeah that all makes sense.  How far away is the rough draft from being ready for PR?

Shanedell:
I mostly need to do the updating of the launch.json and it will be good to go

Shanedell:
@jw3 I have gotten the GUI working good and updating the configurations is working as well now. I also made it so that an exact duplicate config will not be added to the `launch.json`. When the config being saved is already in the `launch.config` the user will get an a warning message that says `The conf trying to be saved already exists in the launch.json`.

Do you think I should add this code to this PR as I based it off this branch or wait till this PR is merged to push up the code?

mbeckerle:
My thought is that you should add it to the existing PR.

jw3:
Agree, this PR

Shanedell:
@mbeckerle @jw3 I have added the code for the GUI and updated the PR description to accommodate the changes I have made.

jw3:
@Shanedell  The GUI looks good but I think needs something for handling multiple configurations, it's going to be confusing otherwise.

Is there any way to add a simple list component to the editing screen, that lists all the configs from the launch.json and when a selection is made it changes what configuration is loaded in the editing view?

Shanedell:
@jw3 Yeah I think I could add something like a button that says `Edit Config` when selected instead of asking for a name it will be a dropdown will the configs listed in it?

jw3:
The drop down sounds like a good approach.

Shanedell:
@jw3 Which would be preferred one or two?

1.) Same layout for `Name` as before but whenever the dropdown selection is not `New Config` the input box is invisible, when selected will appears.

Not Selected:
Screen Shot 2021-11-30 at 11 44 59 AM

Selected:
Screen Shot 2021-11-30 at 11 45 08 AM


2.) When `New Config` is selected a label with text box beside it appears?

Not Selected:
Screen Shot 2021-11-30 at 11 53 17 AM

Selected:
Screen Shot 2021-11-30 at 11 53 23 AM

jw3:
I like 2

Shanedell:
From the most recent commit it utilized option 2, if 1 is desired more I can easily make that update

jw3:
This looks close but the fields do not update when I select different values in from the drop down.

For example one config has `123` as the port another has `456`, when I select either of these the port remains `4711`.

Also for the infoset output type there needs to be a file path field when the `file` type is selected.

Would it be possible for all file path fields to have a file picker button that launches a file chooser?

Shanedell:
Oh yeah I don't have the settings page update fields based on the selected config, that makes a lot of sense though I will fix that.

So instead of the input box able to be edited have a button beside it like `Browse` that opens a file browser and that updates the input box value with the path to file selected?

jw3:
I think you could leave the field be editable, but if you use the picker then it overwrites whatever was there. It is fine if someone wants to edit it after that, or instead of that.

Shanedell:
@jw3 

- Updated the webview to update values based on config selected
- Updated the way the input for the infoset output path shows
- Created a `Browse` button for the `Program` and `Data`
  - When clicked will open a file picker and once file is selected the input box is updated
  - Did not do this for others as they either weren't asking for a file or in the case of the path for infoset output I did as I assumed that may not exist.

mbeckerle (approved):
+1 I am still learning typescript, so not a good code reviewer as yet, but I was able to checkout this PR, build and try it out and this functionality is a big improvement. 

I submitted some separate tickets associated with further enhancements needed (for classpath, etc.)

src/adapter/activateDaffodilDebug.ts
```diff
@@ -291,21 +286,29 @@ class DaffodilConfigurationProvider
...
-        config.stopOnEntry = true
-        config.useExistingServer = false
-        config.infosetOutput = {
-          type: 'file',
-          path: '${file}-infoset.xml',
-        }
-        config.debugServer = 4711
-      }
+      config = getConfig('Launch', 'launch', 'dfdl')
+      // config.type = 'dfdl'
```
jw3, line 290: can this commented out code be removed?

jw3 (approved):

jw3:
+1

src/adapter/activateDaffodilDebug.ts
Shanedell: yes I completely forgot about it will remove it now

PR #53, opened 2021-11-17
Remove create_vsix.sh in favor of yarn build

### LINKED ISSUE #52: daffodil-debugger zip is not included in .vsix file when built with create_vsix.sh
Running `./create_vsix.sh` from a clean repository does not include the daffodil-debugger zip file in the built .vsix file.

I think `yarn build` is now the indented way to create the .vsix file, which does build the zip file and include it correctly.

Is `create_vsix.sh` old and should just be removed in favor of `yarn build`?

stevedlawrence:
No build or README file mentions create_vsix.sh as the way to build a
.vsix file. Instead we only ever mention yarn build. And create_vsix.sh
doesn't build/include the daffodil-debugger-*.zip file, so it isn't even
a correct .vsix file. Remove it in favor of yarn build.

Closes #52


Shanedell (approved):
+1 Good find, `create_vsix.sh` was used for creating specifically only the VSIX file however with the inclusion of `yarn build`, which builds both the debugger and VSIX, this script would no longer be needed

tuxji (approved):
+1

jw3 (approved):
+1

PR #60, opened 2021-12-06
Infoset File Written Popup

### LINKED ISSUE #12: Popup notification when infoset is written
After writing a file a notification should pop up that shows the output path and allows the user to conveniently open it.

An example of the thought here can be seen when using the Chronicler extension

Original issue https://github.com/jw3/example-daffodil-vscode/issues/67

Shanedell:
Updates:

 - Create popup notification when infoset is written.
 - Create code to set a variable that can be shared amongst files to get the current config being debugged
   - This helps to support multiple different configs as it makes sure the code is working on the config being used not just the first config in the `launch.json` file
 
What the popup notification looks like:

Screen Shot 2021-12-06 at 2 03 33 PM

When `Open` is clicked the infoset file will be become the active opened file

Closes #12

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -19,7 +19,7 @@ import { getDebugger, getDataFileFromFolder } from '../daffodilDebugger'
 import { FileAccessor } from './daffodilRuntime'
 import * as fs from 'fs'
 import * as infoset from '../infoset'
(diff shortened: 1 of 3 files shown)


mbeckerle (approved):
+1 While I think popup dialogs are annoying and I'd rather see a more subtle indication that this data is written, this is better than having no detectable visible indication

jw3 (approved):
+1

PR #61, opened 2021-12-06
1 Commit and Code Formatting CI

### LINKED ISSUE #45: Check code formatting in CI
The code formatters prettier and scalafmt should be run in CI to check that the committed code is formatted, if not it should fail the build.

We had prettier integrated in the old repo

https://github.com/jw3/example-daffodil-vscode/blob/master/.github/workflows/release.yml#L61-L63

but it doesn't look like we had scalafmt checked.

### LINKED ISSUE #46: 1-commit PR CI
Add a Github action file that will check if the PR has a single commit only

Shanedell:
Updates:

- Add ScalaFMT to plugins.
- Create `.prettierignore` for files for prettier to ignore. 
- Create CI for running code linting/formatting for typescript, javascript and scala code.
- Create CI for checking PRs are 1 commit.

I also believe since we use prettier for code formatting the TypeScript and JavaScript we can remove the `.eslintignore` and `.eslintrc.json` files as well the eslint packages from package.json. Thoughts on this @jw3 @mbeckerle @stevedlawrence ?

Closes #45 
Closes #46


Shanedell:
The unit tests CI also run a `lint` before the tests do we want to remove this before testing since linting will be its own CI stage?

Shanedell:
@stevedlawrence I believe you had mentioned before about maybe not doing a different a seperate file for each CI action, instead making just one similar to how the Daffodil does. Would you want me to make an issue for once this one is merged that we work on converging into all of the seperate CI files into one `CI.yml` file?

stevedlawrence:
Please do, thanks!

Shanedell:
@stevedlawrence Sounds good. Just to let you know I believe we can converge all the files in `CI.yml` except for `release.yml`. This mostly being because it will require a different branch/tag matching than all the other CI items need.

But also what do you think about also deleting the `.eslint` files and dependencies since we are going with prettier?

stevedlawrence:
Regarding the release, I'm not sure we even want to keep release.yml. ASF is pretty strict about pre releases not being published outside of dist.apache.org while the release is voted on. And the process to make a pre-release an official release is basically a `mv` from the staging area to the release area. We may need to rethink how releases are build and published.

.github/workflows/formatting.yml
```diff
@@ -0,0 +1,54 @@
...
+    name: TypeScript Formatting
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2.3.5
+      - name: Setup Node
+        uses: actions/setup-node@v1.4.6
+        with:
+          node-version: '16'
+      - run: yarn install
+      - run: yarn lint
```
mbeckerle, line 35: How customizable is this lint tool? E.g., can we make it ignore some things that we decide are ok stylistically?

mbeckerle (approved):
+1 I see that the yarn lint is being removed, so my one posted comment is likely no longer relevant.

.github/workflows/formatting.yml
Shanedell: Well `yarn lint` is just being removed from the yarn `pre-test` script as linting on a Windows machine causes issues because of line ending differences.

Other than that `prettier` is very customizable with ignoring. If we want to ignore a entire directory, file, type of file, etc you can that to the `.prettierignore` file. Then you can ignore specific lines of file (JavaScript/TypeScript) like this:
```
// prettier-ignore
```

for other file types, eg yaml, they detail that here https://prettier.io/docs/en/ignore.html

mbeckerle:
re: VISX file. That should not be created and distributed from github. 

Build from source is the right approach for unofficial or between-releases VISX files. If the tool chain for this is too daunting, we can use the approach of giving a docker container and script to build it, but I think it was as simple as just 'yarn build'.

Shanedell:
@mbeckerle Okay in that case I will delete the release.yml from the repo.

stevedlawrence:
Note that we can distribute through github releases, but it must only be for convenience (like how the marketplace is allowed but only for convenience). "Official" releases must be made available via downloads.apache.org from our website. In general, I thin apache perfer that githug releases not be used in favor of website or something like a marketplace, because github will automatically create "release" in some cases, and it can be confusing what is actually an official release that was voted on.

The release process should be something like:
1. Create a zip of the source and a vsix file that is expected to be the final release. The .vsix is considered a convenience binary. 
2. This is called an "-rc1", but rc1 doesn't actually appear anywhere in the source or the convenience binary. Usually it's just the name of a directory in an ASF svn repo (e.g. https://dist.apache.org/repos/dist/dev/daffodil/3.2.1-rc2/) staging files for a vote.
3. We vote on releasing the source and the convenience .vsix binary
4. When the vote passes, we rename the directory to remove the -rc1 label, and then we can upload the convenience binaries where it is most convenient. For example, in Daffodil this is maven central, our webiste, and an RPM repo. For vsix ideally it would be the marketplace, and maybe github releases if it makes sense.

Shanedell:
Yeah that makes sense I don't think it is necessary to add the VSIX file to a github release if it is being deployed to marketplace.

I think even if we want to add a VSIX file to Github, for any reason, all that needs done is for someone to run a `yarn build` and upload the VSIX so it wouldn't be much of a hassle.

jw3:
What's left here?  Would be nice to have the 1-commit check in master.

Shanedell:
@jw3 I don't think anything just second approval then squash and merge

jw3 (approved):
+1

jw3:
Looks good to me.  Looks like you need to resolve a conflict in build.sbt though.

PR #63, opened 2021-12-15
Daffodil Debugger Classpaths

### LINKED ISSUE #54: Included/Imported schema files not found
Something is up with class path. 

We have many schemas where one schema depends on another. E.g., PCAP which depends on ethernetIP. 

When I try to use daffdoil-vscode to debug PCAP, I specify the pcap.dfdl.xsd schema, then the icmp.cap data file, and then I get an error indicating daffodil is unable to resolve the inter-schema references. 

image

### LINKED ISSUE #57: need config dialog ability to setup daffodil classpath
Standard schema file layout uses src/main/resources as a root for include/import files, as well as src/test/resources. Hundreds of DFDL schemas exist using the standard DFDL schema project layout: https://daffodil.apache.org/dfdl-layout/

Users need a way to conveniently specify the directories which must be classpath roots. This can be rudimentary at first. (based on a folder/file chooser)

Furthermore, these directories can span mulltiple different directories - e.g., the PCAP DFDL schema (https://github.com/DFDLSchemas/PCAP) includes the ethernetIP DFDL schema (both are available on github).  So for PCAP one is going to need at least 4 classpath entries.

We need a way to point vscode at the src/main/resource and src/test/resource directories for multiple such schemas. Order on the classpath also matters of course, as schemas sometimes have test resources that must be superceded by DFDL schemas that include/import them. 

Eventually it would be good if vscode had a capability like 'sbt eclipse' for the Eclipse IDE, which outputs a detailed .classpath file that Eclipse uses to setup the classpath properly based on managed dependencies expressed in the build.sbt file. 

Creating something that becomes part of (or is referenced from) the launch.json automatically will be important for dealing with large schemas that include other schemas using managed dependencies. 

(See https://www.scala-sbt.org/1.x/docs/sbt-1.1-Release-Notes.html#VS+Code+extension+update as some sbt support already exists in vscode)

Shanedell:
Daffodil Debugger Classpaths:

- Create launch.json variable for daffodilDebugClasspath.
- Add daffodilDebugClasspath to getConfig and default config
- Create bash and bat templates for the scripts to run the debugger.
  - Created templates from the default of what we used.
  - Updated the templates to appened the value of the ENV DAFFODIL_DEBUG_CLASSPATH if it is set.
- Update launch wizard to have an option for class paths.
  - Created two different check boxes that allows the input box to be replaced or appened with the files selected.
  - If folder selected the extension creates a string with all the full paths to all jars inside of the folder to it.
  - If multiple files are selected they will all be added.
- Update issue with replace method, moved to replaceAll.

Closes #54
Closes #57

### DIFF
--- a/server/core/src/templates/bash-template
+++ b/server/core/src/templates/bash-template
@@ -1,0 +1,378 @@
+#!/usr/bin/env bash
+
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+###  ------------------------------- ###
+###  Helper methods for BASH scripts ###
+###  ------------------------------- ###
+
+die() {
+  echo "$@" 1>&2
+  exit 1
+}
+
+realpath () {
+(
+  TARGET_FILE="$1"
+  CHECK_CYGWIN="$2"
+
+  cd "$(dirname "$TARGET_FILE")"
+  TARGET_FILE=$(basename "$TARGET_FILE")
+
+  COUNT=0
+  while [ -L "$TARGET_FILE" -a $COUNT -lt 100 ]
+  do
+      TARGET_FILE=$(readlink "$TARGET_FILE")
+      cd "$(dirname "$TARGET_FILE")"
+      TARGET_FILE=$(basename "$TARGET_FILE")
+      COUNT=$(($COUNT + 1))
+  done
+
+  if [ "$TARGET_FILE" == "." -o "$TARGET_FILE" == ".." ]; then
+    cd "$TARGET_FILE"
+    TARGET_FILEPATH=
+  else
+    TARGET_FILEPATH=/$TARGET_FILE
+  fi
+
+  # make sure we grab the actual windows path, instead of cygwin's path.
+  if [[ "x$CHECK_CYGWIN" == "x" ]]; then
+    echo "$(pwd -P)/$TARGET_FILE"
+  else
+    echo $(cygwinpath "$(pwd -P)/$TARGET_FILE")
+  fi
+)
+}
+
+# TODO - Do we need to detect msys?
+
+# Uses uname to detect if we're in the odd cygwin environment.
+is_cygwin() {
+  local os=$(uname -s)
+  case "$os" in
+    CYGWIN*) return 0 ;;
+    *)  return 1 ;;
+  esac
+}
+
+# This can fix cygwin style /cygdrive paths so we get the
+# windows style paths.
+cygwinpath() {
+  local file="$1"
+  if is_cygwin; then
+    echo $(cygpath -w $file)
+  else
+    echo $file
+  fi
+}
+
+# Make something URI friendly
+make_url() {
+  url="$1"
+  local nospaces=${url// /%20}
+  if is_cygwin; then
+    echo "/${nospaces//\\//}"
+  else
+    echo "$nospaces"
+  fi
+}
+
+# This crazy function reads in a vanilla "linux" classpath string (only : are separators, and all /),
+# and returns a classpath with windows style paths, and ; separators.
+fixCygwinClasspath() {
+  OLDIFS=$IFS
+  IFS=":"
+  read -a classpath_members <<< "$1"
+  declare -a fixed_members
+  IFS=$OLDIFS
+  for i in "${!classpath_members[@]}"
+  do
+    fixed_members[i]=$(realpath "${classpath_members[i]}" "fix")
+  done
+  IFS=";"
+  echo "${fixed_members[*]}"
+  IFS=$OLDIFS
+}
+
+# Fix the classpath we use for cygwin.
+fix_classpath() {
+  cp="$1"
+  if is_cygwin; then
+    echo "$(fixCygwinClasspath "$cp")"
+  else
+    echo "$cp"
+  fi
+}
+# Detect if we should use JAVA_HOME or just try PATH.
+get_java_cmd() {
+  # High-priority override for Jlink images
+  if [[ -n "$bundled_jvm" ]];  then
+    echo "$bundled_jvm/bin/java"
+  elif [[ -n "$JAVA_HOME" ]] && [[ -x "$JAVA_HOME/bin/java" ]];  then
+    echo "$JAVA_HOME/bin/java"
+  else
+    echo "java"
+  fi
+}
+
+echoerr () {
+  echo 1>&2 "$@"
+}
+vlog () {
+  [[ $verbose || $debug ]] && echoerr "$@"
+}
+dlog () {
+  [[ $debug ]] && echoerr "$@"
+}
+execRunner () {
+  # print the arguments one to a line, quoting any containing spaces
+  [[ $verbose || $debug ]] && echo "# Executing command line:" && {
+    for arg; do
+      if printf "%s\n" "$arg" | grep -q ' '; then
+        printf "\"%s\"\n" "$arg"
+      else
+        printf "%s\n" "$arg"
+      fi
+    done
+    echo ""
+  }
+
+  # we use "exec" here for our pids to be accurate.
+  exec "$@"
+}
+addJava () {
+  dlog "[addJava] arg = '$1'"
+  java_args+=( "$1" )
+}
+addApp () {
+  dlog "[addApp] arg = '$1'"
+  app_commands+=( "$1" )
+}
+addResidual () {
+  dlog "[residual] arg = '$1'"
+  residual_args+=( "$1" )
+}
+addDebugger () {
+  addJava "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=$1"
+}
+
+require_arg () {
+  local type="$1"
+  local opt="$2"
+  local arg="$3"
+  if [[ -z "$arg" ]] || [[ "${arg:0:1}" == "-" ]]; then
+    die "$opt requires <$type> argument"
+  fi
+}
+is_function_defined() {
+  declare -f "$1" > /dev/null
+}
+
+# Attempt to detect if the script is running via a GUI or not
+# TODO - Determine where/how we use this generically
+detect_terminal_for_ui() {
+  [[ ! -t 0 ]] && [[ "${#residual_args}" == "0" ]] && {
+    echo "true"
+  }
+  # SPECIAL TEST FOR MAC
+  [[ "$(uname)" == "Darwin" ]] && [[ "$HOME" == "$PWD" ]] && [[ "${#residual_args}" == "0" ]] && {
+    echo "true"
+  }
+}
+
+# Processes incoming arguments and places them in appropriate global variables.  called by the run method.
+process_args () {
+  local no_more_snp_opts=0
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+             --) shift && no_more_snp_opts=1 && break ;;
+       -h|-help) usage; exit 1 ;;
+    -v|-verbose) verbose=1 && shift ;;
+      -d|-debug) debug=1 && shift ;;
+
+    -no-version-check) no_version_check=1 && shift ;;
+
+           -mem) echo "!! WARNING !! -mem option is ignored. Please use -J-Xmx and -J-Xms" && shift 2 ;;
+     -jvm-debug) require_arg port "$1" "$2" && addDebugger $2 && shift 2 ;;
+
+          -main) custom_mainclass="$2" && shift 2 ;;
+
+     -java-home) require_arg path "$1" "$2" && jre=`eval echo $2` && java_cmd="$jre/bin/java" && shift 2 ;;
+
+ -D*|-agentlib*|-XX*) addJava "$1" && shift ;;
+                 -J*) addJava "${1:2}" && shift ;;
+                   *) addResidual "$1" && shift ;;
+    esac
+  done
+
+  if [[ no_more_snp_opts ]]; then
+    while [[ $# -gt 0 ]]; do
+      addResidual "$1" && shift
+    done
+  fi
+
+  is_function_defined process_my_args && {
+    myargs=("${residual_args[@]}")
+    residual_args=()
+    process_my_args "${myargs[@]}"
+  }
+}
+
+# Actually runs the script.
+run() {
+  # TODO - check for sane environment
+
+  # process the combined args, then reset "$@" to the residuals
+  process_args "$@"
+  set -- "${residual_args[@]}"
+  argumentCount=$#
+
+  #check for jline terminal fixes on cygwin
+  if is_cygwin; then
+    stty -icanon min 1 -echo > /dev/null 2>&1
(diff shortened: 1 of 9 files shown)


mbeckerle:
I am not having any luck.

First I ran vscode and uninstalled the current daffodil extension. Exit vscode.

Then I checked out your PR branch, and did a yarn build.

Then I installed the vsix file into vscode via command line.

Start vscode again. Go to extensions. Daffodil is there. Click on settings gear.

I put this in the "User" tab (not Workspace tab) debug classpath via the extension settings button

/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/main/resources:/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/test/resources:/home/mbeckerle/.ivy2/local/com.owlcyberdefense/dfdl-ethernetip/1.0.0/jars/dfdl-ethernetip.jar

I have verified these are correct.

Those are the root folders for main and test resources for pcap, and the jar for ethernetIP.

I put this in absolute path to schema file:

/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/main/resources/com/tresys/pcap/xsd/pcap.dfdl.xsd

I put this in absolute path to data file:

/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/test/resources/com/tresys/pcap/data/icmp1.cap

There does not seem to be any Save or "Go" button to tell it to use these settings. 

Somehow I got a Daffodil Launch configuration. (launch.json has a configuration named "Daffodil Launch")

The schema file name isn't in the launch.json, It still just has ${file}, so I edited the schema file absolute path to the pcap.dfdl.xsd into the launch.json.

When I run, I get an SDE (and the blocking dialog does not allow me to copy the messages! )

The first SDE says line168 of pcap.dfdl.xsd unable to find eth:Ethernet type definition component. 

That is defined by the ethernetIP schema imported at line 50. 

So it is unable to resolve that file name: com/owlcyberdefense/dfdl/xsd/ethernetIP.dfdl.xsd

Further down in the SDE dialog (that I can't copy from), are a bunch of errors and warnings basically saying it can't find any of the files it is trying to import, including the daffodil built in schema file that this extends. 

So the classpath is definitely not working for me.

Shanedell:
@mbeckerle is this what it looks like for you for the launch wizard?
Screen Shot 2021-12-15 at 7 48 33 PM

If so your config was not properly saved as there seems to be a debug with the launch wizard. On my work machine the wizard looks how it should but on my personal looks like what is posted which means the styles and script did not get loaded properly. This could be a reason behind your issues but I am looking into it now will report back once I get something figured out.

Did you make your own `.vscode/launch.json` with these paths you sent?

Shanedell:
I was able to get the fix the VSIX file so that the launch wizard works properly. @mbeckerle If you have this exact config in `vscode/launch.json` does debugging work?

```
{
    "version": "0.2.0",
    "configurations": [
        {
            "request": "launch",
            "type": "dfdl",
            "name": "Wizard Config",
            "program": "/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/main/resources/com/tresys/pcap/xsd/pcap.dfdl.xsd",
            "data": "/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/test/resources/com/tresys/pcap/data/icmp1.cap",
            "debugServer": 4711,
            "infosetOutput": {
                "type": "none",
            },
            "trace": true,
            "stopOnEntry": true,
            "useExistingServer": false,
            "openHexView": false,
            "openInfosetView": false,
            "openInfosetDiffView": false,
            "daffodilDebugClasspath": "/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/main/resources:/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/test/resources:/home/mbeckerle/.ivy2/local/com.owlcyberdefense/dfdl-ethernetip/1.0.0/jars/dfdl-ethernetip.jar"
        }
    ]
}
```

If so it could have been dealing with the launch wizard setting or you were trying to use the global settings and it didn't work properly.

mbeckerle:
I will try exactly this on Thursday. My settings does NOT look exactly like your screen shot, so something is definitely wrong. Is it possible that the yarn build failed, but left the older vsix file sitting there?

Shanedell:
Could be but for either way the `styles.css` and `launchWizard.js` would not be getting loaded. I will push a commit up here shortly, you will want to pull that update down. Then rebuild and reinstall the VSIX file.

Is the settings you see, look like this?

Screen Shot 2021-12-15 at 8 28 48 PM

If so you were setting the global config not using the launch wizard. As stated in my previous comment there seems to be an issue that those globals do not get loaded when they are set and no `.vscode/launch.json` exists, this will need to be a separate issue I believe.

If you the the config I sent for your `.vscode/launch.json`, I am pretty confident it will work.

mbeckerle:
Does Daffodil-lib jar have to be on the debug classpath in order to use the DFDLGeneralFormat.dfdl.xsd schema file built-into daffodil? It is not on the debug classpath, because I'm assuming daffodil itself doesn't have to be.

Shanedell:
I don't think it needs to be can you try to this for your `.vscode/launch.json`
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "request": "launch",
            "type": "dfdl",
            "name": "Wizard Config",
            "program": "/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/main/resources/com/tresys/pcap/xsd/pcap.dfdl.xsd",
            "data": "/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-pcap/src/test/resources/com/tresys/pcap/data/icmp1.cap",
            "debugServer": 4711,
            "infosetOutput": {
                "type": "none",
            },
            "trace": true,
            "stopOnEntry": true,
            "useExistingServer": false,
            "openHexView": false,
            "openInfosetView": false,
            "openInfosetDiffView": false,
            "daffodilDebugClasspath": "/home/mbeckerle/.ivy2/local/com.owlcyberdefense/dfdl-ethernetip/1.0.0/jars/dfdl-ethernetip.jar"
        }
    ]
}
```

For me only have the jar makes the debug work maybe pointing at those additional resources is causing an issue?

mbeckerle:
Same SDE with this new launch.json

Shanedell:
Are you sure that is the proper path to your dfdl-ethernetip jar? Are you sure it isn't `/home/mbeckerle/Documents/dataiti/git/dfdl-schemas/dfdl-ethernet/target/dfdl-ethernetip-1.0.0.jar`, this would be more similar to the path I used.

The error you are getting is from pcap not being able to include an item that would be in that jar. So wondering if that jar does not exist or just a pathing error.

mbeckerle:
That jar is present at that path. Just double checked it. When I 'sbt publishLocal' on the ethernetIP schema project, this .ivy2/local directory is where the jar ends up.

Shanedell:
Okay I did not realize that the build publishes a jar there as well, when I used it for mine it worked fine maybe another bug with the VSIX or just something my work machine is not catching.

What OS are you using? Linux I assume by directories you listed but which distro? I will try to spin up a VM of the distro and test it out see if I can hit your issue that way as I can't duplicate it locally on my Mac.

mbeckerle:
image

mbeckerle:
Java 11 makes no difference. I rebuilt the vsix file using java 11, reinstalled it into vscode. running the launch config gives same SDE.

Shanedell:
Would the schemes need rebuilt for Java 11?

mbeckerle:
Schema jars do not contain java/scala code class files unless they are for a User-defined-function, or layering extension.

Shanedell:
Uh that is weird to me as I use java 11 so I will have to do some digging see if I can recreate the issue

Shanedell:
@mbeckerle I was able to recreate your issue on my Ubuntu VM so I will be doing some testing to get this fixed, will let you know once I believe I found the fix.

Shanedell:
@mbeckerle I believe I fixed the issue, I was doing bad replaces of characters from the class path causing issues when running, if you use either the first or second `.vscode/launch.json` should work for you now. You will need to uninstall the extensions rebuilt it and install and it should work.

mbeckerle:
I've got your latest commit (a1a7f006669a7a60e1fbe414e5094dfb83e8c2bb)

I did a checkout of that and did 'git clean -xdf' to remove existing cruft and the existing vsix.

That doesn't remove yarn.lock file, so I removed that by hand.

Then 'yarn build' created vsix file.

I installed it into vscode.

I refreshed the launch.json from the examples in this github thread.

I run it.

Same exact SDE as before. Unable to find any of the imports/includes. 

We need to turn on LogLevel Resolver(35) in Daffodil so we can see the classpath it is using. The API call is `setLoggingLevel(LogLevel.Resolver)`

Shanedell:
Okay so try these steps:
- 1.) Uninstall extension and delete the VSIX file
- 2.) Reload VSCode
- 3.) Remove both of these folders
```
~/.local/share/daffodil-dap
server/core/target/
```
- 4.) Run `yarn build`
- 5.) Install the VSIX file
- 6.) Try running schema again

---

If this does not work I think best way for me to pin point where the issue is doing this.

Can you edit this file

```
~/.local/share/daffodil-dap/daffodil-debugger-3.1.0-1.0.0-SNAPSHOT/bin/daffodil-debugger
```

Edit these lines

```
  if [[ "$DAFFODIL_DEBUG_CLASSPATH" != "" ]]; then
    app_classpath="$app_classpath:$DAFFODIL_DEBUG_CLASSPATH"
  fi
```

to be

```
  if [[ "$DAFFODIL_DEBUG_CLASSPATH" != "" ]]; then
    echo "$DAFFODIL_DEBUG_CLASSPATH" > output
    app_classpath="$app_classpath:$DAFFODIL_DEBUG_CLASSPATH"
  fi
```

After saving try debug the schema again and then post a comment here with the contents of this file

```
~/.local/share/daffodil-dap/daffodil-debugger-3.1.0-1.0.0-SNAPSHOT/bin/output
```

If this file does not exist after doing this then it is a different than I am thinking is happening.

mbeckerle:
Yessss. Things are starting to work. 

I can start the debugger with no SDE. 

It stops on the xs:schema element of PCAPHeader element, and within that on the Network element.

I continue and it stops at the PCAPHeader (Yay!)

I continue again and can stop on more things.

I noticed the "step into" button doesn't work "stepIn failure null" in Terminal. 

This is significant progress. 

I can't figure out how to access or set a breakpoint in the ethernetIP schema, which doesn't show up as files anywhere on the left, and since stepInto doesn't work.... Hmmm. 

I will try more variations later today. (Out for a couple hours now. - until noon ish)

mbeckerle:
I also tried using the debug classpath where PCAP and ethernetIP's src/main/resources and src/test/resources folders are on the classpath and that also seems to work. 

However there is no way to step into the ethernetIP schema, or set a breakpoint in there.

Shanedell:
@mbeckerle Excellent! Glad things are starting to work

Yeah I am not sure if there is a way to step into another imported schema, that may require to the Scala code for the debugger to allow for that to happen. I can mess around here and see if I can find anything but I am thinking that maybe a separate issue as I believe it will require and update to the Scala code.

arosien:
> I also tried using the debug classpath where PCAP and ethernetIP's src/main/resources and src/test/resources folders are on the classpath and that also seems to work.
>
> However there is no way to step into the ethernetIP schema, or set a breakpoint in there.

In talking with @Shanedell we'll reproduce what you're seeing @mbeckerle and see if there needs to be a patch to the backend.

Shanedell:
So from testing Mike is correct only the schema file listed as `program` will get debugged and can have breakpoints.

For PCAP it imports a schema in the same directory called `binaryDynamicByteOrder.dfdl.xsd`. It also imports the `ethernetIP.dfdl.xsd` schema file from the `ethernetIP` repo.

I tested this with the classpath importing the `ethernetIP` jar as well as importing its `resources` files. But in both of the imported schemas when I put a breakpoint on every single line of the schema being imported, those breakpoints are never hit (I only did this for `binaryDynamicByteOrder.dfdl.xsd`).

@mbeckerle @arosien Do you think this needs to be implemented in this PR or make a seperate ticket for this? Will this be a blocker for 1.0.0?

mbeckerle:
Getting stepping/breakpoints across multiple files working can be a separate PR. 

I'm still checking out functionality here though. E.g., I haven't yet seen any output XML from a parse if I just continue past all the breakpoints. This is possibly just that I'm doing it wrong. 

I do think debug  PCAP + EthernetIP needs to work, because that's an example we always refer people to, because so many people are familiar with IP packets and such as a format. 

I'm really hoping v1.0.0 of the debugger can be used in training that I need to do this winter. For that it has to let you build a "real" schema, not just one limited to single files. 

That said, "hoping" is the word here. I don't decide. We decide as a community, and if the PMC wants to release it sooner and wants to call it v1.0.0 I am not going to block things based on my preferences.

mbeckerle (approved):
+1 The PCAP + EthernetIP schema works in that I can step through breakpoints that are set in the pcap.dfdl.xsd file (only), and the parse does work in that I get the XML out in the Infoset.xml file when the debug continues to the end of the execution. 

So the classpath aspect is working, which is what this PR is about. 

We need a separate ticket about enabling multi-file breakpoints anywhere in the schema files.

server/core/src/templates/bash-template
```diff
@@ -263,8 +263,8 @@ run() {
     java_opts="${JAVA_OPTS}"
   fi
 
-  if [[ "$DAFFODIL_DEBUG_CLASSPATH" != "" ]]; then
-    app_classpath="$app_classpath:$DAFFODIL_DEBUG_CLASSPATH"
+  if [[ "$daffodilDebugClasspath" != "" ]]; then
```
mbeckerle, line 266: We don't have any actual style guidelines for this sort of thing. (Probably we should.)

The precedent we do have is that the CLI looks for DAFFODIL_CLASSPATH, and it is a documented feature. So perhaps we should stick with that convention, and revert this change?

mbeckerle (approved):
+1 The change is obviously code-wise fine.

I just wonder if this name change is desirable, see comment on first file.

server/core/src/templates/bash-template
Shanedell: So the change most dealt with the linting of TypeScript. In the `daffodilDebugger.ts` file when we export the ENV variable it will give a warning about `DAFFODIL_DEBUG_CLASSPATH` as TypeScript wants it be camel cased. However, if we don't care about this I can revert this change

mbeckerle: I am more comfortable sticking with DAFFODIL_DEBUG_CLASSPATH, particularly if a new lint tool would let us control/remove the warning for this particular style guideline. 

But this is in bash code, so I'm not sure why typescript cares about it or is even able to parse it.

Shanedell: So it doesn't care about it in this particular file the only file where it complains is in the `src/daffodilDebugger.ts` file but I will revert that now

src/daffodilDebugger.ts
```diff
@@ -180,12 +181,29 @@ export async function getDebugger(
...
+        : config.daffodilDebugClasspath
+
       // Start debugger in terminal based on scriptName
       let terminal = vscode.window.createTerminal({
         name: artifact.scriptName,
         cwd: `${rootPath}/daffodil-debugger-${daffodilVersion}-${LIB_VERSION}/bin/`,
         hideFromUser: false,
         shellPath: artifact.scriptName,
+        env: {
+          DAFFODIL_DEBUG_CLASSPATH: daffodilDebugClasspath,
```
jw3, line 205: you can add the following above this line for the linter

`// eslint-disable-next-line @typescript-eslint/naming-convention`

jw3 (approved):
+1

src/daffodilDebugger.ts
Shanedell: @jw3 When my other PR for the CI stuff is merged we will not be using `eslint` anymore as we will be using `prettier` and the way we have it configured, at least right now, it does not complain about this sort of thing.

PR #71, opened 2022-01-10
Fix Launch Wizard for Windows

### LINKED ISSUE #67: Launch Wizard - Windows Fix
Need to fix the launch wizard for Windows. Believe the issue deals with pathing and it causes the launch wizard to not import the stylesheet and JS script it needs to work properly.

Shanedell:
Changes:
- Update launchWizard.ts to work for Windows
  - For this to work as the different paths caused issues, I opted to switch to reading both the javascript and stylesheet files then putting the content into the HTML as this allows for Linux, Mac and Windows to be able to use the Launch Wizard.

Closes #67

### DIFF
--- a/src/launchWizard/launchWizard.ts
+++ b/src/launchWizard/launchWizard.ts
@@ -18,6 +18,7 @@
 import * as vscode from 'vscode'
 import * as fs from 'fs'
 import { getConfig } from '../utils'
+import * as os from 'os'
 
 const defaultConf = getConfig('Wizard Config', 'launch', 'dfdl')
 
@@ -52,12 +53,15 @@ async function createUpdateConfigFile(data, updateOrCreate) {
     fs.mkdirSync(`${rootPath}/.vscode`)
   }
 
+  const launchPath =
+    os.platform() === 'win32'
+      ? `/${rootPath}/.vscode/launch.json`
+      : `${rootPath}/.vscode/launch.json`
+
   // Create launch.json if it doesn't exist already
   if (!fs.existsSync(`${rootPath}/.vscode/launch.json`)) {
     fs.writeFileSync(`${rootPath}/.vscode/launch.json`, data)
-    vscode.window.showTextDocument(
-      vscode.Uri.parse(`${rootPath}/.vscode/launch.json`)
-    )
+    vscode.window.showTextDocument(vscode.Uri.parse(launchPath))
     return
   }
 
@@ -112,9 +116,7 @@ async function createUpdateConfigFile(data, updateOrCreate) {
     }
   }
 
-  vscode.window.showTextDocument(
-    vscode.Uri.parse(`${rootPath}/.vscode/launch.json`)
-  )
+  vscode.window.showTextDocument(vscode.Uri.parse(launchPath))
 }
 
 // Function to update the config values in the webview panel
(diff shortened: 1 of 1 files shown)


src/launchWizard/launchWizard.ts
```diff
@@ -52,12 +53,15 @@ async function createUpdateConfigFile(data, updateOrCreate) {
     fs.mkdirSync(`${rootPath}/.vscode`)
   }
 
+  const launchPath =
+    os.platform() === 'win32'
```
mbeckerle, line 57: I can hardly believe you'll get away without needing something like this:
```
osPath(`${rootPath}/.vscode/launch.json`)
```
Which buries the os-specific check down where it is not clutter. 

We made this mistake of allowing OS conditionalizations to proliferate in the CLI test code in Daffodil, and since we didn't nip it in the bud, there's an open ticket to basically convert hundreds of tests that have inline OS conditionalizations to use an osPath like library and otherwise push this os/path stuff down a level.

src/launchWizard/launchWizard.ts
```diff
@@ -280,10 +282,14 @@ class LaunchWizard {
...
-    ).with({ scheme: 'vscode-resource' })
+    )
     const styleUri = vscode.Uri.parse(
       this.ctx.asAbsolutePath('./src/launchWizard/styles.css')
-    ).with({ scheme: 'vscode-resource' })
+    )
+    const scriptData = fs.readFileSync(scriptUri.fsPath)
+    const styleData = fs.readFileSync(styleUri.fsPath)
+
+    // const styleUri = this.getPanel().webview.asWebviewUri(stylePath)
```
mbeckerle, line 292: Commented out code. I'd say delete or add a comment to the line why you want it retained. 
Eg. "// Todo: keep for now. Still thinking about this."

mbeckerle (approved):
+1 Just small suggestion.

src/launchWizard/launchWizard.ts
Shanedell: So from my testing on a Windows VM, I believe we don't need the osPath. I believe the open text document file function may be converting paths under the hood as I had never had issues with the class paths until trying to load the style and script files for the launch wizard. Everything else worked fine I assume because it may be handled by a lot of the VSCode API functions by default

mbeckerle: Ok. I'm fine with that if the basic libraries already are dealing with this. 

 It seems odd that this is the only place we need to do OS conditionalization though.

mbeckerle: If the basic libraries are really covering this, and this is likely one of the only places we have to do os conditionalization (on paths anyway) then I think this code needs a comment to that effect. 

Reading this code, a naive typescript coder (like me in a year when I've forgotten all this) would just start proliferating these os.platform() === 'win32' checks. We want to avoid that.

Shanedell: Yeah so from my understanding most things handle this but in this case we are constructing the entire path to our file based on the folder and on Windows it needs an additional `/` at the beginning of the path.

jw3 (approved):
+1

I only tested that this continues to work on Linux, I don't have a Windows box available at the moment.

Issue #16, opened 2021-10-08
Restructure Scala code to support JDK 8


jw3:
Follow up to https://github.com/jw3/example-daffodil-vscode/issues/94:

@arosien said

- Daffodil required JDK 8, so we could try to also support it
- We currently need JDK 11 to support use of Path.of, which could perhaps be replaced with something available in older JVMs.
- Nested object in Scala isn't supported until JDK 9 or greater.

Original issue https://github.com/jw3/example-daffodil-vscode/issues/136

stevedlawrence:
I'm going to reopen this issue. We've solved the compile on Java 8 issue, but not the run on Java 8. Running is much lower priority, but it might want to be fixed some day.

scholarsmate:
Is Java 8 still something we need to support or are we requiring Java 11 or greater?

stevedlawrence:
Since Daffodil supports Java 8, and there are systems that a still on Java 8, I think there would be value in being able to debug on Java 8 to ensure environments are the same. This doesn't feel high priority to me though.

naga-panchumarty:
@stevedlawrence do you think we can close this issue? Or do you think we should do anything else?

naga-panchumarty:
Thank you, closing it

PR #73, opened 2022-01-19
Bump extension daffodilVersion to 3.2.1

### LINKED ISSUE #74: Specify Daffodil version at one location
Currently we specify the Daffodil version for SBT in [`build.sbt`](https://github.com/apache/daffodil-vscode/blob/main/build.sbt#L18) and for the backend artifact unpacker in [`daffodilDebugger.ts`](https://github.com/apache/daffodil-vscode/blob/main/src/daffodilDebugger.ts#L44).

This duplication has proven to be error prone and could be eliminated by adding an entry to `package.json` and then reading it out in each place.

### LINKED ISSUE #75: Classpath shouldn't be hardcoded in the template
Here is the offending line
https://github.com/apache/daffodil-vscode/blob/main/server/core/src/templates/bash-template#L369

SBT Native Packager normally generates that for the release based on the dependencies, we do not want to be responsible for that process.

The reason we provide a template is to provide the schema extension classpath from the VS Code side.  That should be able to be injected into the generated template rather than specifying a full override of the template.

Shanedell:
Updates:

- Create `daffodilVersion` entry inside of the package.json
- Update the build.sbt to get the `daffodilVersion` from package.json
- Update extension to have a method for getting the `daffodilVersion` from the `package.json`
- Added a unit test for the extension, that tests the getting of the `daffodilVersion`
- Updated bash and bat templates to not hard code app classpaths.
  - This required of making a new variable, `new_classpath` and `NEW_CLASSPATH`, into the templates as the app classpath vars are READ only.

Tested this on mac and was able to successfully connect to run the debugger.

@jw3 @stevedlawrence This should be good for review.

Closes #74 
Closes #75


Shanedell:
@mbeckerle @stevedlawrence Was the class `org.apache.daffodil.debugger.Debugger` moved to a different package or classpath?

I am having trouble testing the extension with Daffodil 3.2.1 because one of our Scala files imports `org.apache.daffodil.debugger.Debugger` but an exception is thrown that it can't be found.

stevedlawrence:
I don't think anything has changed with debugging. It also looks like the current main branch already has this update, merged in commit  8cdefdd73a8a8fb1db34c2b99e01dbaacf6a5c4e

I can't build the repo at all right now to see if I can reproduce, sbt compile fails with errors about `java.nio.file.Path.of` not being valid, maybe that's related?

Seems the repo is broken right now. Do the github workflows not build the jar?

Shanedell:
@stevedlawrence That commit had it for the `build.sbt` and I believe something in Scala. This update is updating the VS Code extension to target the proper version of Daffodil when its running. 

So to be more specific. How it is now is that the extension will try to run `sbt universal:packageBin` for `3.2.1` creating package `daffodil-debugger-3.2.1-1.0.0-SNAPSHOT.zip`, however the extension will constantly try to run that command because it thinks the package name would be `daffodil-debugger-3.1.0-1.0.0-SNAPSHOT.zip` since `daffodilVersion = 1.0.0`.

As for the sbt issues I am unaware of `java.nio.file.Path.of` as I am not seeing that if I run `sbt compile` or `sbt universal:packageBin`.

I am also unaware of what you mean about the workflows not building the jar, do you mind explaining?

jw3:
> sbt compile fails with errors about java.nio.file.Path.of not being valid

Sounds like the wrong JDK, ensure it is 11.

> I am having trouble testing the extension with Daffodil 3.2.1 because one of our Scala files imports org.apache.daffodil.debugger.Debugger but an exception is thrown that it can't be found

This classpath issue is coming from the [sbt native packager bash template](https://github.com/apache/daffodil-vscode/blob/main/server/core/src/templates/bash-template#L369) which has the classpath hard coded to the 3.1.0 version.  That approach needs improved.  Seems to me there is a way to inject sections rather than specifying the full template.

> merged in commit 8cdefdd

Yeah, looks like I missed updating the version in the other location in that commit.  Instead of just updating it again in this PR it should be fixed to prevent future issues by centralizing to a single location.

The package json seems like the logical place for that, where it can then be read out in the typescript and sbt dependents.

jw3:
Added https://github.com/apache/daffodil-vscode/issues/74

stevedlawrence:
> Sounds like the wrong JDK, ensure it is 11.

Yeah, this was my issue. I didn't realize we require Java 11. Just curious, is there a reason for this?

jw3:
There are a couple reasons, one requiring minimum of 9 and another requiring 11.  We could have refactored the Path.of to get back to 9, but figured we would just do it all at once and get back to 8.

Tracked here, past discussions are linked in it
- https://github.com/apache/daffodil-vscode/issues/16

jw3 (approved):
Tested and working on Linux.

jw3:
The `package.json` could be deserialized into an object to make that aspect a little more safe and scalable, but I'd say save that for the next instance where we need to pull something from that file and get this one merged.

mbeckerle (approved):
+1  So besides the daffodil version, this is also adding DAFFODIL_DEBUG_CLASSPATH entries if defined. Good.

arosien:
@Shanedell @jw3 @mbeckerle tag as 1.0.0?

stevedlawrence:
> tag as 1.0.0?

Are you asking about tagging the code as a 1.0.0 release or tagging this PR/issue as the target milestone?

If code, keep in mind code should not be tagged until it goes through the release process. I think we still need a document that describes the release candidate procedures, likely similar to the Daffodil release workflow: https://cwiki.apache.org/confluence/display/DAFFODIL/Release+Workflow

PR #77, opened 2022-02-02
Multi-file breakpoints: change Path to URI when referencing locations.

### LINKED ISSUE #66: Multi-file Breakpoint
Need to enable breakpoints in imported files/schemas.

A good schema for testing this will be `PCAP` as it imports a schema from the same directory as well as the schemas for `ethernetIP`.

For this item to be completed you should be able to set a breakpoint in every `.xsd` file being imported by a schema and the breakpoints are hit during debug.

- Example if you put breakpoint in every `.xsd` file in the `PCAP` and `ethernetIP` schema repos. When you debug the breakpoints in every file should be hit

arosien:
Fixes #66.

### DIFF
--- a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
+++ b/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
@@ -249,7 +249,7 @@ class DAPodil(
       case DAPodil.State.Launched(debugee) =>
         for {
           _ <- debugee.setBreakpoints(
-            Path.of(args.source.path),
+            Path.of(args.source.path).toUri(),
             args.breakpoints.toList.map(bp => DAPodil.Line(bp.line))
           )
           breakpoints = args.breakpoints.toList.zipWithIndex.map {
@@ -512,7 +512,7 @@ object DAPodil extends IOApp {
     def step(): IO[Unit]
     def continue(): IO[Unit]
     def pause(): IO[Unit]
-    def setBreakpoints(path: Path, lines: List[DAPodil.Line]): IO[Unit]
+    def setBreakpoints(uri: URI, lines: List[DAPodil.Line]): IO[Unit]
     def eval(args: EvaluateArguments): IO[Option[Types.Variable]]
   }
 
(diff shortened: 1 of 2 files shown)


Shanedell:
@arosien So this will not allow the loading of files in the classpath correct? But does it allow for loading local schemas?

What I mean is:
- Say if I am testing the `pcap` daffodil schema that uses the `ethernet` schema. I should see these things right?
  -  I add the JAR for `ethernet` to classpath but the files inside of that JAR cannot have breakpoints set or hit
  - Ability to set break points in the `src/main/resources/com/tresys/pcap/xsd/binaryDynamicByteOrder.dfdl.xsd` and they should be hit
 
Is this correct or am I missing something?

arosien:
> @arosien So this will not allow the loading of files in the classpath correct? But does it allow for loading local schemas?
> 
> What I mean is:
> 
> * Say if I am testing the `pcap` daffodil schema that uses the `ethernet` schema. I should see these things right?
>   
>   * I add the JAR for `ethernet` to classpath but the files inside of that JAR cannot have breakpoints set or hit

Correct.

>   * Ability to set break points in the `src/main/resources/com/tresys/pcap/xsd/binaryDynamicByteOrder.dfdl.xsd` and they should be hit

Correct, although I'm not sure there are any "breakpoint-able" lines in `binaryDynamicByteOrder.dfdl.xsd` because it is only type definitions. I may be wrong.

I *would* like to have an example multi-file schema to add unit tests for, let's look around for one.

mbeckerle:
So you are looking for a schema with import statements that refer to other files of the same schema?

ISO8583 is one such. github https://github.com/DFDLSchemas/ISO8583/blob/master/1993/ISO8583_1993.xsd is a file with two import statements.

arosien:
Hi @Shanedell, @mbeckerle, @stevedlawrence, can you take a look at this PR and give a 👍 or 👎 ?

mbeckerle (approved):
+1 Changes all seem as expected for URI-ifying things.

tuxji (approved):
+1

Scala changes look good.  If you need to test more thoroughly, someone said the [pcap](https://github.com/DFDLSchemas/PCAP) schema is a good one since it uses two includes.

Shanedell:
@mbeckerle Am I good to merge this in since you and @tuxji both approved? I believe @arosien won't be able to. 

I think the changes are good just needed to find a good schema layout to test the breakpoints in different files. But either way these changes are valuable.

mbeckerle:
Yes you can merge this.

You need a DFDL schema for testing with specifically what characteristics?
Main schema is multiple files?

mbeckerle:
Well, secondary schema files will often not have any global elements. But if they have complex types or group definitions those may contain local element decls that want to be breakpoints. I am assuming one can set a breakpoint on a local element decl.

PR #79, opened 2022-02-10
Support Java 8

### LINKED ISSUE #16: Restructure Scala code to support JDK 8

stevedlawrence:
Closes #16

### DIFF
--- a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
+++ b/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
@@ -34,6 +34,7 @@ import fs2.concurrent.Signal
 import java.io._
 import java.net._
 import java.nio.file.Path
+import java.nio.file.Paths
 import org.typelevel.log4cats.Logger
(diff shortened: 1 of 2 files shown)


mbeckerle (approved):
+1

jw3:
The nio changes are good but the nested object issue in #16 remains.

stevedlawrence:
These changes allow building on Java 8, but due to issue #16, it still fails to run on Java 8. We can still build on Java 8 as long as we run on Java 9+. Maybe we can change our build requirements to Java 8+ but run requirements to Java 9+?

jw3:
That sounds fine to me.

jw3 (approved):
+1 for build on 8 run on 9 until we can address the nested object issue

PR #85, opened 2022-02-18
Image, Icon, README Fixes

### LINKED ISSUE #81: yarn disallows SVG's in the README
When running yarn build with yarn version 1.22.15, I get the following error:

> ERROR  SVGs are restricted in README.md; please use other file image formats, such as PNG: https://daffodil.apache.org/assets/themes/apache/img/apache-daffodil-logo.svg

Should we create a png version of the logo on our website that this README can reference?

The README also has links to github workflow badges that are svg. I can't find a way to get png badges instead, but yarn doesn't seem to care about them, so maybe we can ignore them? Alternatively, shields.io (which we use for the Daffodil readme) does seem to have a way to use pngs, although they do not recommend it.

### LINKED ISSUE #82: yarn build fails with missing icon
The package.json file lists the icon as `images/daffodil.jpg`, but this file doesn't exis causing yarn build to fail.

### LINKED ISSUE #83: Rename vsix convenience binary?
The .vsix convenience binary created by running yarn buil is currently name `daffodil-debugger-VERSION.vsix`. This doesn't have "apache" in the name, which ASF usually requires. It is also specific to the debugger aspect of this extension, but the extension may very likely grow to include other capabilities (e.g. auto-completion).

Thoughts on renaming the resulting .vsix file to something with apache in the name and more generic like `apache-daffodil-vscode-VERSION.vsix`. This makes it clear that this is specific to vscode, matches the naming of this repository, and matches what the likely source release would look like (e.g. apache-daffodil-vscode-VERSION-src.zip`).

Shanedell:
Multiple Updates:

- Update apache daffodil logo to be png not svg
- Fix icon issue by:
  - Downloading apache daffodil logo locally to images folder
  - Fix path to it in package.json `images/daffodil.ico`
- Update binary name so that when yarn build run it creates `apache-daffodil-vscode-VERSION.vsix`
  - change `name` attribute inside of package.json to `apache-daffodil-vscode`
- Update ratCheck to ignore `images/daffodil.ico` instead of `images/daffodil.jpg`

Closes #81
Closes #82
Closes #83


package.json
```diff
@@ -10,7 +10,7 @@
   "engines": {
     "vscode": "^1.55.0"
   },
-  "icon": "images/daffodil.jpg",
+  "icon": "images/daffodil.png",
```
stevedlawrence, line 13: Should this be an actual icon file, or a standard square shape? Does this icon show up some where in VS Code?

We do have an actual .ico file used for the Windows MSI that we could reuse for this if an .ico file makes more sense than the log:

https://raw.githubusercontent.com/apache/daffodil/main/daffodil-cli/src/windows/apache-daffodil.ico

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "daffodil-debugger",
+  "name": "apache-daffodil-vscode",
   "displayName": "Apache Daffodil Schema Debugger",
```
stevedlawrence, line 3: Should this name be Apache Daffodil VS Code Extension? Or is this display name for something else?

tuxji (approved):
+1

So easy to review and approve these changes, although I see Steve asked a good question.

package.json
Shanedell: It could be if that would be more desired

package.json
Shanedell: So this icon will appear when you install extension here is a screenshot of mine with the current image

Screen Shot 2022-02-18 at 11 46 54 AM

This is with the image you linked

Screen Shot 2022-02-18 at 11 50 24 AM

Which do you prefer? I personally prefer the image you linked.

stevedlawrence: Yeah, I think the .ico looks better. I also found the [VS Code package.json documentation](https://code.visualstudio.com/api/references/extension-manifest) that says the icon property is:

> The path to the icon of at least 128x128 pixels (256x256 for Retina screens).

Doesn't say anything about the supported formats, but it looks like the .ico  works for you. And I believe the .ico file has a bunch of built-in resolutions, ranging from 16x16 to 256x256, so should scale nicely if VS Code uses it anywhere else.

package.json
stevedlawrence: Yeah, seems commit b5185aa26fefd0c39b47957e84467959875dc348 updated most places to use the "Apache Dafodil VS Code Extension" name. Probably just missed this one.

stevedlawrence (approved):
+1 :+1:

PR #91, opened 2022-03-11
Avoid endElement parsing callbacks when debugger is paused.

### LINKED ISSUE #90: Avoid popping the stack when breakpoint is hit
When a breakpoint is hit in a Daffodil `startElement` callback, it sends a DAP event which subsequently requests the current stack. However, the Daffodil debugger continues and invokes the `endElement` callback with pops the computed stack, so there is a race where the pop wins and the requested stack doesn't match the stack at the point when the breakpoint was hit.

The `endElement` callback needs to respect the pause state as well.

arosien:
Previously it would emit an event that would pop the maintained stack
so when a breakpoint was hit the subsequently fetched stacktrace
was missing the breakpoint.

Fixes #90.


Shanedell (approved):
+1

PR #96, opened 2022-03-26
RC2 Updates

### LINKED ISSUE #97: RC2 Updates
Update repo to fix issues present in the RC2 release these include:

- Update to daffodil 3.3.0
- Update build/bin.NOTICE
- Update dependencies versions, as long as they don't break anything.

Shanedell:
Multiple updates from rc2 release:

- Update daffodilVersion from 3.2.1 to 3.3.0.
- Add plugins to project/plugins.sbt.
  - These plugins allow for checking dependencies for updates, CVEs and viewing the tree of dependencies.
- Update scalaVersion from 2.12.13 to 2.12.15.
  - Can move to 2.13.8 at some point not sure when we want to do this however.
- Update "logback-classic" version from 1.2.3 to 1.2.11 to fix CVEs.
- Update "com.microsoft.java.debug.core" version from 0.31.1 to 0.35.0.
- Update "fs2-io" version from 3.0.4 to 3.0.6.
  - Updating to 3.2.5 breaks the build.
- Update "decline-effect" from 2.1.0 to 2.2.0.
- Update "log4cats-slf4j" from 2.1.0 to 2.1.1.
  - Updating to 2.2.0 breaks the build.
  - CVEs seem to be false positives here.
- Add dependencyOverrides for commons-lang3 to fix a CVE item.
- Update bin.NOTICE files as some of the incorated projects NOTICE's were updated.

Closes #97

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,9 +34,12 @@ lazy val commonSettings = {
       "org.apache.daffodil" %% "daffodil-sapi" % daffodilVer,
       "org.apache.daffodil" %% "daffodil-runtime1" % daffodilVer
     ),
+    dependencyOverrides ++= Seq(
+      "org.apache.commons" % "commons-lang3" % "3.12.0"
+    ),
     licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
     organization := "org.apache.daffodil",
-    scalaVersion := "2.12.13",
+    scalaVersion := "2.12.15",
     scalacOptions ++= Seq("-Ypartial-unification"),
     // remove the -Xcheckinit option added by the sbt tpoletcat plugin. This
     // option leads to non-reproducible builds
@@ -73,11 +76,11 @@ lazy val core = project
   .settings(
     name := "daffodil-debugger",
     libraryDependencies ++= Seq(
-      "ch.qos.logback" % "logback-classic" % "1.2.3",
-      "com.microsoft.java" % "com.microsoft.java.debug.core" % "0.31.1",
-      "co.fs2" %% "fs2-io" % "3.0.4",
-      "com.monovore" %% "decline-effect" % "2.1.0",
-      "org.typelevel" %% "log4cats-slf4j" % "2.1.0"
+      "ch.qos.logback" % "logback-classic" % "1.2.11",
+      "com.microsoft.java" % "com.microsoft.java.debug.core" % "0.35.0",
+      "co.fs2" %% "fs2-io" % "3.0.6",
+      "com.monovore" %% "decline-effect" % "2.2.0",
+      "org.typelevel" %% "log4cats-slf4j" % "2.1.1"
     ),
     buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion, "daffodilVersion" -> daffodilVer),
     buildInfoPackage := "org.apache.daffodil.debugger.dap",
--- a/build/bin.LICENSE
+++ b/build/bin.LICENSE
@@ -1,222 +1,400 @@
 
-                               Apache License
-                         Version 2.0, January 2004
-                      http://www.apache.org/licenses/
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-  TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-  1. Definitions.
+   1. Definitions.
 
-    "License" shall mean the terms and conditions for use, reproduction,
-    and distribution as defined by Sections 1 through 9 of this document.
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
 
-    "Licensor" shall mean the copyright owner or entity authorized by
-    the copyright owner that is granting the License.
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
 
-    "Legal Entity" shall mean the union of the acting entity and all
-    other entities that control, are controlled by, or are under common
-    control with that entity. For the purposes of this definition,
-    "control" means (i) the power, direct or indirect, to cause the
-    direction or management of such entity, whether by contract or
-    otherwise, or (ii) ownership of fifty percent (50%) or more of the
-    outstanding shares, or (iii) beneficial ownership of such entity.
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
 
-    "You" (or "Your") shall mean an individual or Legal Entity
-    exercising permissions granted by this License.
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
 
-    "Source" form shall mean the preferred form for making modifications,
-    including but not limited to software source code, documentation
-    source, and configuration files.
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
 
-    "Object" form shall mean any form resulting from mechanical
-    transformation or translation of a Source form, including but
-    not limited to compiled object code, generated documentation,
-    and conversions to other media types.
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
 
-    "Work" shall mean the work of authorship, whether in Source or
-    Object form, made available under the License, as indicated by a
-    copyright notice that is included in or attached to the work
-    (an example is provided in the Appendix below).
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
 
-    "Derivative Works" shall mean any work, whether in Source or Object
-    form, that is based on (or derived from) the Work and for which the
-    editorial revisions, annotations, elaborations, or other modifications
-    represent, as a whole, an original work of authorship. For the purposes
-    of this License, Derivative Works shall not include works that remain
-    separable from, or merely link (or bind by name) to the interfaces of,
-    the Work and Derivative Works thereof.
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
 
-    "Contribution" shall mean any work of authorship, including
-    the original version of the Work and any modifications or additions
-    to that Work or Derivative Works thereof, that is intentionally
-    submitted to Licensor for inclusion in the Work by the copyright owner
-    or by an individual or Legal Entity authorized to submit on behalf of
-    the copyright owner. For the purposes of this definition, "submitted"
-    means any form of electronic, verbal, or written communication sent
-    to the Licensor or its representatives, including but not limited to
-    communication on electronic mailing lists, source code control systems,
-    and issue tracking systems that are managed by, or on behalf of, the
-    Licensor for the purpose of discussing and improving the Work, but
-    excluding communication that is conspicuously marked or otherwise
-    designated in writing by the copyright owner as "Not a Contribution."
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
 
-    "Contributor" shall mean Licensor and any individual or Legal Entity
-    on behalf of whom a Contribution has been received by Licensor and
-    subsequently incorporated within the Work.
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
 
-  2. Grant of Copyright License. Subject to the terms and conditions of
-    this License, each Contributor hereby grants to You a perpetual,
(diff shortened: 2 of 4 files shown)


tuxji (approved):
+1

mbeckerle (approved):
+1

build/bin.NOTICE
```diff
@@ -165,7 +181,7 @@ ip4s (lib/com.comcast:ip4s-core.jar-<VERSION>.jar)
 
 Apache Commons Lang (lib/org.apache.commons:commons-lang3-<VERSION>.jar)
   Apache Commons Lang
-  Copyright 2001-2021 The Apache Software Foundation
-  
+  Copyright 2001-2022 The Apache Software Foundation
+
   This product includes software developed at
   The Apache Software Foundation (https://www.apache.org/).
```
stevedlawrence, line 187: We just made a bunch of corrections to this file that are accurate for 3.3.0. You might need to incorporate some of those changes.

project/plugins.sbt
```diff
@@ -21,3 +21,17 @@ addSbtPlugin("io.github.davidgregory084" % "sbt-tpolecat" % "0.1.17")
...
+  * dependencies of dependencies.
+  *
+  * Commands:
+  *   Check dependencies for CVEs -> sbt dependencyCheckAggregate
+  *   Print out dependency tree -> sbt dependencyTree
+  *   Check for available dependency updates -> sbt dependencyUpdates
+  */
+addDependencyTreePlugin
+addSbtPlugin("net.vonbuchholtz" % "sbt-dependency-check" % "4.0.0")
+addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.6.1")
```
stevedlawrence, line 37: Personally, I think these plugins should be part of ~/.sbt/plugins/plugins.sbt rather than the repo since these aren't strictly needed for the vscode repo. Minimizes dependencies for people that don't care about this this stuff. But I don't feel that strongly about it.

stevedlawrence (commented):
+1, might need to pull in changes or 3.3.0

stevedlawrence:
Also, In general we try to ensure every PR has an associated Issue associated with it for tracking purposes. It would be good to create an issue for these changes and and add a `Closes` line to the commit message.

build/bin.NOTICE
Shanedell: @stevedlawrence So I updated this on the weekend to contain the contents I found in `daffodil/NOTICE` was that not the correct fix for this?

stevedlawrence: That was correct, but @tuxji opened a new PR (https://github.com/apache/daffodil/pull/777) to fix some NOTICE issues we found during the Daffodil 3.3.0 vote. We decided it wasn't worth blocking the release and fixed them shortly after. That PR hasn't been merged yet, but should probably be merged relatively soon. At which point the relevant parts can be copied in. Note that PR also makes a minor update to the license file too.

Shanedell: So I am confused. That PR is for updating the `bin.NOTICE` and `bin.LICENSE` files for Daffodil. These wouldn't be inside of the `bin.NOTICE` and `bin.LICENSE` correct? Just the root `NOTICE` and `LICENSE` file contents should be inside of the `bin.` files inside of `daffodil-vscode` correct? As the `bin` files between the two should differ as the daffodil-debugger Scala code doesn't use the same dependencies as the daffodil project.

Shanedell: Okay I can work on getting these changes updated, should I just the contents of that PR or whats in main?

stevedlawrence: You might wait until things are merged. I'm not sure what the final PR will look like, and the more similar the two projects are, the easier it is to verify.

Shanedell: Okay that makes sense, do you want those other dependencies removed or anything else updated before I merge in this PR in? After merge would I be good to release RC3 or do you think I should wait till that daffodil PR is merged than fix the bin files and make another PR or keep this one open till that time?

stevedlawrence: I think the dependency changes look fine. RC3 should be good once this is merged, but I might make sense to include the bin.NOTICE and bin.LICENSE updates in this PR. Also, did you confirm that the updated dependencies did not have a change to their LICENSE or NOTICE files that needs to be incorporated into bin.NOTICE/bin.LICENE?

Shanedell: Yeah for all the ones I could fine most didn't change or if they did it was normally a simple bump of copyright year. The main thing was not knowing what I needed to copy over from daffodil files.

Shanedell: @stevedlawrence Hey Steve, I am going through and updating stuff here so if something doesn't have a NOTICE but has a LICENSE it can be excluded from `bin.NOTICE` if the license is added to `bin.LICENSE` correct? What about something that has neither LICENSE or NOTICE does that just not get added to either file? What if the LICENSE is an Apache license does that need added to that file or we bypass those?

Shanedell: So we have these items included but they aren't in either `bin.NOTICE` or `bin.LICENSE` but have Apache Licenses
```
org.typelevel.cats-effect
org.typelevel.cats-effect-kernel
org.typelevel.cats-effect-std
com.monovore.decline-effect
com.monovore.decline
org.typelevel.log4cats-core
org.typelevel.log4cats-slf4j
com.google.code.gson.gson
org.typelevel.literally
```
This one is in neither and has neither
```
org.typelevel.simulacrum-scalafix-annotations
```
So what do you think is best for these ones just leave them out? Adding the Apache license for all of the above, some are in the same project so it lowers the number down a bit, will cause that files to a bit more lengthy

stevedlawrence: For dependencies that are under the Apache License, you must add the dependencies NOTICE content to bin.NOTICE. If it does not have a NOTICE file then you do not need to do anything for that dependency. And because it's ALv2 license, you not need to add anything to the bin.LICENSE file since it's the same as the vscode license.

For dependencies that are under a license other than the Apache License, you must add its LICENSE information to the bin.LICENSE file. Nothing should be added to the bin.NOTICE file, unless the license is a [Category B](https://www.apache.org/legal/resolved.html#category-b) license, but we don't see those very often, and in most cases the information should be in a README rather than the bin.NOTICE.

If a dependency does not have a LICENSE file, then we need to figure out what its license is and handle it according to the above rules. For example, sometimes the license is just mentioned in a README or file headers. If we cannot determine the license, it is assumed to be [Category X](https://www.apache.org/legal/resolved.html#category-x) and we cannot use the dependency until the author/authors specify a license and we determine its compatibility with ASF licensing.

stevedlawrence: Sorry, I didn't see the other comments until after I posted the above, but I think it answers your questions.

> Also in the PR for daffodil some of the items that have no NOTICE in JAR have the license mentioned in bin.LICENSE so could those just be removed from the bin.NOTICE? Some of them also have Apache licenses so should those be removed from the bin.NOTICE file?

I suggested to John that we do not include anything in bin.NOTICE for dependencies that do not have a NOTICE file. I'm not sure if a final decision has been made about that, but that is my preference.

build.sbt
```diff
@@ -34,9 +34,12 @@ lazy val commonSettings = {
       "org.apache.daffodil" %% "daffodil-sapi" % daffodilVer,
       "org.apache.daffodil" %% "daffodil-runtime1" % daffodilVer
     ),
+    dependencyOverrides ++= Seq(
+      "org.apache.commons" % "commons-lang3" % "3.12.0"
+    ),
     licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
     organization := "org.apache.daffodil",
-    scalaVersion := "2.12.13",
+    scalaVersion := "2.12.15",
```
mbeckerle, line 42: This change is, FYI, super important because the latest LTS Java version, 17, requires this version of the scala libraries or obscure errors occur. 

And there are good reasons to want Java 17, including that it is noticibly faster.

mbeckerle (approved):
+1 I have one suggestion due to awkward wording... maybe.

build/bin.NOTICE
```diff
@@ -1,5 +1,5 @@
 Apache Daffodil VS Code Extension
-Copyright 2021 The Apache Software Foundation
+Copyright 2022 The Apache Software Foundation
```
stevedlawrence, line 2: This has already been fixed in main. Might get a merge conflict for this?

build/bin.NOTICE
```diff
@@ -77,9 +84,6 @@ Apache XML Commons Resolver (lib/xml-resolver.xml-resolver-<VERSION>.jar)
   Apache XML Commons Resolver
   Copyright 2006 The Apache Software Foundation.
 
-  This product includes software developed at
-  The Apache Software Foundation http://www.apache.org/
-
```
stevedlawrence, line 82: This line still exists in a few places. This can be removed in all places except for the very first one at the top.

build/bin.NOTICE
```diff
@@ -169,3 +166,71 @@ Apache Commons Lang (lib/org.apache.commons:commons-lang3-<VERSION>.jar)
...
+  Apache Commons Logging
+  Copyright 2003-2014 The Apache Software Foundation
+
+Apache HttpClient (lib/org.apache.httpcomponents.httpclient-<VERSION>.jar)
+  Apache HttpClient
+  Copyright 1999-2020 The Apache Software Foundation
+
+Apache HttpCore (lib/org.apache.httpcomponents.httpcore-<VERSION>.jar)
+  Apache HttpCore
+  Copyright 2005-2020 The Apache Software Foundation
```
stevedlawrence, line 196: We need this Http stuff since it is a dependency of Daffodil, but that's weird. Not sure why http stuff is a depdency of Daffodil. I wonder if that's an optional transitive dependency that we can remove in Daffodil 3.4.0?

build/bin.NOTICE
```diff
@@ -169,3 +166,71 @@ Apache Commons Lang (lib/org.apache.commons:commons-lang3-<VERSION>.jar)
...
+
+Apache Log4j (lib/org.apache.logging.log4j.log4j-api-<VERSION>.jar, org.apache.logging.log4j.log4j-core-<VERSION>.jar)
```
stevedlawrence, line 198: This is already above. One of those can be removed.

build/bin.NOTICE
```diff
@@ -169,3 +166,71 @@ Apache Commons Lang (lib/org.apache.commons:commons-lang3-<VERSION>.jar)
...
+  Apache Log4j
+  Copyright 1999-2019 Apache Software Foundation
+
+  This product includes software developed at
+  The Apache Software Foundation (http://www.apache.org/).
+
+  ResolverUtil.java
+  Copyright 2005-2006 Tim Fennell
+
+  Dumbster SMTP test server
```
stevedlawrence, line 208: The bin.NOTICE file @tuxji just merged does not list Dumbster or some of these other things from Log4j. Did Log4j change their notice?

@tuxji, it would help if you could take a scan at this. You're most faimilar with all the recent changes.

Also, it's a little hard to just diff since the format has changed. Do we want to standardize all of our license/notice files to the same format?

stevedlawrence (changes requested):

build/bin.NOTICE
mbeckerle: Eventually I think a directory of License files, and a directory of Notices files would be better than this squashing of stuff together into a single license and notices file. It's very hard to verify that this includes everything it is supposed to, and does not include things it's not supposed to, as well as not knowing what's verbatim we're including, and what is stuff we're writing.

stevedlawrence: Yeah, we originally just copied from what other projects do, but it should be fine to have a license directory, and is definitely easier to manage. However, I'm not sure you can do the same for NOTICES. I think those have to be in the NOTICE file.

build/bin.LICENSE
```diff
@@ -2207,3 +2207,537 @@
       THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
       (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
       THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+  This product bundles 'Saxon-HE (Home Edition)', including the following files:
```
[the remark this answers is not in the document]
tuxji: Actually we can remove the Saxon license because daffodil-debugger does not depend on daffodil-schematron, and therefore does not include Saxon or its transitive dependencies in the daffodil-debugger zip file.  It's important to build the extension and then look inside daffodil-debugger.zip to see which jars daffodil-debugger actually uses instead of simply starting with a copy of Daffodil's bin.LICENSE and bin.NOTICE.  I was surprised how many differences I found between Daffodil's libs and daffodil-debugger's libs once I actually compared the libs.  I don't simply see a superset or subset of one's libs in the other's libs; some dependencies even have different versions.  We only just released Daffodil 3.3.0, so there hasn't been enough time for Daffodil to bump some dependency versions to newer versions yet.  Someone will have to look closely at how daffodil-debugger specifies its dependencies and figure out why certain dependencies have different versions.

P.S.  I know why the versions are different now - when I built the extension, I'd checked out the main branch but the main branch is still using Daffodil 3.2.1, not 3.3.0.

build/bin.LICENSE
```diff
@@ -2207,3 +2207,537 @@
...
+    - lib/net.sf.saxon.Saxon-HE-<VERSION>.jar in in daffodil-debugger-<VERSION>.zip
+  These files are available under the MPL 2.0 license:
+
+    Most of the open source code in the Saxon product is governed by the Mozilla Public
```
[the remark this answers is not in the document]
tuxji: Saxon is gone now, but I've rearranged the lines mentioning each jar and its license to make it easier to understand which files are covered by which licenses.

build/bin.NOTICE
tuxji: I've removed these lines in my edits.

build/bin.NOTICE
tuxji: Yes, I removed the dependencies that do not have a NOTICE file in my Daffodil PR's final changes as Steve suggested.  I'll  upload my suggested edits directly to this PR as well.

build/bin.NOTICE
tuxji: I removed the http notices because they're not used by daffodil-debugger.  They're in Daffodil's bin.NOTICE only because daffodil-schematron uses Saxon-HE, Saxon-HE uses xml-resolver, and xml-resolver uses these http jars.  You have to build the extension and look at daffodil-debugger's libs as the source of truth before you edit bin.LICENSE or bin.NOTICE.

build/bin.NOTICE
tuxji: Done.

build/bin.NOTICE
tuxji: Aye, I took a pass at this and updated bin.LICENSE and bin.NOTICE.  I noticed that many jars don't have LICENSE or NOTICE files in them but I didn't take the extra time to visit their source repositories and check for LICENSE/NOTICE files there as well, so I don't promise that my edits are the last word.  I just want to give these files a better foundation for the 1.0.0 release and let future PRs improve them.

tuxji (commented):
I've done a pass over both bin.LICENSE and bin.NOTICE to bring them up to date and I will upload my changes to these files using "Edit File" in GitHub's PR user interface so all of you can review my suggested changes.  I made too many changes to simply add comments suggesting what to change in every place.    

In general, I made it easier to see exactly which files are covered by which licenses/notices with better formatting & wording and I sorted all the licenses/notices alphabetically by their jar names to establish a standard ordering in each file.  I'll upload my edits for review next.

tuxji (approved):
+1

If these bin.LICENSE/bin.NOTICE changes look OK, then please rebase, squash to one commit, and merge.

Shanedell:
@mbeckerle @stevedlawrence Mike and Steve, I think the changes added by @tuxji  look good to me, what do you guys think? Want to make sure to get your input before rebasing, squashing and merging.

mbeckerle:
+1 I'm good with these changes.

stevedlawrence:
+1, looks good

PR #108, opened 2022-05-02
Remove unused stopOnEntry options

### LINKED ISSUE #59: stopOnEntry doesn't seem to work
I have set stopOnEntry to false, but the debugger still seemed to stop, not to parse the data.

michael-hoke:
Closes #59 

I'm not sure if this is how we want to fix this issue due to a rebuild being required to change the behavior. Of the 3 stopOnEntry booleans, this was the only one that changed the behavior of the debugger. I also tried the inverse - removing this option, and setting the remaining options to false, and the debugger still stopped.

Note that a rebuild of the vsix file is required in order to pick the changes up.
From the root directory of the repo:
`yarn build`
`code --install-extension apache-daffodil-vscode-1.0.0.vsix`

### DIFF
--- a/package.json
+++ b/package.json
@@ -227,11 +227,6 @@
                   "path": "${workspaceFolder}/infoset.xml"
                 }
               },
-              "stopOnEntry": {
-                "type": "boolean",
-                "description": "Automatically stop after launch.",
-                "default": true
-              },
               "trace": {
                 "type": "boolean",
                 "description": "Enable logging of the Debug Adapter Protocol.",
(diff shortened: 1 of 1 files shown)


arosien:
Oops, didn't see this, let me look at the code to see what might be going on.

arosien (changes requested):
I don't think we should remove the feature as a way to fix it.

Shanedell:
This issue is not able to reproduced. @michael-hoke or @arosien  or myself will leave documentation inside of issue #59

PR #109, opened 2022-05-04
Fix lints

scholarsmate:
Fix a couple of SonarLint problems.


scholarsmate:
Yeah, it was an accidental pull request from our down stream repo(its happened a few times).  We're working on v2 milestones with v1.3.0 in flight.  We're managing access, issues, and things there and making sure we're merge-ready with merging changes from this upstream repo.  If you feel this strategy is ill advised, let's discuss it.  Thanks Steve.

stevedlawrence:
Personally, I'd rather see this work done in the daffodil-vscode repo. If you don't think it can be done in an incremental way in the main branch, as an alternative we could create a dev branch and merge PR's into that branch. Then when the changes are ready for main, we can just merge it in. This way most of the development happens in a public and more easily reviewable manner. It also means if there are any other people checking out the project, they are more likely to see this work and potentially get involved. It also avoids any potential intellectual property clearance stuff, though I'm not positive if that would be required or not.

This is just a personal preference though. It's ultimately up to the community how to handle this work. If everyone else is fine with it where it is and is okay with merge it in later in a big PR, that's fine with me.

scholarsmate:
I don't disagree with any of that, as it's both reasonable, and pragmatic.  Our little group is focused on a 1.3.0 milestone that introduces some basic hex-editing capabilities to the debugger.  Once that's ready, I'd feel better about migrating over.  It would be great to get more visibility and engagement, but we're striving to provide something of reasonable demonstrable value (e.g., not completely busted or half baked) before throwing it over the fence.  I'm very much in favor of small, working, increments over massive PRs.  I'm sure we can converge on a good, productive path forward.  For engaging with the community, shall I present what we're working on to the dev mailing list, or a discussion here on the project GitHub, or somewhere else?  Thanks again Steve.

PR #110, opened 2022-05-06
Address CI issues

### LINKED ISSUE #62: Converge CI Files
Instead of having everything action for the CI being a seperate file/action we want to converge into one file called `CI.yml`. 

However, with this we will still want to leave `release.yml` as its own separate file as it will require a different matching solution than all the other CI.

So take these files
```
.github/workflows/formatting.yml
.github/workflows/one-commit.yml
.github/workflows/rat-check.yml
.github/workflows/tests.yml
```
and move them to be one file, `.github/workflows/CI.yml`

### LINKED ISSUE #84: Yarn package never runs in CI
`yarn package` should be run in CI.

#81 and #82 were both introduced in #72 where the CI was all green.

### LINKED ISSUE #89: Scala code not being built or tested during CI
Only the Typescript code is compiled and tested, the Scala code is only checked for formatting. Add compilation, testing, and packaging checks.

michael-hoke:
- Add scala built/test/package - #89
- Add yarn package - #84
- Converge CI Files = #62

Closes #89 
Closes #84 
Closes #62


tuxji (approved):
+1

CI still works and the CI logs look good.

I noticed, however, that daffodil-vscode has not been receiving dependabot PRs since no one has created the file .github/dependabot.yml yet.  Please consider creating this file in the future since dependabot will keep your CI.yml's github action versions up to date (and if you configure them), your sbt and yarn dependencies up to date as well.

<https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/>
<https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file>
<https://github.com/samsmithnz/Dependabot-Configuration-Builder>

Shanedell:
I have created and assigned #111 to myself for adding dependabot to the CI. @tuxji For this repo we would need it to check npm/yarn packages and github-actions correct? Based on the file https://github.com/apache/daffodil/blob/main/.github/dependabot.yml I assume Scala isn't supported currently.

.github/workflows/CI.yml
```diff
@@ -14,41 +14,79 @@
 # limitations under the License.
 
 ---
-name: Code Formatting
+name: Daffodil CI
```
Shanedell, line 17: Change this to be `Daffodil VS Code CI`

.github/workflows/CI.yml
```diff
@@ -14,41 +14,79 @@
...
+              pull_number: context.issue.number,
+            });
+            core.info("Number of commits in this pull request: " + commits.data.length);
+            if (commits.data.length > 1) {
+              core.setFailed("If approved with two +1's, squash this pull request into one commit");
+            }
 
-  scala-format:
-    name: Scala Formatting
+  format-rat-check:
```
Shanedell, line 43: Having the TypeScript and Scala formatting in one job I think is fine but the ratCheck should its own separate stage.

.github/workflows/CI.yml
```diff
@@ -14,41 +14,79 @@
...
+      - run: yarn lint
+      - name: Run Rat Check
+        run: $SBT ratCheck || (cat target/rat.txt; exit 1)
+      
+  build-test-package:
+    name: Build, Test, and Package
+    strategy:
+      matrix:
+        os: [macos-11, ubuntu-20.04, windows-2019]
+        node: [ '10', '12', '14', '16' ]
```
Shanedell, line 75: @stevedlawrence @mbeckerle Do you think the amount of versions for node used could be lowered? I was thinking maybe just the hightest and lowest, so `[ '10', '16']` so the amount of time this job is ran goes from 12 to 6.

Shanedell (changes requested):

tuxji:
> @tuxji For this repo we would need it to check npm packages and github-actions correct? Based on the file https://github.com/apache/daffodil/blob/main/.github/dependabot.yml I assume Scala isn't supported currently.

Yes, correct. We have to use Scala Stewart to keep sbt dependencies up to date.  See https://www.scala-lang.org/blog/2019/07/10/announcing-scala-steward.html.

.github/workflows/CI.yml
stevedlawrence: I would actually just suggest `CI`. We did this for Daffodil because in some places GitHub displays jobs like `Workflow Name / Job Name`. Sometimes GitHub doesn't proide much space for this, so it cuts off the job name, which is the part that actually matters. So just changing to "CI" ensures the job is always visible.

.github/workflows/CI.yml
stevedlawrence: Agreed.

.github/workflows/CI.yml
stevedlawrence: Seems reasonable to me, but I'm not that familiar with node.

Are there any compatibility issues that might need to be checked for those other versions of node?

Does VS Code ship with a specific version of node that our extension needs to work with? Or do we just bundle it with the extension? If we bundle, can we just test with that specific version? Does the yarn.lock file specify a version of node?

.github/workflows/CI.yml
```diff
@@ -14,41 +14,79 @@
...
-  scala-format:
-    name: Scala Formatting
+  format-rat-check:
+    name: TypeScript/Scala Formatting and Rat Check
     runs-on: ubuntu-20.04
     defaults:
       run:
         shell: bash
     env:
       SBT: sbt -J-Xms1024m -J-Xmx5120m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m ++2.12.13
```
stevedlawrence, line 50: I believe the newest version of Daffodil needs scala 2.12.15. It will work with 2.12.13, but there are some edge cases (I think related to serialized parses?) that require the latest version of Scala.

.github/workflows/CI.yml
```diff
@@ -14,41 +14,79 @@
...
+      - uses: actions/checkout@v2.3.5
+      - name: Install Node.js
+        uses: actions/setup-node@v1.4.6
+        with:
+          node-version: ${{ matrix.node}}
+      - run: yarn build
+      - run: yarn test
+      - run: $SBT test
+      - run: yarn package
+      - run: $SBT package
```
stevedlawrence, line 92: Looks like `yarn package` is new to the CI and it runs `vsce package`. Is that any different than `yarn build`?

Also, `yarn build` will run `sbt universal:packageBin`, which will runs `sbt package`, so I think the sbt package line is redundant.

stevedlawrence (changes requested):

.github/workflows/CI.yml
Shanedell: I think it all depends on the version of VS Code you have. For example, mine says it uses nodejs 16 but someone could not be as up to date so it uses nodejs 14 so maybe just keep the two latest version to test?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Setup Java
+        uses: actions/setup-java@v2.3.1
+        with:
+          distribution: temurin
+          java-version: 11
+      - run: $SBT scalafmtCheck
+      - run: $SBT scalafmtSbtCheck
+      - run: yarn install
+      - run: yarn lint
+      - name: Run Rat Check
```
Shanedell, line 67: @michael-hoke You can remove this line from this job since you do it in a separate one now

Shanedell (approved):

Shanedell (approved):
+1 @stevedlawrence Would like to get your input if anything needs added before I would merge this in.

stevedlawrence:
+1 looks good to me. Once squashed this should be good to merge.

scholarsmate (approved):
+1 LGTM

stevedlawrence (approved):

PR #112, opened 2022-05-10
Add debug option that uses configuration from last debug

### LINKED ISSUE #14: Remember schema and data selection
An optional remember this selection would be handy on the pickers. That would remove the need to edit the launch json to prevent being prompted each time.

Perhaps there is an additional debug command, "debug last" or something that can more easily recall the settings without needing to have an explicit remember this option.

Original issue https://github.com/jw3/example-daffodil-vscode/issues/108

michael-hoke:
Closes #14 

Not sure if changing the type of the currentConfig variable is correct, but I couldn't find a reason that it would need to be a ProviderResult (and it eliminates the complication of needing to treat ProviderResult vs non-ProviderResult differently).

Also, I had considered renaming getCurrentConfg() to getCurrentConfig(). I'll go back and do that if that's what we want to do.


src/adapter/activateDaffodilDebug.ts
```diff
@@ -19,11 +19,15 @@ import { getDebugger, getDataFileFromFolder } from '../daffodilDebugger'
 import { FileAccessor } from './daffodilRuntime'
 import * as fs from 'fs'
 import * as infoset from '../infoset'
-import { getConfig, setCurrentConfig } from '../utils'
+import { getConfig, getCurrentConfg, setCurrentConfig } from '../utils'
```
mbeckerle, line 22: "Confg".... simply a typo that will waste lots of people's time under maintenance. Fix please.

Shanedell (approved):
+1 Mark this `Ready for Review`. After that rebase off main, fix conflicts, squash to one commit and force push to branch. Once these are done we can get this merged in.

Issue #100, opened 2022-04-04
README information shown in VS Code should be less build focused


stevedlawrence:
The README that is shown in the VS Code app is the one that is at the root of our repo, which is focused on how to build the vsix file from source. But since the extension is already installed, then this README information isn't very helpful. We already have different LICENSE and NOTICE files that go in our vsix file. Should we also have a different README file that is more focused on how to use the debugger instead of how to build it, making it more useful to people viewing the README in VS Code?

hdalsania:
@stevedlawrence  You are correct, PR#174 was not merged but Readme file is fixed to provide links to development and usage information wiki and instructions to download vsix file from marketplace with issue #411 and PR #424

PR #174, opened 2022-06-10
Added debugger usage info to the readme

rt320:


scholarsmate:
Seeks to address issue https://github.com/apache/daffodil-vscode/issues/100.

README.md
```diff
@@ -89,6 +89,21 @@ Once you have either downloaded the VSIX file or created it you can now install
...
+
+To step through the schema code, use the step over button
+![Step Over](./images/StepThrough.gif)
+
+To view the infoset, use the veiw the XML infoset button
+![View Infoset](./images/ViewInfoset.gif)
+
+To view the data file in hex, use the view hex button
+![View Hex](./images/ViewHex.gif)
+
```
stevedlawrence, line 106: These images are very wide, like 1800 pixels. When they are rendered in a github readme, e.g.

https://github.com/apache/daffodil-vscode/blob/f838c5c753e97aeed30c71db76d56396e34e40b3/README.md

They end up getting scaled down to fit the width and so we lose a lot of quality--images get blurry and bit hard to read. It looks like anything over 1012 pixels is going to get scaled. I'd suggest creating these with a smaller VS Code window so they fit the full width of a github page without needing to be scaled.

README.md
```diff
@@ -89,6 +89,21 @@ Once you have either downloaded the VSIX file or created it you can now install
 
 Please refer to the Wiki page at https://github.com/apache/daffodil-vscode/wiki
 
+To debug a daffodil schema, choose open a folder from the File menus, then select a dfdl.xsd file. Once the file loads in VS Code, then set a desired breakpoint.
+![Setting Breakpoints](./images/SetBreakpoint.gif)
+
+To start debugging, click on the debug button on the side panel, then click the RUN AND DEBUG button with the dropdown set to Ask for file name.
+![Debugging](./images/StartDebug.gif)
```
stevedlawrence, line 96: This GIF goes really fast, its hard to tell what it's actually trying to show. Does this add anything new that's not on the wiki?

README.md
```diff
@@ -89,6 +89,21 @@ Once you have either downloaded the VSIX file or created it you can now install
 
 Please refer to the Wiki page at https://github.com/apache/daffodil-vscode/wiki
 
+To debug a daffodil schema, choose open a folder from the File menus, then select a dfdl.xsd file. Once the file loads in VS Code, then set a desired breakpoint.
```
stevedlawrence, line 92: We have a link to the wiki which looks to show very similar things. I'm not sure we need them both in the wiki and here. The wiki seems to have more information.

stevedlawrence (commented):
My intention with #100 was to not to add usage information to this README, but to have a completely different README file that is installed into the vsix file, much like how we have completely different LICENSE files for the repo vs the vsix file.

I would argue the README in this repo should be focused on how to build the vsix, with links links to usage, as it currently is.

The README that is installed to the vsix file doesn't need that build information though, maybe it just has very basic information and then links to various pages in the wiki as those become more advanced and improved?

README.md
rt320: I'm not sure what you are saying, are you saying the wiki is all we need?  Or are you saying to move the wiki info to the readme? Or something else?

README.md
rt320: I can redo the GIFs, but as in the previous comment I'm not sure that is what you are asking for.

README.md
rt320: I captured the images on a laptop, so the images looked small to me.  My resolution on the laptop screen is 1920 x 1080.  Should I create the images at a lower resolution 800 x 600?

rt320: I'll remove the debugger use info from the original readme.  But I'll wait on your feedback to the other comments before proceeding with anything else.

README.md
stevedlawrence: I don't really have an opinion if the information lives in the wiki or lives in the repo. What I'm saying is we should not have the same information in both the wiki *and* the repo. That's too much of a maintenance burden. The information on the wiki seems more detailed, so I would suggest we just keep it there.

scholarsmate:
@rt320 @stevedlawrence Debugger usage is now documented in the project wiki.  Given that, can we close this PR out?

stevedlawrence:
It's fine with me to close this PR. Note that issue #100 is still not resolved, so that should stay open.

PR #187, opened 2022-06-17
Dependabot updates

Shanedell:
- Update node package and github-action versions based on dependabot findings.
- Update dependabot to ignore two node packages that cause the building of the code to break when updated.


.github/dependabot.yml
```diff
@@ -28,3 +28,6 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "typescript"
+      - dependency-name: "ts-loader"
```
stevedlawrence, line 33: Should we create an issue to update these so we don't forget? I don't think it's critical that these be updated so it'd be good to have a reminder somewhere.

package.json
```diff
@@ -62,17 +62,17 @@
...
     "ts-loader": "8.1.0",
     "typescript": "4.3.5",
-    "vsce": "1.87.1",
+    "vsce": "2.9.1",
     "vscode-debugadapter-testsupport": "1.46.0",
-    "webpack": "4.46.0",
-    "webpack-cli": "3.3.12"
+    "webpack": "5.40.0",
+    "webpack-cli": "4.10.0"
   },
```
stevedlawrence, line 76: Did you confirm that these new dev dependencies, and their transitive dependencies, are still ALv2 compliant according to ASF?

Since they are dev dependencies I'd assume that means no part of them ends up in `extension.js` or the .vsix file (is this true?) and so we don't need to update the `bin.{LICNESE,NOTICE} files, but they do still need to be ALv2 compatible since they are required to build.

Just so it's clear to everyone:
* packages listed in `dependencies` and their transitive dependencies *must* be listed in `bin.LICENSE` or `bin.NOTICE`, with the assumption they end up in the vsix file
* packages listed in "devDependencies" and their transitive dependencies *should not* be listed anywhere, which the assumptino that they don't end up in the vsix file
* all packages need to be ALv2 compatible

I think we do this correctly, but I'm not positive. Is there a way to list the LICENSE/NOTICE information of all dependencies and their transitives that *aren't* devDependencies?

Shanedell: I didn't mean to update the webpack items those were in part of testing the other two things that broke items I will revert those.

But yes your are correct the `devDependencies` do not end up in the VSIX but the items in `dependencies` do which is how we have everything setup. I also do know that all but one of the items under `devDependencies` are ALv2 compatible. Is the ISC license compatible with ALv2? `glob` (https://github.com/isaacs/node-glob) is the only package under `devDependencies` that uses something other MIT or AL

.github/dependabot.yml
Shanedell: Yeah I made this issue https://github.com/apache/daffodil-vscode/issues/188 to investigate

package.json
stevedlawrence: Does updating webpack break other things? I was asking more generally about all the new dev deps. If that updates without any problem I would just keep it as is.

Here's the list of license approved by ASF: https://www.apache.org/legal/resolved.html

ISC is Category A which is allowed. Note that we also need to verify the transitive dependencies of the devDependencies. It's possible that one of our devDependencies depends on something that ASF doesn't allow.

Shanedell: Nothing breaks with the current updates. All items are allowed under ALv2 but I am not sure about their attached deps so I will look into that

Shanedell: @stevedlawrence Pretty handy tool
```
yarn global add license-checker
```
then you can run
```
license-checker --summary
```
Here's the result
```
├─ MIT: 253
├─ ISC: 38
├─ BSD-3-Clause: 20
├─ BSD-2-Clause: 16
├─ Apache-2.0: 11
├─ MIT*: 2
├─ Python-2.0: 1
├─ Custom: http://github.com/substack/node-bufferlist: 1 # <-- MIT
├─ CC-BY-4.0: 1
├─ (MIT OR WTFPL): 1
├─ (BSD-2-Clause OR MIT OR Apache-2.0): 1
└─ 0BSD: 1
```
These all are approved correct? Only one I wasn't sure about was `0BSD`

stevedlawrence: The ASF licensing policy says BSD and variants are acceptable. Though, according to https://opensource.org/licenses/0BSD, 0BSD is basically ISC. Either way, 0BSD should be fine.

Sounds good then. No concerns with any of the licenses.

Shanedell:
@stevedlawrence I upgraded the version of a couple more packages, licenses didn't change, just in case you wanted to take another quick look

stevedlawrence:
Updates look fine to me.

stevedlawrence:
Latest updates look good. I've confirmed it works with https://github.com/apache/daffodil/pull/798

Shanedell:
@stevedlawrence You said to wait for two approval for this and the scala steward one correct?

I am working on a new way of handling the building it is similar to what we do now but uses node instead of trying to run commands for everything so it will function nicer, I planned on putting this JS file inside of `build`. Will it be okay if I refactor some of the code in that PR as well? Basically it will split out the `Artifact` class as it used in two different files, then it will also remove the `src/omega_edit/download.ts` file to make a static JS file in build, `build/omega_edit.js` but will function the exact same way. If you don't like the JS files going into `build`, another way I was thinking of was making a folder `src/scripts/` for both of these new JS files to go into, just let me know which you prefer.

stevedlawrence:
Yeah, 2 +1's are required before this can be merged. Hopefully one of our many commiters can find some time to take a look.

> I am working on a new way of handling the building

Switching to node and your refactoring ideas sound very reasonable to me.

I have no strong preference about `build` vs `src/scripts`.

That said, my first thought is that it feels natural for only code that is part of the extension to go in `src/`, so any build related scripts that aren't directly part of the extension want to be somewhere else. For example, in sbt, all build related files are in `build.sbt` or `project/`, and `src/` is where the actual scala files go. So that might be an argument to have the scripts in `build/` or something.

And if that's the case, maybe it makes sense to organize the stuff in build a bit diferently, e.g. `build/scripts` and `build/vsix` to make it clear what are just build scripts and what are files are for the vsix extension? Or maybe `src/` wants to become `extension/src`, and then the stuff currently in `build/` can go in `extension/resources/`, and then build related scripts go in `build`? Lots of options.

Another related thought, though maybe not a good one, does it make sense to rewrite the build config in sbt? It looks like there are sbt plugins for things related to what we do:

https://github.com/jokade/sbt-node
https://github.com/stonexx/sbt-webpack

Those look fairly old, so maybe they aren't maintained and it's not a good idea. But it would be nice to have only a single build system. And as difficult as sbt can be, it is fairly powerful. From what I've seen from yarn, it's pretty limited. Though, maybe the node changes you've suggested resolve those issues, and in a better way than old sbt plugins would.

arosien:
+1

arosien (approved):
+1

PR #199, opened 2022-06-23
Add Scala Steward GitHub Action

### LINKED ISSUE #185: Enable Scala Steward dependency checking/updates
Dependabot does not know about Scala/SBT dependencies or how to update them. For that, [Scala Steward](https://github.com/scala-steward-org/scala-steward) can be used. We should open a pull request to the Scala Steward project to add apache/daffodil-vscode to repos-github.md so to enabble dependency updates for the scala code in this repo.

Shanedell:
- Currently having it running at 2pm everyday so I could test it a bit before merging. This will be change to only run on Sunday mornings every week.

Closes #185

### DIFF
--- a/.github/workflows/scala-steward.yml
+++ b/.github/workflows/scala-steward.yml
@@ -1,0 +1,34 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This workflow will launch at 00:00 every Sunday
+# This file is based on docs from https://github.com/scala-steward-org/scala-steward-action
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+name: Scala Steward CI
+
+jobs:
+  scala-steward:
+    if: github.repository == 'apache/daffodil-vscode'
+    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/scala-steward.yml
```diff
@@ -0,0 +1,18 @@
...
+on:
+  schedule:
+    - cron: '0 12 * * *'
+
+name: Scala Steward CI
+
+jobs:
+  scala-steward:
+    runs-on: ubuntu-20.04
+    name: Launch Scala Steward
```
stevedlawrence, line 12: Should we add an `if` condition so that this is only triggered on the `apache/daffodil-vscode` repo? If someone creates a fork of the repo we probably don't want their fork to be spammed with PR's updating dependencies. Dependabot does that right now with no way to disable it and I find it pretty annoying.

I think something like would work:
```yaml
jobs:
  scala-steward:
    if: github.repository == 'apache/daffodil-vscode'
    ...
```

.github/workflows/scala-steward.yml
```diff
@@ -0,0 +1,18 @@
+# This workflow will launch at 00:00 every Sunday
```
stevedlawrence, line 1: Needs an apache license to pass the rat check.

.github/workflows/scala-steward.yml
Shanedell: Makes sense to me I added this in

.github/workflows/scala-steward.yml
Shanedell: Yeah missed that initially, I added it

stevedlawrence (approved):
+1 :+1: 

I imagine we can't really test this until it's merged? But it seems pretty straightforward and not a big deal if we need to adjust things later.

Shanedell:
Yeah I am going to make an initial commit to have it run everyday so we can get a feel for how it works but after we have it figured out and most of the stuff is updated I would like to move it to be only once a week no Sundays or something. That way for both Node and Scala updates we can get them squared away on Mondays

arosien:
+1

Shanedell:
@arosien Can you:
- go to `Files Changed`
- click `Review Changes`
- select `Approve`
- type `+1` as the comment
- clci `Submit Review`

That will make sure we get the green checkmark that you approved the changes of the PR. I believe you approved the other PR as well so you will need to do this for that one as well, thank you

arosien (approved):
+1

PR #204, opened 2022-06-28
Dependabot Updates

Shanedell:
Dependabot Updates:

- Only run weekly instead of daily
- Fix ignore entries
- Update mocha to 9.1.1
- Fix issue with xdg-apps-path (tsconfig.json)


stevedlawrence (approved):
+1

arosien (approved):
+1

PR #206, opened 2022-07-06
Build Script & General updates

### LINKED ISSUE #121: Create better build setup
The current way we build the VSIX file, by moving the `LICENSE` and `NOTICE` files to temp files, `build/bin.NOTICE` and `build/bin.LICENSE` to `LICENSE` and `NOTICE` then moving things back is not the cleanest.

This ticket is to come up with a better way to bundle the `build/bin.NOTICE` and `build/bin.LICENSE` files inside of the VSIX as the `NOTICE` and `LICENSE` files.

- This ticket stems from this comment https://github.com/apache/daffodil-vscode/pull/117#discussion_r883590704

Shanedell:
Multiple Updates:

- Create node script to handle pre and post build processes instead of using OS commands
- Move location of script to download omega_edit. Also made it a static script instead of one that is compiled.
- Remove run-script-os
- Create artifact class

Closes #121

### DIFF
--- a/build/scripts/omega_edit_download.ts
+++ b/build/scripts/omega_edit_download.ts
@@ -15,40 +15,38 @@
  * limitations under the License.
  */
 
-import * as fs from 'fs'
-import * as path from 'path'
-import * as os from 'os'
-import { HttpClient } from 'typed-rest-client/HttpClient'
-import * as crypto from 'crypto'
+// @ts-nocheck <-- This is needed as this file is basically a JavaScript script
+//                 but with some TypeScript niceness baked in
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+const HttpClient = require('typed-rest-client/HttpClient').HttpClient
+const crypto = require('crypto')
 const hashSum = crypto.createHash('sha512')
 
-// Method to get omega-edit version from a JSON file
-export function getOmegaEditPackageVersion(filePath: fs.PathLike) {
-  return JSON.parse(fs.readFileSync(filePath).toString())['dependencies'][
-    'omega-edit'
-  ]
+/**
+ * The below classes (Backend, Artifact) are defined similarly in src/classes/artifact.ts.
+ * However for this script to work properly they needed to be redefined in a more JS style.
+ */
+
+class Backend {
+  constructor(owner, repo) {
+    this.owner = owner
+    this.repo = repo
+  }
 }
 
-// Method to get omegaEditServerHash from a JSON file
-export function getOmegaEditServerHash(filePath: fs.PathLike) {
-  return JSON.parse(fs.readFileSync(filePath).toString())['omegaEditServerHash']
-}
-
-export class Backend {
-  constructor(readonly owner: string, readonly repo: string) {}
-}
-
-export class Artifact {
-  constructor(readonly omegaEditVersion: string) {}
-
-  name = `omega-edit-scala-server-${this.omegaEditVersion}`
-  archive = `${this.name}.zip`
-  archiveUrl = (backend: Backend) =>
-    `https://github.com/${backend.owner}/${backend.repo}/releases/download/v${this.omegaEditVersion}/${this.archive}`
-
-  scriptName = os.platform().toLowerCase().startsWith('win32')
-    ? 'example-grpc-server.bat'
-    : './example-grpc-server'
+class Artifact {
+  constructor(omegaEditVersion) {
+    this.omegaEditVersion = omegaEditVersion
+    this.name = `omega-edit-scala-server-${this.omegaEditVersion}`
+    this.archive = `${this.name}.zip`
+    this.archiveUrl = (backend) =>
+      `https://github.com/${backend.owner}/${backend.repo}/releases/download/v${this.omegaEditVersion}/${this.archive}`
+    this.scriptName = os.platform().toLowerCase().startsWith('win32')
+      ? 'example-grpc-server.bat'
+      : './example-grpc-server'
+  }
 
   getOsFolder() {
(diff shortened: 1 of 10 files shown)


yarn.lock
```diff
@@ -36,9 +36,9 @@
     "@jridgewell/trace-mapping" "^0.3.9"
 
 "@jridgewell/resolve-uri@^3.0.3":
-  version "3.0.8"
-  resolved "https://registry.yarnpkg.com/@jridgewell/resolve-uri/-/resolve-uri-3.0.8.tgz#687cc2bbf243f4e9a868ecf2262318e2658873a1"
-  integrity sha512-YK5G9LaddzGbcucK4c8h5tWFmMPBvRZ/uyWmN1/SbBdIvqGUdWGkJ5BAaccgs6XbzVLsqbPJrBSFwKv3kT9i7w==
+  version "3.1.0"
```
mbeckerle, line 39: Should yarn.lock be checked in at all? Should we be ignoring it in .gitignore?

build/scripts/process_build.ts
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-nocheck <-- This is needed as this file is basically a JavaScript script
+//                 but with some TypeScript niceness baked in
+const fs = require('fs')
+const execSync = require('child_process').execSync
+
+function prebuild() {
```
mbeckerle, line 23: What is this doing? Add comment that this "Combines license and notice information into the bin.License or whatever so that we don't have to maintain the bin.NOTICE manually. Assuming I've guessed correctly why we need this. 

Why are all these calls "Sync" variants? What's happening that requires that?

mbeckerle (commented):
I can't approve this as yet. I simply don't understand the motivation for this code.

It seems to be doing a lot of build stuff that one would hope that we wouldn't even have to build our own scripts for. 

Did I miss comments to the effect of "This script is needed because maven/sbt won't do xyz for us" ?

build/scripts/process_build.ts
Shanedell: So what this script is doing is replacing the OS specific commands to move `bin.LICENSE` and `bin.NOTICE` to be `LICENSE` and `NOTICE` before packaging the VSIX, then reverting them back once the VSIX is done.

So the sync variants just ensure that nothing else is going to happen before this item completes its tasks.

build/scripts/omega_edit_download.ts
```diff
@@ -15,40 +15,38 @@
...
-  scriptName = os.platform().toLowerCase().startsWith('win32')
-    ? 'example-grpc-server.bat'
-    : './example-grpc-server'
+class Artifact {
+  constructor(omegaEditVersion) {
+    this.omegaEditVersion = omegaEditVersion
+    this.name = `omega-edit-scala-server-${this.omegaEditVersion}`
+    this.archive = `${this.name}.zip`
+    this.archiveUrl = (backend) =>
+      `https://github.com/${backend.owner}/${backend.repo}/releases/download/v${this.omegaEditVersion}/${this.archive}`
```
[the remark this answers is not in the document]
Shanedell: So we can't do that because the `omega-edit` scala server is only packaged using `universal:packageBin`, like we do with the bugger, so creates a zip with the needed scripts and JAR files. Moreover, this ZIP not published to any maven/sbt repos it is only published to the GitHub release page. So this file is not building anything necessarily it is only downloading the `omega-edit` scala server from the github releases.

yarn.lock
Shanedell: Yes it should be checked in. The `yarn.lock` ensures that no matter who pulls this code when they run `yarn install` everybody would get the same version of the packages. In a handful of places in the `package.json` we need to use `^VERSION` meaning anything above and including that specified version can be used.

Without the `yarn.lock` the items that don't have a strict version can change depending on when the `yarn install` command is ran. Eg on 5/6 John Smith cloned the repo, ran `yarn install` and got version `7.1.3` of `@types/glob`. However, later that day version `7.1.4` was released to npmjs. On 5/7 Jane Doe cloned the repo, ran `yarn install` and got version `7.1.4` of `@types/glob` but for this individual something broke in `daffodil-vscode` from the changes made in `@types/glob` between `7.1.3` and `7.1.4`.

build/scripts/omega_edit_download.ts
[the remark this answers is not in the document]
mbeckerle: I took a look and I didn't find exactly the thing I wanted which is sbt dependencies that look for github releases. You can do source code dependencies, but not binary dependencies.

mbeckerle (approved):
+1

build/scripts/process_build.ts
mbeckerle: I guess I'd like to see comments about why we need this contortion? I.e., what are we doing here? Why do we need LICENCE and NOTICE in the VSIX (is that where vscode expects to find? is this an Apache requirement that packaged releases like the vsix have these  and why the bin.LICENSE, and bin.NOTICE naming convention?  What does that achieve for us?

This is probably also done in Daffodil libraries, I just don't know that any of these requirements are written down anywhere which is my frustration. The scripting always seems fragile and unmotivated to me, as it is addressing a bunch of unstated requirements. 

I'll think about creating a JIRA ticket to document these build-an-artifact requirements for LICENSE/NOTICE and how we make that work in a modular fashion, etc. 

No action required for this comment. Just thinking aloud.

scholarsmate:
Is this just waiting on another +1?

mbeckerle:
Yes, I think just waiting for a 2nd reviewer.

arosien (approved):
👍

PR #219, opened 2022-07-11
Update scala dependencies

Shanedell:
This PR is simply to make the review of the Scala dependency update easier. 

fs2-io was not updated as with its latest version we will need to transfer from using `java.nio` for everything path related to use `fs2.io`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ import play.api.libs.json._
 lazy val packageData = Json.parse(scala.io.Source.fromFile("./package.json").mkString).as[JsObject]
 lazy val daffodilVer = packageData("daffodilVersion").as[String]
 
-lazy val commonSettings = {
+lazy val commonSettings =
   Seq(
     version := {
       val versionRegex = raw"""  "version": "(.*)",""".r
@@ -39,14 +39,15 @@ lazy val commonSettings = {
     ),
     licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
     organization := "org.apache.daffodil",
+    // scala-steward:off
     scalaVersion := "2.12.15",
+    // scala-steward:on
     scalacOptions ++= Seq("-Ypartial-unification"),
     // remove the -Xcheckinit option added by the sbt tpoletcat plugin. This
     // option leads to non-reproducible builds
     scalacOptions --= Seq("-Xcheckinit"),
     startYear := Some(2021)
   )
-}
 
 lazy val ratSettings = Seq(
   ratLicenses := Seq(
@@ -77,10 +78,10 @@ lazy val core = project
     name := "daffodil-debugger",
     libraryDependencies ++= Seq(
       "ch.qos.logback" % "logback-classic" % "1.2.11",
-      "com.microsoft.java" % "com.microsoft.java.debug.core" % "0.35.0",
+      "com.microsoft.java" % "com.microsoft.java.debug.core" % "0.38.0",
       "co.fs2" %% "fs2-io" % "3.2.10",
-      "com.monovore" %% "decline-effect" % "2.2.0",
-      "org.typelevel" %% "log4cats-slf4j" % "2.1.1"
+      "com.monovore" %% "decline-effect" % "2.3.0",
+      "org.typelevel" %% "log4cats-slf4j" % "2.3.2"
     ),
     buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion, "daffodilVersion" -> daffodilVer),
     buildInfoPackage := "org.apache.daffodil.debugger.dap",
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -39,14 +39,13 @@ lazy val commonSettings = {
     ),
     licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
     organization := "org.apache.daffodil",
-    scalaVersion := "2.12.15",
+    scalaVersion := "2.12.16",
```
mbeckerle, line 42: I think we should not do this update.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -84,8 +84,8 @@ object DAPSession {
       dispatcher.unsafeRunSync {
         for {
           _ <- Logger[IO].info(show"R> $request")
-          _ <- requests.offer(Some(request)).recoverWith {
-            case t => Logger[IO].error(t)(show"error during handling of request $request")
+          _ <- requests.offer(Some(request)).recoverWith { case t =>
```
mbeckerle, line 87: If this is just a whitespace change, can we revert

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -194,7 +194,7 @@ class DAPodil(
           case Left(errors) =>
             state.set(DAPodil.State.FailedToLaunch(request, errors, None)) *>
               Logger[IO].warn(show"error parsing launch args: ${errors.mkString_(", ")}") *> session
-              .sendResponse(request.respondFailure(Some(show"error parsing launch args: ${errors.mkString_(", ")}")))
+                .sendResponse(request.respondFailure(Some(show"error parsing launch args: ${errors.mkString_(", ")}")))
```
mbeckerle, line 197: ditto. Revert just whitespace changes because they cause trouble trying to understand the version changes which is what this PR is about.

mbeckerle: all the changes in this file appear to be just whitespace

src/daffodilDebugger.ts
```diff
@@ -204,7 +204,8 @@ export async function getDebugger(
         name: artifact.scriptName,
         cwd: `${rootPath}/daffodil-debugger-${daffodilVersion}-${LIB_VERSION}/bin/`,
         hideFromUser: false,
-        shellPath: artifact.scriptName,
+        shellPath: '/bin/bash',
```
mbeckerle, line 207: This seems to be a change separate from the updates of the dependencies?

mbeckerle (approved):
+1

I'd prefer if this was ONLY changes to files that have dependency version information. There's some other changes in here as well. 

That said the other changes appear to be benign enough, so that's just a suggestion.

build.sbt
Shanedell: I will look into seeing about ignoring this via scala-steward

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
Shanedell: Same issue https://github.com/apache/daffodil-vscode/pull/219#discussion_r918337631 will look into ignoring spacing items

mbeckerle: Yeah, if you can just pull these from this commit that would be good. A subsequent commit of just code-formatting fixups is fine of course, particularly if it brings the code into coherence with a standard formatting.

Shanedell: @mbeckerle It seems scalafmt doesn't allow us just to ignore certain lines. So I think either the PR has to have to formatting fixes in it or I revert the formatting stuff and make separate PR for it.

arosien: See https://scalameta.org/scalafmt/docs/configuration.html#-format-off for a solution

arosien:
Even though the current Scala Steward doesn't seem to kick off the CI checks, I'd still prefer the individual PRs per dependency upgrade.

Shanedell:
> Even though the current Scala Steward doesn't seem to kick off the CI checks, I'd still prefer the individual PRs per dependency upgrade.

@arosien After this PR is merged that is how it will be handled from now on. I just an encompassing PR since we there were so many keeping up with them not will be less. So this is basically a one time same thing we did with dependabot.

arosien (approved):
👍

PR #223, opened 2022-07-13
Update issue inside of artifact.ts

Shanedell:


Shanedell:
@mbeckerle I added a change to fix the a checkbox for the omega-edit UI not sure if that changes your +1 just wanted to make sure you were aware that change will be in this merge as well.

mbeckerle (approved):
+1

stevedlawrence (approved):
+1

Minor nit on the git commit message, "Update issue" isn't very descriptive. It would be much more helpful to reviewers and people looking at past commits if you described what the actual issue is, e.g. `Fix special case comparison for omega-edit download url` or something. Adding an explanation of why this is needed is also helpful. It's not clear to me what "type" normally is and why a straight === comparison doesn't work.

That's all maybe a bit verbose, but I've never complained about having too much information in a commit message about what's going on.

PR #232, opened 2022-07-18
Update build script.

### LINKED ISSUE #121: Create better build setup

Shanedell:
Setup a new build script.

Closes #121


build/scripts/process_build.ts
```diff
@@ -18,39 +18,81 @@
...
+
+    if (file.includes('zip')) {
+      var dir = getDir(file)
+
+      fs.readdirSync(dir).forEach((dirFile) => {
+        if (dirFile.includes('zip')) {
+          file = `${dir}/${dirFile}`
+        }
+      })
+    }
```
stevedlawrence, line 77: This code seems a bit fragile. For example, it removes `*`'s which are useful since some of the lines in .vscodeignore are globs. Also, the way it handles zip is fragile since it just copies all zips from a directory? I think because we use globs for zip files in the .vscodeignore file?

What are you thoughts on this approach, where the key differences is 1) we only look at lines that start with an exclamation, and 2) those lines are treated as globs using the glob package:

```node
let lines = fs.readFileSync('.vscodeignore').toString().split('\n')
for (var i = 0; i < lines.length; i++) {
  let line = lines[i]
  // ignore any lines that don't start with an exclamation
  if (!line.startsWith("!")) continue
  // remove the exclamation
  let pattern = line.substring(1)
  // find all files matching the line glob and copy into dist/package/
  glob(pattern, (error, files) => {
    for (var j = 0; j < files.length; j++) {
      let file = files[j]
      // code here from below that creates directories and copies file/dir
    }
  })
```

This isn't tested, and I'm sure there's some node issues, but something along those lines?

build/scripts/process_build.ts
```diff
@@ -18,39 +18,81 @@
...
+    let directory = fs.statSync(file).isFile() ? getDir(file) : file
+
+    fs.mkdirSync(`${pkg_dir}/${directory}`, { recursive: true })
+
+    fs.statSync(file).isFile()
+      ? fileCopy(file, `${pkg_dir}/${file}`)
+      : dirCopy(file)
   }
+
+  fileCopy('.vscodeignore', `${pkg_dir}/.vscodeignore`)
```
stevedlawrence, line 88: Do we also need to copy bin.NOTICE and bin.LICENSE files?

build/scripts/process_build.ts
```diff
@@ -18,39 +18,88 @@
...
+  execSync('yarn install', { cwd: pkg_dir })
+  execSync('yarn package', { cwd: pkg_dir })
```
stevedlawrence, line 90: yarn package just calls vsce, thoughts on just directly calling vsce ourselves here and removing the "yarn package script? Is there ever a case where we would want to run "yarn package" without all thise new copy stuff?

vsce package also has a `--out` parameter which could be used instead of having to call get_vsix? Then `yarn build` just becomes this:

```
"build": "node -e \"require('./build/scripts/process_build.ts').build_package()\"",
```

and it simplifies the package.json a bit?

build/scripts/process_build.ts
```diff
@@ -18,39 +18,88 @@
...
+
+    fs.mkdirSync(`${pkg_dir}/${directory}`, { recursive: true })
+
+    fs.statSync(file).isFile()
+      ? fileCopy(file, `${pkg_dir}/${file}`)
+      : dirCopy(file)
   }
+
+  fileCopy('.vscodeignore', `${pkg_dir}/.vscodeignore`)
+  execSync('yarn install', { cwd: pkg_dir })
```
stevedlawrence, line 89: Is it possible to remove or replace yarn install some how? I think this is needed to create the `node_modules` stuff needed to run vsce? It would be really nice if yarn package/vsce package could just reuse the existing `node_modules` in the root directory somehow?

If it's not possible, do we need to also copy yarn.lock into `dist/package`? It looks like this creates a new one that isn't exactly the same as the existing one if we don't copy?

build/scripts/process_build.ts
```diff
@@ -18,39 +18,88 @@
...
+}
+
+function build_package() {
+  if (fs.existsSync(pkg_dir)) {
+    fs.rmdirSync(pkg_dir, { recursive: true })
+  }
+
+  fs.mkdirSync(pkg_dir)
+
+  let fileList = fs.readFileSync('.vscodeignore').toString().split('\n')
```
stevedlawrence, line 53: This is minor, but what are your thoughts on moving the .vscodeignore file to another directory since it is really only used during packaging? I'm thinking we could put .vscodeignore, bin.LICENSE, and bin.NOTICE files in something like `build/package/`. After we've copied all the files from .vscodeignore into `dist/package`, we could then just copy `build/package/*` into `dist/package/`. That way adding new static files to `dist/package` is as easy as putting them in `build/package`.

stevedlawrence (commented):
Really nice improvements. I like this approach a lot.

Just a few minor suggestions for cleanup and robustness.

build/scripts/process_build.ts
Shanedell: Yeah I forgot to add those my bad, I will make sure to fix that one the next commit

build/scripts/process_build.ts
Shanedell: Yeah I like the sound of that

build/scripts/process_build.ts
Shanedell: Fixed this I believe

build/scripts/process_build.ts
Shanedell: I just copied in the current node_modules into `dist/package` to make sure it the same packages are used.

Shanedell:
@stevedlawrence This should be good for another look. I think I resolved all the issues you addressed, other than the one left open in case I didn't fully satisfy what you wanted their.

build/scripts/process_build.ts
Shanedell: Actually I think I did this backwards I will fix that

Shanedell: Actually maybe I did it in a similar way. Will wait for input on what is currently being done before changin

Shanedell: So to do this how you were saying however, we would need to still run a `yarn install` before hand. But could copy in the yarn.lock, still ignore it and keep it out of the VSIX, to make sure nothing breaks and then it would be good to go.

build/scripts/process_build.ts
```diff
@@ -18,39 +18,98 @@
...
   }
+
+  // Copy ignore file
+  fileCopy('build/package/.vscodeignore', `${pkg_dir}/.vscodeignore`)
+
+  // Copy required package files into package directory
+  copyGlob('build/package/{.,}*', true)
+
+  // Copy node_modules into package directory
+  copyGlob('node_modules/{.,}**/{.,}*')
```
stevedlawrence, line 107: Is it fine to copy node_modules without also copying yarn.lock? Also, do we even need yarn install if we're copying node_modules? yarn is magic so I'm not sure what's actually needed.

package.json
```diff
@@ -25,21 +25,19 @@
...
     "publish": "vsce publish",
-    "package-ext": "webpack --mode production --config ./build/extension.webpack.config.js",
     "pretest": "yarn run compile",
     "test": "node node_modules/mocha/bin/_mocha -u tdd --timeout 999999 --colors ./out/tests",
     "sbt": "sbt universal:packageBin",
-    "prebuild": "node -e \"require('./build/scripts/process_build.ts').prebuild()\"",
-    "build": "yarn sbt && yarn install && yarn compile && yarn package",
-    "postbuild": "node -e \"require('./build/scripts/process_build.ts').postbuild()\""
+    "prebuild": "yarn sbt && yarn install && yarn compile && yarn webpack",
+    "build": "node -e \"require('./build/scripts/process_build.ts').build_package()\""
```
stevedlawrence, line 40: Just out of curiosity, what's the intended difference between compile/precompile and build/prebuild? What kind of things should happen in one and what should happen in the other? The two seems like similar concepts and it's not clear to me what the difference is.

For example, yarn omega-edit-download happens in compile, but that's not actually compiling anything, right? It just downloads a zip? Seems like maybe that wants to happen in a build or prebuild stage? Likewise, "yarn sbt" does potentially compile scala code, so feels like it wants to happen in compile stage? Or maybe "sbt compile" should happen in compile stage, and "sbt universal:packageBin" should happen in the build stage?

I'm not sure it really matters, but it's not totally clear to me. Just trying my best to understand yarn.

build/scripts/process_build.ts
```diff
@@ -18,39 +18,83 @@
...
+
+      fileCopy(file, `${pkg_dir}/${file}`)
+    }
   }
+
+  // Copy ignore file
+  fileCopy('build/package/.vscodeignore', `${pkg_dir}/.vscodeignore`)
+
+  // Copy required package files into package directory
+  copyGlob('build/package/*', true)
```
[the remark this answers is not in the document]
stevedlawrence: Looks like you've already fixed this.

build/scripts/process_build.ts
Shanedell: So it would be one or the other. I think I just forgot to remove one. I will remove yarn install.

build/scripts/process_build.ts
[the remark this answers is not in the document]
Shanedell: Yeah so this update must have just missed you. I have `{.,}*` which makes it check for any hidden files

build/scripts/process_build.ts
```diff
@@ -18,39 +18,98 @@
...
 const execSync = require('child_process').execSync
+const pkg_dir = 'dist/package'
+var fileCount = 0
 
-function prebuild() {
-  fs.renameSync('LICENSE', 'tmp.LICENSE')
-  fs.renameSync('NOTICE', 'tmp.NOTICE')
-  fs.copyFileSync('build/bin.NOTICE', 'NOTICE')
-  fs.copyFileSync('build/bin.LICENSE', 'LICENSE')
+function copyGlob(pattern, skipDirectory = false) {
```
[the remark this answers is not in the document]
Shanedell: Oh okay this makes sense to me

build/scripts/process_build.ts
```diff
@@ -18,39 +18,98 @@
...
+
+      fs.mkdirSync(`${pkg_dir}/${directory}`, { recursive: true })
+
+      fileCopy(file, `${pkg_dir}/${file}`)
+      fileCount++
+    }
   }
+
+  // Copy ignore file
+  fileCopy('build/package/.vscodeignore', `${pkg_dir}/.vscodeignore`)
```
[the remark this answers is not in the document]
Shanedell: Yeah I forgot to delete that line

package.json
```diff
@@ -25,21 +25,19 @@
...
     "omega-edit-download": "node -e \"require('./build/scripts/omega_edit_download.ts').downloadServer()\"",
-    "vscode:prepublish": "yarn run package-ext",
     "precompile": "node -p \"'export const LIB_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
     "compile": "tsc -p ./ && yarn omega-edit-download",
     "lint": "yarn run prettier src -c",
     "watch": "yarn omega-edit-download && webpack --watch --devtool nosources-source-map --config ./build/extension.webpack.config.js",
     "watch2": "tsc -watch -p ./",
-    "package": "vsce package",
+		"webpack": "webpack --mode production --config ./build/extension.webpack.config.js",
+		"package": "vsce package",
```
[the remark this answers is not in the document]
Shanedell: Yeah it can be removed that makes sense to me. We don't necessarily use publish because its not a node package we publish to npmjs

package.json
Shanedell: So basically compile should be everything dealing with creating the out js file, extension webpack file, I think technically it should have sbt running as well, maybe I left that out to make it quicker I am not fully sure. But I think with removing the `vsce package` command we actually replace the `build` script with `package` as that is what it is really doing anyway. Would that make sense to do?

build/scripts/process_build.ts
```diff
@@ -18,39 +18,98 @@
...
+        fs.mkdirSync(directory, { recursive: true })
+      }
+
+      fileCopy(file, `${pkg_dir}/${file}`)
+      fileCount++
+    }
+  })
+}
+
+function getDir(file) {
```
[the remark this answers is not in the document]
Shanedell: yeah I will delete that out

package.json
Shanedell: Also instead of the script being `build/scripts/process_build.ts` you think `build/scripts/package.ts` would be better?

Shanedell: I removed the copying of node_modules as it seemed to break in some cases, so I copy in `yarn.lock` then run `yarn install`

stevedlawrence: Renaming to package.ts seems reasonable to me. yarn.lock + yarn install seems fine too.

Shanedell:
> Why skip directories that match the glob? If they are listed in .vscodeignore don't we want them too?

So with the copy method you can't copy directories only files. That is why sub-directories are skipped, but I guess you could just run another glob recursively, is that what you mean?

build/scripts/package.ts
```diff
@@ -0,0 +1,84 @@
...
+const path = require('path')
+const glob = require('glob')
+const execSync = require('child_process').execSync
+const pkg_dir = 'dist/package'
+
+async function copyGlob(pattern) {
+  glob(pattern, (error, files) => {
+    for (var i = 0; i < files.length; i++) {
+      // Skip dirs as they are most likely listed in the .vscodeignore as well
+      if (fs.statSync(files[i]).isDirectory()) continue
```
stevedlawrence, line 30: Why skip directories that match the glob? If a glob matches a directory in .vscodeignore don't we want them too?

Shanedell:
Do you want maybe an entry to add to package.json like `packageDir` and this value is used to get the pkg_dir?

build/scripts/package.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      let src = files[i]
+      let dst = path.join(pkg_dir, src.replace('build/package', ''))
```
stevedlawrence, line 33: I'd prefer we don't hardcode `build/package` in this function. I imagine we might want to use this for something else in the future. Did the cwd approach mentioned in another comment not work?

build/scripts/package.ts
```diff
@@ -0,0 +1,84 @@
...
+
+// Create VSIX package
+function create() {
+  execSync('yarn install', { cwd: pkg_dir })
+  execSync('yarn vsce package --out ../../', { cwd: pkg_dir })
+}
+
+module.exports = {
+  setup: setup,
+  create: create,
```
stevedlawrence, line 83: Just out of curiosity, previously you had the exec stuff done after files were copied and now you have it split into separate functions. This is fine with me, but I'm trying to get a better grasp of how yarn works--is there a special reason this is needed, or just thought it would be good to split things up?

build/scripts/package.ts
stevedlawrence: > Do you want maybe an entry to add to package.json like packageDir and this value is used to get the pkg_dir?

It's more that I dont' think this function show know about the build/package directory at all.

For example, what if we also wanted

```
copyGlob("build/other-stuff/*")
```

Now we need to special case build/other-stuff too. Just trying to make this as generic as possible so we don't ever run into edge cases.

build/scripts/package.ts
Shanedell: See https://github.com/apache/daffodil-vscode/pull/232#issuecomment-1189027422 for comment about that

build/scripts/package.ts
Shanedell: So it is split because when I try to run those `execSync` commands in the same method as the copies, those cmds try to run before everything is copied causing package issues. So splitting out was done to make sure the package directory is completely setup before actually trying to run the package commands

build/scripts/package.ts
stevedlawrence: > Would something like be better, having a optional param `copyToRoot` and this is how the file is determined
> 
> ```ts
> let dst = copyToRoot
>         ? path.join(pkg_dir, src.replace(path.dirname(src), ''))
>         : path.join(pkg_dir, src)
> ```

That removes the entire directory part of the src though, leaving just the filename, right? We might want directories as well. I previously comment an alternative copyGlob idea that takes a dir parameter to tell what the root is, thoughts on something like this:

```
function copyGlob(pattern, dir = ".") {
  glob(pattern, {cwd=dir}, (error, files) => {
    for (var i = 0; i < files.length; i++) {
      let src = path.join(dir, file[i])
      let dst = path.join(pkg_dir, file[i])
      let parentDir = path.dirname(dst)
      ...  // ensure parentDir exists
      fileCopy(src, dst)
    }
  })
}
```

Then the usage is 
```
copyGlob("*", "build/package")
```
But also allows more complex things like

```
copyGlob("foo/*", "build/bar")
```
This will copy everything in `build/bar` that matches `foo/*`, including the `foo` directory. I think the copyToRoot idea doesn't do that and would just copy everything in foo/* without the foo directory.

It's a similar idea, but gives more control over what is the root and what isn't.

build/scripts/package.ts
Shanedell: Would this work

```typescript
if (fs.statSync(src).isDirectory() || !fs.existsSync(dstDir)) {
    fs.mkdirSync(dstDir, { recursive: true })
    if (fs.statSync(src).isDirectory()) continue
}
```

build/scripts/package.ts
Shanedell: So this makes sense but for the files inside of `build/package` don't we want them directly inside of `dist/package` and not `dist/package/build/package`?

build/scripts/package.ts
stevedlawrence: I'm not sure I understand the logic in that snippet. What's the reason for the existsSync check? mkdirSync with `recrusive:true` doesn't error if it already exist so shouldn't be needed?

I also generally perfer if/else instead of an if- with a continue in the block, with one block being dir logic and another block being file logic.

build/scripts/package.ts
stevedlawrence: If we do something like `copyGlob("*", "build/package")`, then `cwd` is set to `build/`  and the glob is run from there. That way the list of `files` won't include `build/package/`. That's why we then need to add `dir` to get the source file and `pkg_dir` to get the dst file.

For example, if we had `copyGlob("foo/*", "build/package")`, then `files` would be something like `[foo/bar.txt, foo/baz.txt]`. We then want to copy "dir + / + those" (e.g. build/package/foo/bar.txt) to "pkg_dir + / + those" (e.g. dist/package/foo/bar.txt). This way we keep the foo/ part of the glob but exclude build/package.

build/scripts/package.ts
Shanedell: Yeah since this TypeScript is nodejs script not sure if the async/await works completely

build/scripts/package.ts
Shanedell: Does the update made look better?

build/scripts/package.ts
```diff
@@ -0,0 +1,81 @@
...
+    for (var i = 0; i < files.length; i++) {
+      let src = path.join(dir, files[i])
+      let dst = path.join(pkg_dir, files[i])
+      let dstDir = path.dirname(dst)
+
+      fs.mkdirSync(dstDir, { recursive: true })
+
+      if (fs.statSync(src).isFile()) {
+        fs.copyFileSync(src, dst)
+      }
```
stevedlawrence, line 37: This means we ignore directories? As long as are globs only care about files in .vscodeignore I guess that's fine.

I'm not sure what the right behavior should be if a glob did match a directory. Would it mean we would want to recursively copy the directory? Or maybe ignoring is the right behavior. I don't feel strongly about it, so this behavior is fine and we can deal with it when it comes up.

build/package/.vscodeignore
```diff
@@ -17,8 +17,8 @@
 */**
 !dist/ext/extension.js
 !images/*
-!LICENSE
-!NOTICE
+# !LICENSE
+# !NOTICE
```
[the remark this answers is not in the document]
Shanedell: Yeah I added the ignore to make sure the build/package ones were being copied in, I will remove these comments but I think the package script will need to ignore them as it will copy in the root directory ones

build/scripts/package.ts
Shanedell: So there would be two options:

- List over sub directories inside of `.vscodeignore` and then skip over them in the glob since the allowed ones will be listed in the file (currently being done)
- Only list main directory and recursively call glob instead of listing sub dirs in the `.vscodeignore`

build/package/.vscodeignore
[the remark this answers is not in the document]
stevedlawrence: As long as `copyGlob("*", "build/package")` is after the vscodeignore copy it should be fine--it should just overwrite the ones copied from the root. Which is maybe a nice feature and an easy way to override files copied from the root if needed (e.g. maybe we'll want to to do that for the README file as well (see issue #100)

build/scripts/package.ts
stevedlawrence: I say leave it as is. We can decide how to handle this if it ever comes up.

Shanedell: Alright makes sense

stevedlawrence (approved):
+1 :+1: looks great to me!

scholarsmate (approved):

arosien (approved):
👍

PR #245, opened 2022-08-01
omega-edit updates

### LINKED ISSUE #125: REQ-FTS-01 Data editor needs to support any fixed length (non-streaming) file Daffodil is capable of opening
Generally, any file type can be opened and displayed by a hex editor. The file type and extension do not influence the rendering of the file in hex or binary formats.

### LINKED ISSUE #126: REQ-UI-01 Data editor needs to be responsive and provide a good Visual Studio Code User Experience
Existing third-party VS Code hex editors will decrease in responsiveness while rendering medium to large size files. The editor will handle file sizes common to Daffodil without impacting overall usability.

### LINKED ISSUE #160: REQ-CMD-09 Data editor needs to support searching for patterns
Search function similar to a text editor find text using literal text. This pattern would literally be searched for in each given representation.

### LINKED ISSUE #161: REQ-CMD-10 Data editor needs to support replacing searching results with new patterns
Search function similar to a text editor find text using literal text and replace with alternate text. This pattern would literally be searched for in each given representation and replaced using text that is valid within said representation.

Shanedell:
omega-edit updates:

- Implement search
- Implement search and replace
- Update UI to support both search and search/replace
- Update redo and undo to use server instead of work around
- Update omega-edit script to be omega-edit-grpc-server instead of example-grpc-server

Closes #125 
Closes #126 
Closes #160 
Closes #161


src/omega_edit/client.ts
```diff
@@ -97,25 +97,31 @@ export function activate(ctx: vscode.ExtensionContext) {
         randomId().toString(),
```
scholarsmate, line 97: I recommend leaving the desired session IDs undefined (unless they are explicitly named) to allow the server to generate them instead.  The C++ server will use a 36-character UUID and the Scala server will use an 8-character ID.  The servers make sure there are no collisions in the IDs.

src/omega_edit/client.ts
```diff
@@ -97,25 +97,31 @@ export function activate(ctx: vscode.ExtensionContext) {
...
         s,
         0,
-        1000
+        1000,
```
scholarsmate, line 100: Are we limited to 1000 lines?

scholarsmate (changes requested):
Heading in the right direction.  It's good to see all the lines that were removed from here and sent on to the server to handle.  We need to fix Scala RPC search.  Also, please fix the viewport subscription call back so it extracts data from the viewport event.

src/omega_edit/utils.ts
```diff
@@ -60,7 +59,8 @@ export function viewportSubscribe(
   commandViewport: string,
   commandHex: string | null
 ) {
-  client.subscribeToViewportEvents(new ObjectId().setId(vp1)).on('data', () => {
+  var objId = new ObjectId().setId(vp1)
+  client.subscribeToViewportEvents(objId).on('data', () => {
```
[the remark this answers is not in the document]
scholarsmate: Then it's probably another Scala RPC issue.  It works fine with the C++ RPC server.

src/omega_edit/omega_edit.ts
```diff
@@ -127,11 +84,69 @@ export class OmegaEdit {
...
+
+    // if no data sent don't insert
+    if (data !== '') {
+      await omegaEditChange.insert(sessionId, offset, data)
+    }
+  }
+
+  // Perform search getting on a single result
+  async singleSearch(fileSize: number, searchPattern: string): Promise<number> {
+    var result = await this.search(fileSize, searchPattern)
```
scholarsmate, line 117: To optimize this, set the limit to 1 so the server can stop searching after it finds the first instance of the pattern.

scholarsmate (changes requested):
Looks great overall, but needs a couple of changes.

src/omega_edit/client.ts
```diff
@@ -137,7 +137,8 @@ export function activate(ctx: vscode.ExtensionContext) {
             message.newFile ? message.newFile : false,
             fileSize,
             searchPattern,
-            message.replaceText ? message.replaceText : ''
+            message.replaceText ? message.replaceText : '',
+            message.caseSensitive ? message.caseSensitive : false
```
scholarsmate, line 141: This should be message.case**In**sensitive, and false is the correct default value.

src/omega_edit/interface.html
```diff
@@ -181,6 +181,13 @@
               />
               <span style="margin-left: 0px" class="checkmark"></span>
             </label>
+            <label
+              style="margin-left: 142%; margin-top: -16px; width: 105px"
+              class="container"
+              >Case Sensitive?
+              <input type="checkbox" id="caseSensitive" />
+              <span style="margin-left: 0px" class="checkmark"></span>
+            </label>
```
scholarsmate, line 190: The label should be Case Insensitive, the id should be caseInsensitive, and the default should be unchecked.

Ideally the UI would get a byte frequency profile from the server, then only allow this setting if the profile indicates that the data is all ASCII.

src/omega_edit/omega_edit.js
```diff
@@ -94,18 +94,20 @@ function replace(data, offset, len) {
   })
 }
 
-function searchSession(searchPattern) {
+function searchSession(searchPattern, caseSensitive) {
```
scholarsmate, line 97: caseInsensitive

src/omega_edit/omega_edit.js
```diff
@@ -94,18 +94,20 @@ function replace(data, offset, len) {
...
   vscode.postMessage({
     command: 'search',
     searchPattern: searchPattern,
+    caseSensitive: caseSensitive,
```
scholarsmate, line 101: caseInsensitive

src/omega_edit/omega_edit.js
```diff
@@ -94,18 +94,20 @@ function replace(data, offset, len) {
...
   })
 }
 
-function searchAndReplace(searchPattern, replaceText) {
+function searchAndReplace(searchPattern, replaceText, caseSensitive) {
```
scholarsmate, line 105: caseInsensitive

src/omega_edit/omega_edit.js
```diff
@@ -94,18 +94,20 @@ function replace(data, offset, len) {
...
   vscode.postMessage({
     command: 'searchAndReplace',
     searchPattern: searchPattern,
     replaceText: replaceText,
+    caseSensitive: caseSensitive,
```
scholarsmate, line 110: caseInsensitive

src/omega_edit/omega_edit.js
```diff
@@ -190,10 +192,14 @@ function executeSearchOrReplace() {
   if (isReplace) {
     searchAndReplace(
       document.getElementById('searchInput').value,
-      document.getElementById('replaceInput').value
+      document.getElementById('replaceInput').value,
+      !document.getElementById('caseSensitive').checked
```
scholarsmate, line 196: caseInsensitive

src/omega_edit/omega_edit.js
```diff
@@ -190,10 +192,14 @@ function executeSearchOrReplace() {
...
     )
   } else {
-    searchSession(document.getElementById('searchInput').value)
+    searchSession(
+      document.getElementById('searchInput').value,
+      !document.getElementById('caseSensitive').checked
```
scholarsmate, line 201: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -86,15 +86,17 @@ export class OmegaEdit {
 
   async search(
     fileSize: number,
-    searchPattern: string
+    searchPattern: string,
+    caseSensitive: boolean,
```
scholarsmate, line 90: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -86,15 +86,17 @@ export class OmegaEdit {
...
+    limit: number = 0 // unlimited for omega-edit is 0
   ): Promise<Array<number>> {
     return await omegaEditSession.searchSession(
       this.sessionId,
       Buffer.from(searchPattern),
-      true,
-      0,
+      caseSensitive,
```
scholarsmate, line 96: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -113,17 +115,22 @@ export class OmegaEdit {
   }
 
   // Perform search getting on a single result
-  async singleSearch(fileSize: number, searchPattern: string): Promise<number> {
-    var result = await this.search(fileSize, searchPattern)
+  async singleSearch(
+    fileSize: number,
+    searchPattern: string,
+    caseSensitive
```
scholarsmate, line 121: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -113,17 +115,22 @@ export class OmegaEdit {
...
+  ): Promise<number> {
+    var result = await this.search(fileSize, searchPattern, caseSensitive, 1)
```
scholarsmate, line 123: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -113,17 +115,22 @@ export class OmegaEdit {
...
     return result[0]
   }
 
   async searchAndReplace(
     fileSize: number,
     searchPattern: string,
-    replaceText: string
+    replaceText: string,
+    caseSensitive: boolean
```
scholarsmate, line 131: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -113,17 +115,22 @@ export class OmegaEdit {
...
   ) {
-    var index = await this.singleSearch(fileSize, searchPattern)
+    var index = await this.singleSearch(fileSize, searchPattern, caseSensitive)
```
scholarsmate, line 133: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -135,7 +142,7 @@ export class OmegaEdit {
         replaceText
       )
 
-      index = await this.singleSearch(fileSize, searchPattern)
+      index = await this.singleSearch(fileSize, searchPattern, caseSensitive)
```
scholarsmate, line 145: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -146,7 +153,8 @@ export class OmegaEdit {
     newFile: boolean,
     fileSize: number,
     searchPattern: string,
-    replaceText: string
+    replaceText: string,
+    caseSensitive: boolean
```
scholarsmate, line 157: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -171,12 +179,21 @@ export class OmegaEdit {
         await this.replace(this.sessionId, this.offset, this.len, this.data)
         break
       case 'search':
-        var searchResults = await this.search(fileSize, searchPattern)
+        var searchResults = await this.search(
+          fileSize,
+          searchPattern,
+          caseSensitive
```
scholarsmate, line 185: caseInsensitive

src/omega_edit/omega_edit.ts
```diff
@@ -171,12 +179,21 @@ export class OmegaEdit {
...
         if (searchResults)
           vscode.window.showInformationMessage(searchResults.toString())
         break
       case 'searchAndReplace':
-        await this.searchAndReplace(fileSize, searchPattern, replaceText)
+        await this.searchAndReplace(
+          fileSize,
+          searchPattern,
+          replaceText,
+          caseSensitive
```
scholarsmate, line 195: caseInsensitive

scholarsmate (changes requested):
You need to rename caseSensitive to caseInsensitive everywhere and fix the checkbox labelling.  Defaulting it to false is correct.

Once the byte frequency profiling is available in Ωedit, we can enhance the UI by allowing case insensitive searching only on ASCII data.

scholarsmate (approved):

scholarsmate:
@stevedlawrence and @mbeckerle, will you guys review this PR?  It's the big feature for version 1.2.0.

stevedlawrence (approved):
+1, assuming no license changes/issues

Minor comments otherwise. I

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "VS Code extension for Apache Daffodil DFDL schema debugging",
   "version": "1.1.0",
   "daffodilVersion": "3.3.0",
-  "omegaEditServerHash": "2f8fec09a67aa2a9d844bcb7c0c0dd54f702186ddaa1eb13e73b1ec6b3cf6d73b687b3e9b35b46da609a14746ff7ec884f7d6abc5ae4ac9ce634235e78c3d56d",
+  "omegaEditServerHash": "ffabce0d095364d1c863466f522dfa0ac2f0574d6394a82c0239a613bb84baf7c7b9fd92fdb0457e1d3c9321249acc21567d7dcfd6ff0528e0e792da1b487407",
```
mbeckerle, line 7: Is there a comment syntax for this json, or some way to explain this hash, like where it comes from, how created, etc. 
Or is that documented elsewhere?

src/omega_edit/client.ts
```diff
@@ -93,56 +93,52 @@ export function activate(ctx: vscode.ExtensionContext) {
...
             message.len,
             panel
           )
+
+          var fileSize = await omegaEditSession.getComputedFileSize(s)
+          var searchPattern = message.searchPattern ? message.searchPattern : ''
+
+          if (searchPattern !== '' && fileSize < searchPattern.length) {
+            throw new Error("Search pattern can't be larger than file")
+          }
```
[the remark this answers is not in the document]
mbeckerle: So data could be a small number of bytes, but the search pattern might be using hex, or even individual bits (someday) so this notion that something is wrong if the pattern is larger than the data may not hold. For now it seems sensible, but I can see this having to be taken out in the future. 

E.g., There's a complete positive VMF test message that is 11 bytes long, and given the bit-packed nature of it, searching in it for a bit string might use things like search for "0110001" (0x31, but as 7 bits only), in which case the patterns, could be, in bytes (0 or 1 are each a byte), bigger than the data in bytes.

src/omega_edit/interface.html
```diff
@@ -39,13 +37,13 @@
       }
     </style>
     <style></style>
+    <link rel="stylesheet" href="../styles/styles.css" />
   </head>
   <body>
     <div>
       <!-- <div id="server">uri</div> -->
```
mbeckerle, line 44: Need these? If not remove.

src/omega_edit/omega_edit.js
```diff
@@ -88,6 +85,32 @@ function saveSession() {
...
+
+function searchSession(searchPattern, caseInsensitive) {
+  vscode.postMessage({
+    command: 'search',
+    searchPattern: searchPattern,
+    caseInsensitive: caseInsensitive,
+  })
+}
+
+function searchAndReplace(searchPattern, replaceText, caseInsensitive) {
```
mbeckerle, line 105: I'm not sure this makes sense, but .... So is this text only, or is this used for binary/bytes data also with the replacement text being hex?

I guess I'm looking for a way to keep straight what's for binary vs. text only vs. are there things used in multiple modes?

src/omega_edit/omega_edit.js
```diff
@@ -88,6 +85,32 @@ function saveSession() {
...
+    command: 'searchAndReplace',
+    searchPattern: searchPattern,
+    replaceText: replaceText,
+    caseInsensitive: caseInsensitive,
+  })
+}
+
 function setupViewport(viewportId) {
   viewportElement = document.getElementById(viewportId)
   viewportElement.addEventListener('keydown', (event) => {
```
mbeckerle, line 116: Hmmm. Keydown event for specific keys, feels like a user-configurable JSON file should specify the key bindings used, not straight code. 

Fine if that's in a future plan, but I'm going to want to swtich all these bindings to be EMACS-like.

src/omega_edit/omega_edit.ts
```diff
@@ -127,11 +84,81 @@ export class OmegaEdit {
...
+  ) {
+    await omegaEditChange.del(sessionId, offset, len)
+
+    // if no data sent don't insert
+    if (data !== '') {
+      await omegaEditChange.insert(sessionId, offset, data)
+    }
+  }
+
+  // Perform search getting on a single result
```
mbeckerle, line 117: Reword comment? "on a" ??

mbeckerle (approved):
+1 

Minor comments only. 

Most important thing for future reviews is for me to fix a beginner bug or few with vscode extension so that I can know better what I am doing in these reviews.

src/omega_edit/omega_edit.js
```diff
@@ -132,29 +155,15 @@ function setupViewport(viewportId) {
       }
       insertByte('clipboard', viewportElement.selectionStart, 0)
     } else if ((event.ctrlKey || event.metaKey) && keyLower === 'z') {
```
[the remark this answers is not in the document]
Shanedell: So this was to support `Command + Z` on Mac, possibly map to Alt on Linux/Windows I think that may have been overlooked on my end

package.json
```diff
@@ -47,7 +47,7 @@
     "google-protobuf": "3.20.1",
     "hexy": "0.3.4",
     "moo": "0.5.1",
-    "omega-edit": "0.9.8",
+    "omega-edit": "0.9.20",
```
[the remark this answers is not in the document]
Shanedell: No. From my knowledge omega-edit uses mostly tools we have included in the LICENSE and NOTICE currently

src/omega_edit/client.ts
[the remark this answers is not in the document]
Shanedell: So this check is for making sure when a pattern is given that it is not bigger than the size of file being searched.

src/omega_edit/omega_edit.js
```diff
@@ -88,6 +85,32 @@ function saveSession() {
...
+function replace(data, offset, len) {
+  vscode.postMessage({
+    command: 'replace',
+    data: data,
+    offset: offset,
+    len: len,
+  })
+}
+
+function searchSession(searchPattern, caseInsensitive) {
```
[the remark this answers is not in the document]
Shanedell: Yes you are correct I may have just throw the session suffix in there where initially creating it.

package.json
Shanedell: So normally this is me downloading the package file manually and running `sha512sum` on it to get this hash value.

src/omega_edit/omega_edit.js
Shanedell: So I think with the webview you can't make a json with key mappings. From what I was able to find this is the only may I could get the key bindings to work using a webview.

package.json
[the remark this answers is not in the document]
stevedlawrence: I did a diff of the release in 0.9.8 and 0.9.20, and there are differences:

New jars:
* com.beachape.enumeratum-macros_2.13-1.6.1.jar
* com.beachape.enumeratum_2.13-1.7.0.jar
* com.monovore.decline_2.13-2.3.0.jar
* org.scala-lang.scala-reflect-2.13.8.jar
* org.typelevel.cats-core_2.13-2.7.0.jar
* org.typelevel.cats-kernel_2.13-2.7.0.jar
* org.typelevel.simulacrum-scalafix-annotations_2.13-0.5.4.jar

Updated jar:
* com.github.jnr.jnr-ffi-2.2.11.jar -> com.github.jnr.jnr-ffi-2.2.12.jar

I didn't check all of these, but enumeratum is MIT and isn't listed, so at the very least that needs to be added to build/package/LICENSE. Please double check I didn't miss any jars and make sure they are mentioned in the LICENSE/NOTICE file.

src/omega_edit/omega_edit.js
Shanedell: So currently we are only searching and replacing using text in the UI. Currently we only have the editing of text fully implemented in the UI. In the future we plan to support both editing the binary data as well, this is already support by omega-edit just needs implemented in the frontend, but for now mostly focus on text.

src/omega_edit/client.ts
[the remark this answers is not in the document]
stevedlawrence: > So this check is for making sure when a pattern is given that it is not bigger than the size of file being searched.

Understood, but why does that matter? Do things break if the pattern is larger than the input data?

package.json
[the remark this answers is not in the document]
Shanedell:
```
org.scala-lang.scala-reflect
org.typelevel.cats-core
org.typelevel.cats-kernel
```
Are in the LICENSE/NOTICE already based on the daffodil-debugger code we have inside of the repo. The others do need added so I will make sure to do that.

src/omega_edit/client.ts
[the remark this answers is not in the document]
Shanedell: Yeah I believe it causes the server to crash but @scholarsmate would be better at explaining why than myself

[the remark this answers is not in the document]
stevedlawrence: Seems like a bug somewhere else?

package.json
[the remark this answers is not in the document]
Shanedell:
```
org.typelevel.simulacrum-scalafix-annotations
com.monovore.decline
com.github.jnr.jnr-ffi
```
are licensed under ALv2 with no NOTICE file. So those just didn't get listed right @stevedlawrence ? I added the two enum ones but those were the only two that actually needed to be added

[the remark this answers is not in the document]
stevedlawrence: Yep, ALv2 with no NOTICE does not need to be anywhere.

src/omega_edit/client.ts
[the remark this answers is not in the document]
scholarsmate: If the needle is larger than the haystack, the library will return a nullptr, indicating a search error, rather than a lack of matches.  The check in question happens here: https://github.com/ctc-oss/omega-edit/blob/main/src/lib/search.cpp#L47.  I'm open to discussion on how and where we think such problems should be handled, but I'd prefer the bad input not to be sent in the first place, thus this sanity check being done early in the client.  Other options include ignoring the problem and returning an empty list of matches, or having to bubble up error status and dealing with it after a round trip has been taken.

src/omega_edit/omega_edit.js
```diff
@@ -132,29 +155,15 @@ function setupViewport(viewportId) {
...
       insertByte('clipboard', viewportElement.selectionStart, 0)
     } else if ((event.ctrlKey || event.metaKey) && keyLower === 'z') {
-      var lastChangeElement = document.getElementById('lastChange')
-      var changeCount = document.getElementById('changeCount').value
-
-      if (lastChangeElement.value === '') {
-        return
-      }
-
       if (event.shiftKey) {
```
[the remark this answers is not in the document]
Shanedell: @stevedlawrence Do you want this updated or how do we proceed with this one?

src/omega_edit/omega_edit.js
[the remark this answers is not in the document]
Shanedell: @stevedlawrence Do you want this updated/changed or how do we proceed with this one?

src/omega_edit/omega_edit.js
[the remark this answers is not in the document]
scholarsmate: `Command + Z` is undo and `Command + Shift + Z` is redo on macOS.  As another data point, neither `Ctrl + Y` nor `Command + Y` seem to have any affect in my editors on macOS.

src/omega_edit/client.ts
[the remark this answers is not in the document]
scholarsmate: Hex, Octal, ASCII, etc. are presentation details (views) of the same underlying bit pattern (model).  At this level in the stack, we're sending the bit pattern (in bytes), so it is expected to always be no greater than the haystack.  The smallest addressable unit is a byte.

If we want to search for bit patterns that are not aligned on byte boundaries then we can use bit shifts and if the bit patterns are not divisible by 8, then we can use masks.  This advanced searching feature (or corresponding
replace) hasn't been implemented in Ωedit yet.  Currently it only works on aligned bytes.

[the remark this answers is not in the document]
stevedlawrence: >  but I'd prefer the bad input not to be sent in the first place,

I'm not sure I'd consider that bad input, I'd just consider it a search pattern that just will never match. If you do `myString.indexOf("foo")`, java doesn't throw an exception if "foo" is larger than the length of `myString`. It's not an error, it just won't match and returns the "no match" value of -1.

>  I'm open to discussion on how and where we think such problems should be handled,

IMO, I wouldn't treat it any different than any other search pattern. Send the pattern to whatever does the search, return no matches if the pattern doesn't match (and it won't in the case of too large data), and handle no match like any other pattern that has no matches.


That said, it's fine with me if that's the design decision, but I think if this check is going to be made here it needs a comment as to why it's there, because it feels non-standard to me. Especially if it could lead to a null pointer exception if someone removed it because they decided they didn't want what appears to just be an optimization.

src/omega_edit/omega_edit.js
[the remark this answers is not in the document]
stevedlawrence: I have no strong preference.

It looks like VS Code (on Linux) supports both `Ctrl+Shift+Z` and `Ctrl+Y` for redo, though the `Edit` menu lists `Ctrl-Y` as redo. And wikipedia has this to say:

> In most [Microsoft Windows](https://en.wikipedia.org/wiki/Microsoft_Windows) applications, the [keyboard shortcut](https://en.wikipedia.org/wiki/Keyboard_shortcut) for the undo command is [Ctrl+Z](https://en.wikipedia.org/wiki/Control-Z) or Alt+Backspace, and the shortcut for redo is [Ctrl+Y](https://en.wikipedia.org/wiki/Control-Y) or [Ctrl](https://en.wikipedia.org/wiki/Control_key)+[Shift](https://en.wikipedia.org/wiki/Shift_key)+Z.
>
> In most [Apple Macintosh](https://en.wikipedia.org/wiki/Apple_macintosh) applications, the shortcut for the undo command is [Command](https://en.wikipedia.org/wiki/Command_key)-Z, and the shortcut for redo is [Command](https://en.wikipedia.org/wiki/Command_key)-[Shift](https://en.wikipedia.org/wiki/Shift_key)-Z.

So it seems to be mish-mash of shortcuts.

I assume there's no way to just get an undo/redo event from VS Code so we don't even have to worry about what the particular shortcut is? If not, I wouldn't put much effort into it, what you have is fine. We can change it in the future if this ends up causing confusion with users.

src/omega_edit/client.ts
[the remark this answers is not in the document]
scholarsmate: Okay, I'm good with the comment approach.

Since we have the data right at our fingertips, and a simple check tells us that no matches are possible, I don't see a need to take the trip even if I modified the Ωedit code to return an empty set.  If I change the behavior in Ωedit later, this code is still fine, requiring no changes.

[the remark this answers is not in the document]
scholarsmate: @Shanedell, here is the comment I'd propose adding above this check:

          // If the search pattern exceeds the length of the file, matches are
          // not possible.  Ωedit (as implemented currently) considers
          // patterns that are longer than the length of the file to be an
          // error (it will return a null pointer instead of an empty list).

src/omega_edit/omega_edit.js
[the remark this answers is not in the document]
scholarsmate: @Shanedell, I think we proceed "as-is" with this code.  You have 2 binding +1s.

src/omega_edit/omega_edit.js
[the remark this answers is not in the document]
Shanedell: I will keep it at `Ctrl+Shift+Z` for now. Maybe add in the `Ctrl+Y` later down the road if it desired.

Issue #173, opened 2022-06-08
REQ-TDML-06 Extension shall provide an item in the command palette (ctrl + shift + p) for ‘Run Test Case in TDML File’


scholarsmate:
**The extension shall provide an item in the command palette (ctrl + shift + p) for ‘Run Test Case in TDML File’.**

Selecting this command shall bring up menus allowing the user to select the following:

- TDML File Name
- Test Case to run (This list shall be populated with data in the selected TDML File)
- 

This command shall start the Daffodil process in run mode. This command should provide an option to start the Daffodil process in debug mode. The location of the DFDL Schema shall be expected to be relative to the location of the TDML File. It shall be the user who created the TDML file’s responsibility to ensure that packaging of their TDML file is correct.

michael-hoke:
Unit Testing:

- Create a TDML file with Generate. Run it through Execute and verify the output.
- Test that the LaunchArgs validation works in the scala side (Parse.scala)
- Mock the TDML code, and test a pass case for a Execute
- Test that the validation works on the Typescript side

arosien:
Closed by #253.

Issue #172, opened 2022-06-08
REQ-TDML-05  The extension shall provide an item in the command palette (ctrl + shift + p) for ‘Add Test Case to TDML File’


scholarsmate:
**The extension shall provide an item in the command palette (ctrl + shift + p) for ‘Add Test Case to TDML File’.**

Selecting this command shall bring up menus allowing the user to select the following:

- TDML File Name
- Name for the test case
- Description for the test case
- DFDL Schema
- Data Document

This selection shall work in the same way as the DFDL debugger where if you select the command from a DFDL Schema, it will automatically use that in place of a selection.

michael-hoke:
Unit Testing:

- Make a TDML file with Generate, and call Append using that file. Verify that the new file contains two test cases and that the added test case and the pre-existing test case are those test cases.
- Test that the LaunchArgs validation works in the scala side (Parse.scala)
- Mock the TDML code, and test a pass case for an append
- Test that the validation works on the Typescript side

arosien:
Closed by #253.

arosien:
Closed by #253.

Issue #171, opened 2022-06-08
REQ-TDML-04 The extension shall provide an item in the command palette (ctrl + shift + p) for ‘Generate TDML File’


scholarsmate:
**The extension shall provide an item in the command palette (ctrl + shift + p) for ‘Generate TDML File’.**

Selecting this command shall bring up menus allowing the user to select the following:

- TDML File Name
- Name for the test case
- Description for the test case
- DFDL Schema
- Data Document

This selection shall work in the same way as the DFDL debugger where if you select the command from a DFDL Schema, it will automatically use that in place of a selection. The TDML File shall be created in the workspace directory. The DFDL Schema and Document files shall also be file names only.

- These file names shall be relative to the workspace directory. It will be the user’s responsibility to organize everything when creating a TDML file and to package the files up for distribution.
- The name of the TDML file shall be the name of the DFDL schema used with ‘.tdml’ appended to the end.

michael-hoke:
Unit testing:

- Test that the createTestCase function produces expected output
- Test that the file produced by generate contains the XML section from createTestCase and that it only contains one test case
- Test that the LaunchArgs validation works in the scala side (Parse.scala)
- Mock the TDML code, and test a pass case for a generate

Issue #170, opened 2022-06-08
REQ-TDML-03 Test Data Markup Language features shall be written in Scala and shall read/write XML by using XML bindings 


scholarsmate:
Test Data Markup Language features shall be written in Scala and shall read/write XML by using XML bindings (e.g., Jaxb/scalaxb).

stevedlawrence:
One thing to maybe consider is that Daffodil's TDML runner already has a object model of TDML files. It might be easier to incorporate changes into that library and then reference them from the vs code extension, rather than start from scratch.

mbeckerle:
While I generally prefer Scala to Java, one experiment I have not done is try "new" Java, i.e., write Java, but not the old bad Java style, but by making use of all the new stuff that has been added over the past years, much of which was borrowed from Scala to support a more functional programming style.  If you want to try this, I'd support this idea of trying Java for this on that basis. 

But otherwise I'd still advocate that we should use Scala. 

If scalaxb is not working but jaxb is, you can certainly use the jaxb stuff from Scala. It may not be as smooth or type-safe as scalaxb, but would still let you write in Scala.  

Also, since Daffodil's tdml.xml is open-source, we can submit a good bug report to the scalaxb maintainers so they can reproduce whatever the problem is, and maybe with our help it gets fixed.

michael-hoke:
Sorry for my lack of response on this for so long, but I did take your feedback, and I believe I am getting close to having a PR ready. I'm hoping we can start resolving any issues you think are outstanding before the introduction of the PR.

Scala vs Java: I ended up using Scala for the necessary back end changes. This likely would have been inevitable due to how tied to the execution of DFDL that the new TDML operations are.

Daffodil's TDML Runner: I looked into this briefly but ended up sticking with JAXB for the initial version. I believe the JAXB portion of the code has been modularized enough so that if someone wants to extend this effort and share the TDMLRunner between this repo and the Daffodil repo, they should be able to do so without too much headache.

[...]

PR #253, opened 2022-08-11
Daffodil VsCode TDML

### LINKED ISSUE #9: Generate TDML files
Implement a VS Code command that will prompt for a schema file and data file and construct and save a TDML file.

Envisioned as a way to package up examples and test cases.

Original issue https://github.com/jw3/example-daffodil-vscode/issues/44

### LINKED ISSUE #168: REQ-TDML-01 External files needed by the TDML file shall be incorporated as relative paths into the TDML file
All external files needed by the TDML file shall be incorporated as relative paths into the TDML file.

michael-hoke:
Initial support for generating and executing TDML files from the Daffodil extension:
- generate a TDML file with a single test case from a schema and data file;
- append a test case to an existing TDML file from a schema and data file;
- execute a debug session for a single test case from an existing TDML file.

Closes #9 #168 #169 #170 #171 #172 #173


stevedlawrence:
Please squash into a single commit and rebase on top of the main branch. And please make the comment desribe what and why you're adding. It looks like some files are just copied form Daffodil source or other pleases. Please mention that and if any changes were made so we know what and what not to focus on.

Also, in the future, please try to split of PR's into smaller bite size chunks. This PR closes 8 issues. I would usually prefer 8 separate PR's specific to each issue. That makes things much easier to review.

stevedlawrence:
Also, it looks like a "daffodil-vscode-tdml" branch was pushed to the asf repo. Was that intentional? If not, we should delete that and only look at this PR from a fork.

michael-hoke:
It has already been rebased (as of yesterday) on top of main. Not sure why Git says it can't be rebased... 

If there is another branch out there, I am unaware of it, and it should be removed. The only branch I worked off of was the one in my fork. It looks like there are a lot of duplicate commits in there - I'll see if I can remove them before squashing - maybe that will fix the rebase issue.

michael-hoke:
No idea what's going on with the CI right now. I pushed up formatting and Rat changes. The formatting is passing on my local machine... For the licensing, I think the bindings.xjb file is what it is complaining about - do I add that to the exclude list, or is there a license you have in mind for an XML file?

Also not sure why ubuntu would be failing - I'm using wsl/ubuntu and it builds locally for me. I'll look into these more on Monday.

michael-hoke:
In order to pass the Rat check, I added the JAXB bindings.xjb file to the exclude list. If there is an appropriate license for an xjb file, let me know, and I'll put that in.

build.sbt
```diff
@@ -88,3 +90,183 @@ lazy val core = project
...
+  .in(file("server/sbtXjc"))
+  .enablePlugins(SbtXjcPlugin)
+  .settings(
+    name := "daffodil-xjc",
+    libraryDependencies ++= Seq(
+      "javax.activation" % "activation" % "1.1.1",
+      "com.sun.xml.bind" % "jaxb-xjc" % "2.1.6"
+      // "org.relaxng" % "jing" % "20220510" % "runtime",
+
+      // "com.sun.istack" % "istack-commons-tools" % "4.1.1"
```
arosien, line 106:
```suggestion
```

build.sbt
```diff
@@ -88,3 +90,183 @@ lazy val core = project
...
+    ),
+    xjcCommandLine += "-nv",
+    xjcCommandLine += "-p",
+    xjcCommandLine += "org.apache.daffodil.tdml",
+    xjcBindings += "bindings.xjb",
+    xjcJvmOpts += "-classpath",
+    xjcJvmOpts += Seq(
```
arosien, line 113: Concern: it looks very suspicious that all of these runtime dependencies have to be specified this way. What is requiring them, the XJC tool itself?

project/plugins.sbt
```diff
@@ -21,3 +21,6 @@ addSbtPlugin("io.github.davidgregory084" % "sbt-tpolecat" % "0.1.22")
 addSbtPlugin("org.musigma" % "sbt-rat" % "0.7.0")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.6")
 addSbtPlugin("com.github.battermann" % "sbt-json" % "0.5.0")
+addSbtPlugin("org.scala-sbt.plugins" % "sbt-xjc" % "0.10")
+
+// resolvers += "Java Net" at "http://download.java.net/maven/2/"
```
arosien, line 26:
```suggestion
```

server/core/src/main/scala/org.apache.daffodil.tdml/TDML.scala
```diff
@@ -0,0 +1,214 @@
...
+                    .normalize()
+                    .toString()
+                  return (schemaPath, dataPath)
+              }
+            }
+          }
+      }
+    }
+
+    // If there is no test case in the TDML file meeting the name/description criteria, return empty
```
arosien, line 211: Suggestion: the comment implies that the return type should be `Option[(String, String)]`.

build.sbt
```diff
@@ -46,7 +47,8 @@ lazy val commonSettings =
     // remove the -Xcheckinit option added by the sbt tpoletcat plugin. This
     // option leads to non-reproducible builds
     scalacOptions --= Seq("-Xcheckinit"),
-    startYear := Some(2021)
+    startYear := Some(2021),
+    Compile / managedResourceDirectories += baseDirectory.value / "server" / "sbtXjc" / "target" / "scala-2.12" / "src_managed_cxf" / "org" / "apache" / "daffodil" / "tdml"
```
arosien, line 51:
```suggestion
```

This line isn't needed; the generated sources are automatically added.

build.sbt
michael-hoke: Yes. The sbt-xjc plugin can be found here: https://github.com/sbt/sbt-xjc. The details for the different keys are provided there as well.

The "-p org.apache.daffodil.tdml" on the command line specifies the namespace for the generated files.

I'd love to be able to take out the classpath, but the generation of the JAXB files errors out because the underlying java process can't find classes that it's looking for. Each item on the classpath was found manually through build failures.

server/core/src/main/scala/org.apache.daffodil.tdml/TDML.scala
michael-hoke: Does rephrasing the end portion from "... return empty" to "... return an empty tuple" remove this implication?

build.sbt
arosien: I guess I find it weird the tool isn't self-contained, but I'll check that out out-of-band.

michael-hoke:
Should the version be incremented as a part of this PR - going from 1.1.0 to 1.1.1? If version 1.1.0 of the extension has already been used on a PC and this branch is installed and run, the Scala backend won't be re-downloaded and updated, and the old, existing backend will be used. This results in the TDML commands showing up in the frontend, but no TDML code will actually be executed in the Scala backend.

arosien:
> Should the version be incremented as a part of this PR - going from 1.1.0 to 1.1.1? If version 1.1.0 of the extension has already been used on a PC and this branch is installed and run, the Scala backend won't be re-downloaded and updated, and the old, existing backend will be used. This results in the TDML commands showing up in the frontend, but no TDML code will actually be executed in the Scala backend.

The convention in JVM libraries is to have, on the main branch, a version string of `x.y.z-SNAPSHOT` which means the artifact is continually changing and therefore should be re-fetched on every resolution. The download-the-scala-backend functionality in the extension should have something similar, or at least have a way to override the backend location for local development.

.vscode/launch.json
```diff
@@ -48,6 +48,15 @@
 			],
 			"internalConsoleOptions": "openOnSessionStart",
 			"preLaunchTask": "npm: compile"
+		},
+		{
+			"type": "scala",
+			"request": "launch",
+			"name": "DAPodil",
```
stevedlawrence, line 55: What does this actually launch? What is DAPodil?

bindings.xjb
```diff
@@ -0,0 +1,58 @@
+<jxb:bindings version="2.1"
```
stevedlawrence, line 1: This needs a license header.

Also, is this generated? Can this just be created as part of the build process?

bindings.xjb
```diff
@@ -0,0 +1,58 @@
...
+        <jxb:bindings node="//xs:complexType[@name='PropertyType']">
+            <jxb:class name="PropertyTypeClass"/>
+        </jxb:bindings>
+    </jxb:bindings>
+    <jxb:bindings schemaLocation="resources/xsd/dfdlx.xsd" version="1.0">
+        <jxb:bindings node="//xs:simpleType[@name='PropertyNameType']">
+            <jxb:typesafeEnumClass name="PropertyNameTypeX"/>
+        </jxb:bindings>
+    </jxb:bindings>
+</jxb:bindings>
```
stevedlawrence, line 58: What is the decision point for what bindings get included? There seems to be very small subset of everything defined in the schemas?

build.sbt
```diff
@@ -88,3 +89,180 @@ lazy val core = project
...
+        "https",
+        "repo1.maven.org",
+        "maven2",
+        "com",
+        "sun",
+        "xml",
+        "bind",
+        "jaxb-impl",
+        "2.2.11",
+        "*"
```
stevedlawrence, line 263: Is adding all these paths manually really the expected usage of this? This seems incredibly fragile and hard to maintain, especially with the verisions in the. Isn't any version change going to break this? Does the sbt xjc plugin not handle all of this?

project/Rat.scala
```diff
@@ -34,7 +34,9 @@ object Rat {
     file("images/daffodil.ico"),
     // yarn and rpm generated files
     file("yarn.lock"),
-    file("package-lock.json")
+    file("package-lock.json"),
+    // JAXB Bindings files
+    file("bindings.xjb")
```
stevedlawrence, line 39: If a file can have a license, it should have one. An xjb file is just XML and so can have a license. This should be removed from the Rat excludes.

resources/xsd/DFDL_part3_model.xsd
```diff
@@ -0,0 +1,368 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
stevedlawrence, line 1: I assume all these files in `resources/xsd` have just been copied and unmodified from Daffodil? Can they instead be pulled out of the the daffodil-lib dependency?

Otherwise these files may start to drift from the ones in Daffodil when we make changes and we'll have to remember (and most likely forget) to keep things updated.

build.sbt
```diff
@@ -88,3 +89,180 @@ lazy val core = project
...
+      ).mkString(java.io.File.separator)
+    ).mkString(java.io.File.pathSeparator),
+    Compile / xjc / sources := Seq(file("resources/xsd")),
```
stevedlawrence, line 266: Should these files be in the `server/sbtXjc` subproject? Though, maybe I'm not sure what the sbtXjc subproject is for? That directory doesn't actually exist?

resources/xsd/xml.xsd
```diff
@@ -0,0 +1,155 @@
...
+<!--
+  Copyright © 2018 World Wide Web Consortium, (Massachusetts
+  Institute of Technology, European Research Consortium for Informatics and
+  Mathematics, Keio University, Beihang). All Rights Reserved. This work is
+  distributed under the W3C® Software License [1] in the hope that it will be
+  useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+  [1] http://www.w3.org/Consortium/Legal/2002/copyright-software-20021231
+-->
```
stevedlawrence, line 11: Many of these files you've copied have different licenses. If they are going to be copied like this (which I hope there is an alternative), then you need to update the LICENSE/NOTICE files accordingly.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
```diff
@@ -176,71 +180,333 @@ object Parse {
...
+            .leftMap(t => s"'infosetOutput.path' field from launch request is not a valid path: $t")
+            .ensureOr(file => s"can't write to infoset output file at ${file}") { f =>
+              val file = Paths.get(f).toFile()
+              // If an empty string is passed in, it will be set to the workspace directory by default
+              // This is inside the Java code, so we have to make sure that the TDML file we
+              //   are working with is not a directory
+              !file.isDirectory() && (file.canWrite || (!file.exists && file.getParentFile.canWrite))
+            }
+        )
+        .toEitherNel
```
stevedlawrence, line 463: I'm a little concerned that this duplicates a decent amount of logic from the daffodil tdml library. And there's a lot of non trivial code in the TDML library that isn't implemented that we definitely don't want to reimplement. For example, input data can be defined in binary, hex, or characters, and even more complicates like bitOrder, replaceDFDLEntitites, etc. And this is just the data portion. There's a lot of complexity to TDML files that I would hate to have duplicated.

.vscode/launch.json
michael-hoke: This is the launch config for the scala backend. I was using this during debugging. I can remove this if we don't want it in the repo, but I think we should at least have it documented somewhere because it is needed to debug the scala backend.

bindings.xjb
michael-hoke: This is a manually generated file. Which license header should be added?

bindings.xjb
michael-hoke: Trial and error with the jaxb build. I would try to build it, and it would fail.

build.sbt
michael-hoke: I'm not thrilled with this either, but I don't know another way of doing it. This is here because the sbt-xjc plugin invokes java from the command line, and it needs to find the libraries downloaded as dependencies during the build somehow. These lines are adding the required jars to the classpath.

resources/xsd/DFDL_part3_model.xsd
michael-hoke: Yes, that's where they came from. I will give that a shot.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
michael-hoke: The intent of this feature was just to implement very basic TDML test cases into the extension. I agree that the code should not be duplicated. I'm hoping that someone will be willing to come along to refactor this to reference the TDML code in the main daffodil repo (either directly, or refactor it into a separate library), but that seemed out of scope for this implementation.

The main use case for this feature as it exists now is for a user of the extension to use the 'Generate TDML' command to create a TDML file that can be added to a DFDL schema and an input file, and sent to the user list for debugging purposes. Then, someone should be able to take those files and execute the test case from the TDML file to get the same execution as the original poster.

build.sbt
stevedlawrence: > find the libraries downloaded as dependencies

These are dependencies of xjc correct? Seems odd that this would be needed, I would expect xjc to handle that. I wonder if we are using/configuring it incorrectly? None of the examples in the xjc repo require manually listing xjc dependencies.

michael-hoke: I also would expect it to handle this automatically, but it was throwing ClassNotFoundExceptions before I added all of this. Maybe there is some extra configuration needed, but nothing jumps out at me from their documentation.

@arosien is there a smarter way to have the java process recognize/find the jars?

stevedlawrence: It looks like sbt-xjc is already adding jars to the classpath:

https://github.com/sbt/sbt-xjc/blob/master/src/main/scala/com/github/retronym/sbtxjc/SbtXjcPlugin.scala#L96

I do see references to Ivy in that file, it's possible it doesn't work with Coursier? You might try settting `useCoursier := false`, in your sbt configs? We do that for daffodil because something doesn't work with coursier.

michael-hoke: I added that setting, commented out the classpath lines I added, and I'm getting the ClassNotFoundExceptions. It looks like it's re-downloading the dependencies, so I'm assuming that means it's not using couriser anymore. I'll dig around with this for a bit to see if I can get it to work.

michael-hoke:
There are a couple of places that I could have put this, so I'll just put it as a separate comment at the end:

This last push includes the bigger changes - removing hard coded classpath creation and downloading the daffodil-lib jar to get the schema files

I wanted to note that what appears to be duplication of dependencies in the libraryDependencies and xjcLibs keys is intentional and necessary. It appears that including a jar in libraryDependencies is necessary for downloading it, and including a jar in xjcLibs is necessary for properly setting up the classpath for the JAXB generation.

Anyways, it looks like the Windows build is failing again because paths are different somewhere...

.vscode/launch.json
stevedlawrence: What about "Daffodil Debug Server"? I'm not sure if most people will know what a "DAP" server is. Also, if it's only for extension developers, maybe it should be hidden behind an "advanced settings" option or something?

bindings.xjb
stevedlawrence: ALv2

.vscode/launch.json
arosien: You're right it's only for extension developers, and one would only see this if they cloned the repo and were using VS Code, so I think it's ok to be specific about its DAP purpose.

stevedlawrence: > would only see this if they cloned the repo and were using VS Code

How does that work? I don't see anything that only shows this menu for extension developers?

stevedlawrence: Ah, maybe I just had to look up, is this part of the "Extension" configuration? Is that only enabled by developers somehow?

michael-hoke:
Couldn't figure out how to get Windows/Mac to build. Linux is the only OS that builds right now. As best as I can tell, the reason that it isn't building lies with the extraction itself. I checked the permissions of the server/sbtXjc directory on Windows, and those seemed okay to me. Even after I changed them to include every available option, it still wouldn't build. The permissions on the jar in the ~/.ivy/cache folder looked okay as well. Other than the IO.unzip extraction that is currently being used, I also made an attempt using Java's ZipFile class. This did not change the output of the build.

NolanMatt:
Trying to pick this up while working on Windows since its a problem, and when I attempt to build, I get an error saying

 error while decoding C:\Users\nolanm\Documents\GitHub\daffodil-vscode\server\sbtXjc\resources\xsd\org\apache\daffodil\xsd\xml.xsd with UTF-8: C:\Users\
nolanm\Documents\GitHub\daffodil-vscode\server\sbtXjc\resources\xsd\org\apache\daffodil\xsd\xml.xsd (The system cannot find the path specified)←[0m←[0J

This happens for every xsd file because the directory doesn't exist.  After server/sbtXjc/resources/xsd the directory structure is flat and none of org/apache/daffodil/xsd exist.  Is this whats happening on windows with other people as well?

michael-hoke:
I'm seeing the same thing now - thought it was just something with my local changes. You can run the following to get the branch to a more stable state:

git reset hard d0de0700e

If that's not enough of the hash, you can run `git status` and find the full hash. I'll be pushing more changes soon to test the windows/mac builds, and you're probably not going to want those, so I'd stay based off of that commit. If you get it working on Windows, we can figure out the easiest way to get your changes up.

michael-hoke:
I think I have the extraction working properly on non-Linux platforms, but it looks like it's failing a bit further in the compilation process now:

- If I leave the generator as a resourceGenerator, it doesn't look like Compile / xjc / sources is able to pick up the generated sources because they're generated after the sources list is compiled.

- If I change it to a sourceGenerator, sbt tries to read the XML as source files, which causes syntax errors.

- The bindings file is using Linux slashes/paths in the schemaLocation. I'm not sure yet if that matters (it looks like the plugin might automatically switch), but I'll note it here so it doesn't get forgotten.

NolanMatt:
After pulling in your latest changes @michael-hoke I was able to successfully build the vsix on windows.

michael-hoke:
Can you check the hash?? If it's d37932a577, that's mildly concerning that it would be working on your local machine but not on the git runners... Maybe the remaining issue is in the runner rather than the build.sbt? I feel like this is probably unlikely, though.

Can you check the following directory - ./server/sbtXjc/target/scala-2.12/src_managed_cxf/org/apache/daffodil/tdml

That should contain the Java files compiled from the xjc plugin. I've got ~150 generated files in that directory.

NolanMatt:
$ git log
commit d37932a577a08c092891b3b8b1c99b31932f14b5 (HEAD -> pr/253, github-desktop-michael-hoke/daffodil-vscode-tdml)
Author: Michael Hoke <michael.hoke@nteligen.com>
Date:   Mon Nov 28 13:16:44 2022 -0500

nolanm@EVC-D4269 MINGW64 ~/Documents/GitHub/daffodil-vscode/server/sbtXjc/target/scala-2.12/src_managed_cxf/org/apache/daffodil/tdml (pr/253)
$ ls | grep java | wc -l
150

michael-hoke:
Thanks, that's what it should be.... Wonder if it actually is an issue with the runner then?

michael-hoke:
Looks like it does build in Windows/Mac if I take out the Windows-specific portion of the CI. It's failing in the Test portion now, which makes sense. The settings for XJC that start with Compile / ... need to be duplicated to Test / ...

Hopefully, there's a way to do that without duplicating all of the code

michael-hoke:
I think we're almost there. I had it building successfully in CI before rebasing on the current apache:main. It looks like the tests that were added to main need to be updated - doesn't look like it should be that much of an issue.

Thanks again @NolanMatt for checking the Windows build.

NolanMatt:
@michael-hoke Davin mentioned next step is to hook the feature into the test suite.  It seems like that's what you ended up doing yesterday.  Am I right in thinking that or is that still to be done?

michael-hoke:
That's what I'm working on now. I'm just not sure on the extent right now. I need to talk to Davin to figure out what needs to be done.

.vscode/launch.json
arosien: The .vscode/launch.json file is used to define runnable programs within the current VS Code project. It doesn't necessarily have anything to do with extension development or not.

.vscode/launch.json
```diff
@@ -44,6 +61,15 @@
 				"${workspaceFolder}/out/tests/**/*.js",
 			],
 			"preLaunchTask": "npm: compile"
+		},
+		{
+			"type": "scala",
+			"request": "launch",
+			"name": "DAPodil",
```
arosien, line 68:
```suggestion
			"name": "Daffodil Debug Server",
```

build.sbt
```diff
@@ -62,6 +64,8 @@ lazy val ratSettings = Seq(
   ratFailBinaries := true
 )
 
+lazy val xjcTask = taskKey[Seq[File]]("Create JAXB Bindings for DFDL Schemas")
```
arosien, line 67: Suggestions: This seems to be unused and not necessary. Delete?

.vscode/launch.json
arosien:
```suggestion
			"name": "Daffodil Debug Server",
```

build.sbt
```diff
@@ -91,3 +95,85 @@ lazy val core = project
     packageName := s"${name.value}-$daffodilVer"
   )
   .enablePlugins(commonPlugins: _*)
+  .dependsOn(sbtXjcProject)
+  .aggregate(sbtXjcProject)
+
+lazy val sbtXjcProject = project
```
arosien, line 101: Question: I'm not sure if this needs to be a separate sbt project. Was there a particular reason why it was done this way?

I can see if locally I can move this stuff into the core project.

bindings.xjb
```diff
@@ -0,0 +1,78 @@
+<!--
```
arosien, line 1: Suggestion: can this file be moved into the server directory? The example project at https://github.com/sbt/sbt-xjc/blob/master/src/sbt-test/sbt-xjc/xero/build.sbt puts both the xjb and xsd files under src/main/resources.

server/core/src/main/scala/org.apache.daffodil.tdml/TDML.scala
arosien: Sorry, I mean I'm actually suggesting changing the return type to `Option[(String, String)]` so the caller doesn't have to check for empty strings.

server/core/src/main/scala/org.apache.daffodil.tdml/TDMLWrapper.scala
```diff
@@ -0,0 +1,119 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.tdml
+
+import java.io.File
+import java.nio.file.Path
+import java.nio.file.Paths
+
+object TDMLWrapper {
```
arosien, line 24: Suggestion: I think these should be moved to the `TDML` companion object since they are just overloads of existing methods.

server/core/src/main/scala/org.apache.daffodil.tdml/TDML.scala
```diff
@@ -0,0 +1,214 @@
...
+    val basePath = Paths.get(tdmlPath).toAbsolutePath().getParent()
+
+    val testCaseList = JAXBContext
+      .newInstance(classOf[TestSuite])
+      .createUnmarshaller()
+      .unmarshal(new File(tdmlPath))
+      .asInstanceOf[TestSuite]
+      .getTutorialOrParserTestCaseOrDefineSchema()
+
+    testCaseList.forEach { tc =>
```
arosien, line 183: This isn't Scala.

server/core/src/main/scala/org.apache.daffodil.tdml/TDML.scala
```diff
@@ -0,0 +1,214 @@
...
+                      basePath + File.separator + doc
+                        .getContent()
+                        .get(0)
+                        .asInstanceOf[JAXBElement[DocumentPartType]]
+                        .getValue()
+                        .getValue()
+                    )
+                    .normalize()
+                    .toString()
+                  return (schemaPath, dataPath)
```
arosien, line 204: Worry: this is an early return, so if there are multiple test cases then only the first is executed.

src/tests/data/png.dfdl.xsd
```diff
@@ -0,0 +1,764 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Shanedell, line 1: As I have learned from @stevedlawrence , do not add this file into the repo. Its license is not compatible with ALv2 I believe. You will instead want to make a custom schema for testing, something similar to what we have in `main`.

arosien:
I get this error when activating the Launch Config Wizard via the "gear" icon next to the open file tabs:

```
2023-03-08 10:46:41.855 [error] SyntaxError: Unexpected token } in JSON at position 790
	at JSON.parse (<anonymous>)
	at LaunchWizard.getWebViewContent (/Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11760:29)
	at /Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11689:43
	at Generator.next (<anonymous>)
	at /Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11562:71
	at new Promise (<anonymous>)
	at __awaiter (/Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11558:12)
	at createWizard (/Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11686:12)
	at Object.<anonymous> (/Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11575:19)
	at Generator.next (<anonymous>)
	at /Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11562:71
	at new Promise (<anonymous>)
	at __awaiter (/Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11558:12)
	at /Users/arosien/nteligen/daffodil-vscode/dist/ext/extension.js:11574:87
	at l.h (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:96:108784)
	at l.$executeContributedCommand (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:96:109688)
	at c.N (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:106:11223)
	at c.M (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:106:10941)
	at c.H (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:106:10034)
	at c.G (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:106:9015)
	at /Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:106:7803
	at f.invoke (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:63:145)
	at g.deliver (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:63:2066)
	at h.fire (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:63:1704)
	at p.fire (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:72:14907)
	at /Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:122:16556
	at f.invoke (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:63:145)
	at g.deliver (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:63:2066)
	at h.fire (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:63:1704)
	at p.fire (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:72:14907)
	at MessagePortMain.<anonymous> (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:122:14682)
	at MessagePortMain.emit (node:events:526:28)
	at MessagePortMain._internalPort.emit (node:electron/js2c/utility_init:5:364)
	at Object.callbackTrampoline (node:internal/async_hooks:130:17) launch.config {"value":"asf.apache-daffodil-vscode","c":"asf.apache-daffodil-vscode"}
```

.vscode/launch.json
```diff
@@ -18,6 +18,30 @@
 {
 	"version": "0.2.0",
 	"configurations": [
+		{
```
arosien, line 21: I don't think this should be in the launch config.

arosien:
@Shanedell would you be available to take a look at this? If you don't feel too comfortable going in to the Scala you could just check it out and try out the feature.

@mbeckerle, @stevedlawrence need your 👀 also.

Shanedell:
@scholarsmate Don't pull main into branches. The workflow is to rebase and force push to sync branches with main. @stevedlawrence Can correct me if I am misspeaking but I believe we don't want these merged in commits inside of branches or the repository.

michael-hoke:
I'll go in and do some rebasing magic to rebase this branch and move the commit out to my fork's main.

Shanedell (approved):
+1 👍 . I looked over everything and it looks good. Worked good locally for me through debugging. I was also able to create the VSIX, install it and get the option for adding TDML as well as it seemingly running fine. The scala and all other code looked good to me but would still like @mbeckerle or @stevedlawrence to give their thoughts on this PR before merging this into `main`.

One minor thing is for `yarn test` maybe have the scala tests run before the node tests, as if these fail we should not run the node tests since it depends on the debugger.

michael-hoke:
Thanks for looking @Shanedell 

The force push that just happened was reverting the merge commit and updating branches to point to the latest apache:main.

package.json
```diff
@@ -35,7 +35,7 @@
     "package-setup": "node -e \"require('./build/scripts/package.ts').setup()\"",
     "package-create": "node -e \"require('./build/scripts/package.ts').create()\"",
     "pretest": "yarn compile && yarn webpack",
-    "test": "node ./out/tests/runTest.js",
+    "test": "node ./out/tests/runTest.js && sbt test",
```
Shanedell, line 38:
```suggestion
    "test": "sbt test && node ./out/tests/runTest.js",
```

This is the only small thing I had. I believe we want `sbt test` first since some of the node tests depend on the debugger so if it fails an sbt test we should not try to run the node ones.

PR #274, opened 2022-08-31
Multiple Updates

### LINKED ISSUE #266: When debug session ends, the binary display of hex data goes away
When debug session ends, the binary display of hex data goes away. Why? I might be very interested in looking at that still. It can be marked inactive like the infoset file is, but shouldn't just be closed. You may not want to fix this soon because I expect the hex display will be replaced by the data editor in a view mode at some point, but anyway, it shouldn't auto-close just because the parse ended.

### LINKED ISSUE #267: Config Wizard text didn't display in a useable way in light UI mode
Config Wizard - did not work for me - text didn't display in a way I could use this. With dark theme I could use it.

Shanedell:
Multiple Updates:

- Make launch wizard display properly when using light themes.
- No longer closer the hexview once debug has ended.

Closes #266
Closes #267

### DIFF
--- a/src/styles/styles.css
+++ b/src/styles/styles.css
@@ -86,7 +86,6 @@
 }
 
 .setting-div {
-  color: white;
   font-size: 16px;
   cursor: pointer;
 }
--- a/src/hexView.ts
+++ b/src/hexView.ts
@@ -101,11 +101,7 @@ export class DebuggerHexView {
   // Overriden onTerminatedDebugSession method
   onTerminatedDebugSession(session: vscode.DebugSession) {
     if (session.type === 'dfdl') {
-      vscode.window.visibleTextEditors.forEach((editior) => {
-        if (editior.document.fileName === this.hexFile) {
-          editior.hide() // method is deprecated but is only way to close specific editor not just the active one
-        }
-      })
+      this.decorator.dispose()
       this.dataFile = ''
       this.bytePos1b = -1
     }
--- a/src/launchWizard/launchWizard.ts
+++ b/src/launchWizard/launchWizard.ts
@@ -385,7 +385,7 @@ class LaunchWizard {
       <script nonce="${nonce}">
         ${scriptData}
       </script>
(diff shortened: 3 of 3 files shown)


src/styles/styles.css
```diff
@@ -86,7 +86,7 @@
 }
 
 .setting-div {
-  color: white;
+  /* color: white; */
```
scholarsmate, line 89: Don't bother commenting this out, just delete this line.

scholarsmate (changes requested):

scholarsmate:
@stevedlawrence and @mbeckerle, will you guys review this PR?  This addresses a couple of (minor) issues that Mike brought to our attention in his review of version 1.1.0 when it went up for a release vote.

stevedlawrence (approved):
+1

Note, In general I would really prefer a single PR per issue, even when they are small one liners. When you include multiple changes in a single PR the commit subject is something generic like "multiple updates" which is very unhelpful when looking back at history. I would much rather see two PRs and commits with appropriate commit messages.

Shanedell:
> +1
> 
> Note, In general I would really prefer a single PR per issue, even when they are small one liners. When you include multiple changes in a single PR the commit subject is something generic like "multiple updates" which is very unhelpful when looking back at history. I would much rather see two PRs and commits with appropriate commit messages.

Yeah this makes sense I will make sure to do them as single PRs from now one.

scholarsmate:
@mbeckerle, will you review this PR, we need your vote, and these address issues that you've found during the last release vote.

tuxji (approved):
+1

I suggest removing line 89 from style.css in a local commit and rebasing, squashing, and force-pushing the PR back to 1 commit with an improved commit message which says something like "Improve launch wizard display, don't close hexview" instead of "Multiple Updates".  Then git history will look better after you merge the PR.

Shanedell:
@tuxji Does the commit look better?

mbeckerle (approved):
+1

scholarsmate (approved):
+1

Looks good, thanks!

PR #278, opened 2022-09-04
Choice fix

rt320:


stevedlawrence:
This PR has merge commits and and a lot more changes in it than just a fix to choices. Please rebase your changes to the latest asf/main branch and remove the merge commits and update the PR>

We have a pretty thoroughly documented workflow for Daffodil that should be used for this repo as well:

https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow

That helps to avoid this issues with merge commits and unintended changes added to a PR. I would especially recommend never running `git pull`. For some reason it's what everyting online says to use, but it really just causes issues and should be avoided. The mentioned workflow never uses `git pull`.

stevedlawrence (changes requested):
-1 until the PR is fixed.

Its not even clear what this PR is about. The PR only says "Choice fix" with no description of the changes. And it includes three commits, one that seems unrelated to choices, and the other two being merge commits. This cannot be reviewed in its current state.

Issue #165, opened 2022-06-03
Autocompletion remove wholeLine checks


Shanedell:
I still don't understand these `wholeLine.includes` checks. The context to use for autocompletion doesn't have do with the line of text--it has to do with things like what is the name of the current tag that the cursor is inside, what are the values of it's attributes, etc. This just feels very fragile and likely to give the wrong results.

It seems like instead you want something like
```
if (nearestOpenItem.includes("element)) {
  if (nearesOpenItem.hasAttribute("name") || nearestOpenItem.hasAttribute("ref")) {
     ...
  }
}
```

A simple examle where this breaks, it's not uncommon for schemas to use "xsd" instead of "xs" (or even no prefix). In those cases looking for "xs:element" is going to fail.

I wont' block this merging for this, but hope it's updated in the future to improve how it determines the appropriate context for choosing completions.

_Originally posted by @stevedlawrence in https://github.com/apache/daffodil-vscode/pull/117#discussion_r889129269_

rt320:
VS Code has a built in interface for handling auto-complete. Whenever an autocomplete trigger is activated, the built in interface supplies a position in the document where the autocomplete was triggered.  With the position there is a method to get the "lineAt()" string.  The wholeLine variable contains the string returned by lineAt(). A more accurate variable name instead of "wholeLine" might have been "triggerLine".  Based on VS Code examples for autocompletion, the string method "contains()" was used to return true or false to a string comparison of possible XML components that were in the line where the autocomplete was triggered.  In the current autocomplete code no attempt is made to parse out individual XML components, string comparisons are the basis for determining what autocompletion items should be suggested.  An "openItem" is determined by parsing lines backwards to find the last end tag. Any XML components after that last end tag are considered to be "open". One relatively easy way to provide an alternative to the hardcoded namespace "xs" could be to use the JavaScript/TypeScript object "regExp" to define a regex that compares '<' + any characters + "element", and uses the regExp test() method to compare the regex to wholeLine and return true of false, similar to how the process currently works.  Another suggestion was to use regex to parse the namespace from wholeLine and then use the namespace value to build an end tag value to look for, I'm not sure how the search for the end tag would work.  The entire document could also be parsed into XML components and associated with positions and some method developed to query if a tag is open by position, although thought would have to be given to how to handle inserting attributes into already closed tags of existing schemas. There are probably any number of solutions. Each solution will have benefits and drawbacks, including length of development time. I'm not sure the best way to address this issue.

[...]

PR #283, opened 2022-09-07
xs:choice component modified to allow attributes, added tag close for…

rt320:
… xs:choice, fixed snippet return for choiceDispatchKey, fixed brace autocomplete for choice dispatchkey


stevedlawrence (approved):
+1

This suffers from issue #165, but appears to be consistent with existing code, so looks fine to me.

Issue #120, opened 2022-05-27
DFDL Language remove "xs:" prefix


Shanedell:
- This only needs to be done once Daffodil decided to fully move away from the prefix.
- Issue made based on comment --> https://github.com/apache/daffodil-vscode/pull/117#discussion_r883666880

Shanedell:
> One thing to consider is this "xs" prefix isn't mandatory and some schemas use different prefixes (usually "xsd" or no prefix). Ideally this would look at the in scope namespaces an find the prefix associated with the schema prefix.
> 
> Again, I won't block the merge for that, but hope those kinds of features are added in the future.


Also note this comment from @stevedlawrence

scholarsmate:
It appears there are still several instances in the TypeScript code where the `xs:` prefix remains.

PR #286, opened 2022-09-08
renamed wholeLine variable to triggerText

rt320:


stevedlawrence (changes requested):
-1

This does not address the problem described in #165. Even though you have renamed `wholeLone` to `triggerText`, it is still using the entire line as the autocompletion context. You cannot use an entire line and what strings appear in that line as the context for autocompletion. It is possible and not uncommon for large chunks of schema to be on a single line and so the completion suggestions will be incorrect.

To determine the correct context, you need to scan backwards character by character from the cursor position to determine what attribute/tag that you are in.

rt320:
In utils.ts, the function nearest open looks back at each previous line in the document to find the nearest open tag.  Can you provide me an example of a dfdl schema attribute/tag where the current code doesn't work so I understand what you see as the issue?

stevedlawrence:
I'm not totally sure if is related to the whole line issue, but if I have this:

```xml
<xs:element name="foo" |
```

And press Ctrl+Space, then I correctly get a list DFDL properties that can be used for `xs:element` tags.

However, if I have this:
```xml
<xs:sequence><xs:element name="foo" |
```
And press Ctrl+Space, then I incorrectly get a list of schema tags (e.g. xs:element, xs:group, etc.). This may or maybe not be a wholeLine issue, but it does seem that the extension is not correctly figuring out the context when xs:element isn't the only thing on the line.

rt320:
I'm not sure how you were able to enter the above code.  When I typed the '>' character for "<xs:sequence", auto complete added the sequence closing tag and moved the cursor to the next line.  After I typed "<xs:element name="foo" |" then a space, a list of attribute suggestions displayed. 
image
I also tried to cut and paste the single line from your comment into the dfdl editor, but it broke it in to 3 lines.
image
The cursor was at the end of the sequence tag instead of after the '|', but when I moved the cursor up to after the bar character and typed space, the same attributes were suggested.
image

rt320:
I also just tried deleting the sequence end tag, and backspacing to move the element tag to the end of the sequence tab.  When I placed the cursor after the bar and typed control space the attribute suggestions appear.
image

rt320:
Although I just noticed that some of the higher level components also appear after the attributes.  If someone were to delete the sequence auto-complete tag and backspace the element tag to the same line as the beginning sequence tag, the auto-suggestions would be incorrect.  Is that what you feel needs to be fixed

stevedlawrence:
> I also just tried deleting the sequence end tag, and backspacing to move the element tag to the end of the sequence tab.

Yes this is what I did. Note that the bar was just to indicate the position of the cursor when I pressed Ctrl+Space. I didn't have an actual bar in my data.

> Although I just noticed that some of the higher level components also appear after the attributes. 

The `dfdl:` suggestions are not attributes, those are tags e.g. `<dfdl:defineVariable ...>`. This is what I was seeing where it was not giving any suggestions for `xs:element` attributes, but instead giving suggestions as if it were not in an element at all.

I admit the steps to reproduce are a bit contrived, but it shows that this isn't using the right context when trying to figure out what to suggest. The `wholeLine` stuff works because the schemas you write don't put things on multiple lines (your autocompletion actually makes it a bit difficult to do), but that isn't always the case.

That said, most schemas aren't going to put multiple tags on a single line, which is why I don't considering using `wholeLine` a blocker. But the logic isn't correct and so simply renaming the variable doesn't actually fix the underlying bug.

rt320:
@stevedlawrence @mbeckerle I am making some changes to address both issues mentioned above. I haven't fully tested the changes and I haven't made all the changes for missing xs: prefixs. Here's is an animated GIF that shows how the changes are working. One thing to note, when typing any component in a schema the auto-complete will always add the prefix.  The change makes the code work without a prefix when determining suggestions, but it will always add the prefix when a suggestion is taken. Is that acceptable? or should auto-complete leave off prefixes in instances where the schema isn't using them?  Would it be just "xs:" prefixes that are left off or all prefixes?  Also I don't have a way to trigger the end tag for sequence, or a way to know when it should be added/triggered, so it was added manually.
ContrivedTest2

stevedlawrence:
You don't need to make the prefix change as part of the whole line fix. In fact, I suggest you don't. Keep separate features/bug fixes in separate PRs. For now, it's fine to always assume the xs prefix will be used, and that can be fixed in a separate PR for issue #120.

As to your questions about a prefix, the main discussion for how to handle that is best done in #120. That said, ideally it would find and use the prefix from the first namespace declaration associated with the XML Schema namespace (i.e. http://www.w3.org/2001/XMLSchema). So for example if your schema had this:

```xml
<schema xmlns="//www.w3.org/2001/XMLSchema" ... >
```
Then it would autocomplete with no namespace prefix. But if you had

```xml
<xs:schema xmlns:xs="//www.w3.org/2001/XMLSchema" ...>
```
Then it would autocomplete with the xs prefix. Note that as Mike pointed out, you can have multiple prefixes for the same namespace, e.g.:

```xml
<schema
  xmlns="//www.w3.org/2001/XMLSchema"
  xmlns:xs="//www.w3.org/2001/XMLSchema" ...>
```
In this example, I think we want to autocomplete with the first namespace prefix that matches the XML schema namespace,  i.e. no prefix

mbeckerle:
> In this example, I think we want to autocomplete with the first namespace prefix that matches the XML schema namespace, i.e. no prefix

This is in fact the behavior I have seen in other XML editors (IntelliJ IDEA). First xmlns decl in text lexical order "wins".

stevedlawrence:
Another option, use the prefix that is used on the `schema` element. Whatever prefix the user picked for that is probably what they want to use for the rest of the schema. So `<xs:schema ...` means we use an `xs` prefix, and `<schema` means we use no prefix. And maybe we just make an assumption that the namespace prefixes are defined correctly.

scholarsmate:
I think this is a good improvement and maybe we can spend a _little_ more time on this to get it into a reasonably good working state, but time is a very finite resource.  We would like to get this included in the next release (1.2.0), which we'd really like to get out this month.  We can use @rt320 to help us improve the test cases to cover things like Intellisense so when we make improvements, we can have some degree of confidence that the improvements didn't cause an unintended regression.

PR #292, opened 2022-09-12
Added function to determine prefix in schema definition. Added calls …

rt320:
…to function to determine schema prefix. Modified top level snippets to return variable with prefix value plus the preset snippet text. Renamed wholeLine variable name to triggerText.

### DIFF
--- a/src/language/providers/attributeCompletion.ts
+++ b/src/language/providers/attributeCompletion.ts
@@ -25,6 +25,7 @@ import {
   checkSequenceOpen,
   checkElementOpen,
   checkSimpleTypeOpen,
+  getNsPrefix,
 } from './utils'
 
 import { attributeCompletion } from './intellisense/attributeItems'
@@ -75,31 +76,38 @@ export function getAttributeCompletionProvider() {
         document: vscode.TextDocument,
         position: vscode.Position
       ) {
-        const wholeLine = document
+        const triggerText = document
           .lineAt(position)
           .text.substr(0, position.character)
+        var nsPrefix = getNsPrefix(document, position)
         var nearestOpenItem = nearestOpen(document, position)
 
         if (
           !checkBraceOpen(document, position) &&
-          !wholeLine.includes('assert') &&
+          !triggerText.includes('assert') &&
           !nearestOpenItem.includes('none')
         ) {
           if (nearestOpenItem.includes('element')) {
             var preVal = ''
-            if (!wholeLine.includes('xs:element')) {
+            if (!triggerText.includes('<' + nsPrefix + 'element')) {
               if (lineCount(document, position) === 1) {
                 preVal = '\t'
               } else {
                 preVal = ''
               }
             }
-            var additionalItems = getDefinedTypes(document)
+            var additionalItems = getDefinedTypes(document, position)
 
+            // if (
+            //   checkLastItemOpen(document, position) &&
+            //   (triggerText.includes('<xs:element name="') ||
+            //     triggerText.includes('<xs:element ref="') ||
+            //     checkElementOpen(document, position))
+            // ) {
             if (
               checkLastItemOpen(document, position) &&
-              (wholeLine.includes('<xs:element name="') ||
-                wholeLine.includes('<xs:element ref="') ||
+              (triggerText.includes('<' + nsPrefix + 'element name="') ||
+                triggerText.includes('<' + nsPrefix + 'element ref="') ||
                 checkElementOpen(document, position))
             ) {
               return getCompletionItems(
@@ -134,7 +142,7 @@ export function getAttributeCompletionProvider() {
 
           if (nearestOpenItem.includes('sequence')) {
             var preVal = ''
-            if (!wholeLine.includes('xs:sequence')) {
+            if (!triggerText.includes(nsPrefix + 'sequence')) {
               if (lineCount(document, position) === 1) {
                 preVal = '\t'
               } else {
@@ -144,7 +152,7 @@ export function getAttributeCompletionProvider() {
 
             if (
               checkLastItemOpen(document, position) &&
-              (wholeLine.includes('<xs:sequence') ||
+              (triggerText.includes(nsPrefix + 'sequence') ||
                 checkSequenceOpen(document, position))
             ) {
               return getCompletionItems(
@@ -160,8 +168,8 @@ export function getAttributeCompletionProvider() {
             }
           }
 
-          if (wholeLine.includes('choice')) {
-            if (!wholeLine.includes('>')) {
+          if (triggerText.includes('choice')) {
+            if (!triggerText.includes('>')) {
               return getCompletionItems([
                 'dfdl:choiceLengthKind=',
                 'dfdl:choiceLength=',
@@ -173,10 +181,10 @@ export function getAttributeCompletionProvider() {
           }
 
           if (
-            wholeLine.includes('simpleType') ||
+            triggerText.includes('simpleType') ||
             checkSimpleTypeOpen(document, position)
           ) {
-            if (!wholeLine.includes('>')) {
+            if (!triggerText.includes('>')) {
               return getCompletionItems([
                 'dfdl:length=',
                 'dfdl:lengthKind=',
@@ -187,16 +195,16 @@ export function getAttributeCompletionProvider() {
             }
           }
 
-          if (wholeLine.includes('defineVariable')) {
+          if (triggerText.includes('defineVariable')) {
             var preVal = ''
-            if (!wholeLine.includes('dfdl:defineVariable')) {
+            if (!triggerText.includes('dfdl:defineVariable')) {
               if (lineCount(document, position) === 1) {
                 preVal = '\t'
               } else {
                 preVal = ''
               }
             }
-            var additionalItems = getDefinedTypes(document)
+            var additionalItems = getDefinedTypes(document, position)
 
             var xmlItems = [
               {
@@ -219,7 +227,7 @@ export function getAttributeCompletionProvider() {
               },
             ]
 
-            if (!wholeLine.includes('>')) {
+            if (!triggerText.includes('>')) {
               let compItems: vscode.CompletionItem[] = []
               xmlItems.forEach((e) => {
                 const completionItem = new vscode.CompletionItem(e.item)
@@ -241,7 +249,7 @@ export function getAttributeCompletionProvider() {
 
           if (nearestOpenItem.includes('setVariable')) {
             var preVal = ''
-            if (!wholeLine.includes('dfdl:setVariable')) {
+            if (!triggerText.includes('dfdl:setVariable')) {
               if (lineCount(document, position) === 1) {
                 preVal = '\t'
               } else {
@@ -253,7 +261,7 @@ export function getAttributeCompletionProvider() {
             xmlValue.insertText = new vscode.SnippetString('value="$1"$0')
             xmlValue.documentation = new vscode.MarkdownString('')
 
-            if (!wholeLine.includes('>')) {
+            if (!triggerText.includes('>')) {
               return [xmlValue]
             }
           }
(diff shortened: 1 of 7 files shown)


src/language/providers/utils.ts
```diff
@@ -251,46 +316,72 @@ export function checkSetVariableOpen(
...
+  var initialLineNum = position.line
+  var lineNum = 0
+  while (initialLineNum !== 0) {
+    const lineText = document
+      .lineAt(0)
+      .text.substr(0, document.lineAt(lineNum).range.end.character)
+    if (lineText.includes('<schema')) {
+      return nsPrefix
+    }
+    if (lineText.match(new RegExp('<.*:schema'))) {
```
stevedlawrence, line 333: `.*` is greedy so there could be some edge cases where it matches too much, for example:

```xml
<xs:schema xmlns:xs="..." xmlns:urn="foo:bar:schema" />
```

Instead we should use something like `[^ ]+` so it matches only non-space characters. Also, I think we can use a single regex to match both cases, and a capture group to get the matched prefix, which simplifies things a bit. For example:

```js
var prefixRegex = new Regex('<(|[^ ]+:)schema')
...
while(...) {
  const lineText = document.lineAt(lineNum)
  var prefixRegexMatch = prefixRegex.exec(lineText) 
  if (prefixRegexMatch != null) {
     return prefixRegexMatch[1]
  }
  ... 
}
```

The regex will match either an emtpy string followed by schema, or one or more non-space charcters followed by a colon followed by schema. And the prefix (including colon) is part of the capture group.

This isn't tested, but that's the idea.

src/language/providers/utils.ts
```diff
@@ -251,46 +316,72 @@ export function checkSetVariableOpen(
...
 }
 
+export function getNsPrefix(
+  document: vscode.TextDocument,
+  position: vscode.Position
+) {
+  var nsPrefix = ''
+  var initialLineNum = position.line
+  var lineNum = 0
+  while (initialLineNum !== 0) {
```
stevedlawrence, line 326: initialLineNum never changes, this isn't a good condition for breaking the loop. If a match is never found (e.g. and empty file) then this will be stuck in an infinite loop.

src/language/providers/attributeCompletion.ts
```diff
@@ -75,31 +76,38 @@ export function getAttributeCompletionProvider() {
...
             }
-            var additionalItems = getDefinedTypes(document)
+            var additionalItems = getDefinedTypes(document, position)
 
+            // if (
+            //   checkLastItemOpen(document, position) &&
+            //   (triggerText.includes('<xs:element name="') ||
+            //     triggerText.includes('<xs:element ref="') ||
+            //     checkElementOpen(document, position))
+            // ) {
```
stevedlawrence, line 106: Why was this commented out? I would prefer we don't change any behavior and limit this PR to only namespace changes.

src/language/providers/attributeCompletion.ts
```diff
@@ -266,21 +274,28 @@ export function getAttributeCompletionProvider() {
...
+function getDefinedTypes(
+  document: vscode.TextDocument,
+  position: vscode.Position
+) {
   var additionalTypes = ''
   var lineNum = 0
   const lineCount = document.lineCount
+  // const position = new vscode.Position(
+  //   lineCount,
+  //   document.lineAt(lineCount).range.end.character)
```
stevedlawrence, line 286: Remove comment.

src/language/providers/attributeCompletion.ts
```diff
@@ -266,21 +274,28 @@ export function getAttributeCompletionProvider() {
...
+  const nsPrefix = getNsPrefix(document, position)
```
stevedlawrence, line 287: We call `getNSPrefix` quite a bit. Can it be called once per autocomplete instantiation and shared mong all the calls so we aren't constantly scanning for the schema tag? Or maybe we already do taht? It's not clear to me what the entrypoint into autocomplete is.

src/language/providers/closeElement.ts
```diff
@@ -27,59 +27,71 @@ export function getCloseElementProvider() {
...
         const nearestOpenItem = nearestOpen(document, position)
-        const wholeLine = document
+        const triggerText = document
           .lineAt(position)
           .text.substr(0, position.character)
 
         if (
-          wholeLine.endsWith('>') &&
-          (wholeLine.includes('xs:element') ||
+          !triggerText.includes('</') &&
```
stevedlawrence, line 37: This is a change not related to prefix stuff I think? Please try to keep unrelated bug fixes separate.

src/language/providers/closeElement.ts
```diff
@@ -27,59 +27,71 @@ export function getCloseElementProvider() {
...
+              triggerText.includes('<' + nsPrefix + 'sequence')) ||
             nearestOpenItem.includes('sequence')
           ) {
-            insertSnippet('>\n\t$0\n</xs:sequence>', backpos)
+            insertSnippet('>\n\t$0\n</' + nsPrefix + 'sequence>', backpos)
           } else if (
-            (wholeLine.endsWith('>') && wholeLine.includes('xs:simpleType')) ||
+            (triggerText.endsWith('>') &&
+              triggerText.includes('<' + nsPrefix + 'choice')) ||
+            nearestOpenItem.includes('choice')
```
stevedlawrence, line 83: I think this is part of the choice PR that was just merged?

src/language/providers/intellisense/elementItems.ts
```diff
@@ -29,84 +29,84 @@ export const elementCompletion = (definedVariables, dfdlFormatString) => {
...
       },
       {
         item: 'xs:sequence',
-        snippetString: '<xs:sequence',
+        snippetString: '<' + nsPrefix + 'sequence',
       },
       {
         item: 'xs:choice',
-        snippetString: '<xs:choice$1>\n\t$0\n</xs:choice>',
+        snippetString: '<' + nsPrefix + 'choice',
```
stevedlawrence, line 101: More changes that aren't prefix related. This actually changes what is autocompleted to not add close tag. That should be done as a separate PR which a separate issue if that's what we want.

src/language/providers/utils.ts
```diff
@@ -25,31 +25,47 @@ export function insertSnippet(snippetString: string, backpos: vscode.Position) {
...
-    previousLine.includes('</') ||
-    previousLine.includes('/>') ||
-    ((wholeLine.includes('element') ||
-      wholeLine.includes('sequence') ||
-      wholeLine.includes('group') ||
-      wholeLine.includes('Variable')) &&
-      (wholeLine.includes('</') || wholeLine.includes('/>')))
-  ) {
-    return false
+  if (lineNum > 0) {
```
stevedlawrence, line 40: I'm not sure I understand what this change is? Doesn't seem related to prefixes.

src/language/providers/utils.ts
```diff
@@ -63,13 +79,13 @@ export function lineCount(
...
+    const triggerText = document
       .lineAt(lineNum)
       .text.substr(0, document.lineAt(lineNum).range.end.character)
     if (
-      wholeLine.includes('<xs:element') &&
-      !wholeLine.includes('</xs:element') &&
-      !wholeLine.includes('/>')
+      triggerText.match(new RegExp('<.*element')) &&
+      !triggerText.match(new RegExp('</.*element')) &&
+      !triggerText.includes('/>')
```
stevedlawrence, line 88: This is fragile, `.*` can match way too much text because of its greediness. Can you use the `"<" + nsPrefix + "element"` that you've used elsewhere?

src/language/providers/intellisense/elementItems.ts
```diff
@@ -29,84 +29,84 @@ export const elementCompletion = (definedVariables, dfdlFormatString) => {
       },
       {
```
stevedlawrence, line 30: Should we change the `xs:schema` autocomplete above to be opinionated about whether or not it uses the xs prefix? I think we're starting to move away from that prefix? We can do that in a separate PR, but it's worth considering doing now that we can support different prefixes. We may also want to reconsider some of these other things in the xs:schema tag. For example, I think we're also moving towards `elementFormDefault=unqualified`, and I'm not sure if xmlns:daf is used anywhere anymore--I believe that namespace has been deprecated.

stevedlawrence (changes requested):
The commit message should include an issue number that this PR closes. If there isn't an issue yet, please create one.

src/language/providers/attributeCompletion.ts
```diff
@@ -75,31 +76,38 @@ export function getAttributeCompletionProvider() {
         document: vscode.TextDocument,
         position: vscode.Position
       ) {
-        const wholeLine = document
+        const triggerText = document
```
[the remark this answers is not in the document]
rt320: I'll revert the triggerText variable name to wholeLine.

src/language/providers/utils.ts
```diff
@@ -187,20 +222,50 @@ export function checkSequenceOpen(
   return false
 }
 
+export function checkChoiceOpen(
```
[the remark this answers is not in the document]
rt320: I pulled in the latest code from main

src/language/providers/utils.ts
```diff
@@ -251,46 +316,72 @@ export function checkSetVariableOpen(
   return false
 }
 
+export function getNsPrefix(
```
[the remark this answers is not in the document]
rt320: I'll make the change.

src/language/providers/utils.ts
```diff
@@ -251,46 +316,72 @@ export function checkSetVariableOpen(
...
+  document: vscode.TextDocument,
+  position: vscode.Position
+) {
+  var nsPrefix = ''
+  var initialLineNum = position.line
+  var lineNum = 0
+  while (initialLineNum !== 0) {
+    const lineText = document
+      .lineAt(0)
+      .text.substr(0, document.lineAt(lineNum).range.end.character)
```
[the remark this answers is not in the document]
rt320: Yes

src/language/providers/utils.ts
```diff
@@ -251,46 +316,72 @@ export function checkSetVariableOpen(
   return false
 }
 
+export function getNsPrefix(
+  document: vscode.TextDocument,
+  position: vscode.Position
+) {
+  var nsPrefix = ''
```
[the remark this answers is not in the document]
rt320: I think I have to store the result from the substring function in a separate variable but I'll test it.

src/language/providers/intellisense/elementItems.ts
rt320: The change to not close the choice tag was done as part #283. It is in the code in current "main" branch"  This change removes the hardcode "xs:".

stevedlawrence: The change I'm saying is different here is that you've changed the autocomplete from

```xml
<xs:choice>
</xs:choice>
```

to

```xml
<xs:choice
```

I don't have a strong preference about what is autocompleted, but I dont' think that should change as part of this PR.

src/language/providers/utils.ts
rt320: I think it was part of the fix for incorrect suggestions for multiple tags on the same line not returning the correct suggestions.  If the schema tag was missing, as in test snippets of code, it prevent trying to get a negative line number. I missed removing it when removed the rest of that code. I'll remove it for this PR

src/language/providers/utils.ts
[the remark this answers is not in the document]
rt320: I removed nsPrefix

src/language/providers/intellisense/elementItems.ts
rt320: Due to the limited amount of time I have available for this project in the next few months, I think it would be better to do those changes as a separate PR.

rt320:
It was easier to pull the latest branch and make the changes there than to rebase the code is this PR, I'm going to close and delete this PR and replace it with the new code.

src/language/providers/intellisense/elementItems.ts
stevedlawrence: Agreed, that's fine for a separate PR.

PR #294, opened 2022-09-14
Resolves Remove Prefixes Issue #120

rt320:
-add function to parse prefix from schema tag and return the prefix plus a colon or return nothing -added calls to the new function
-modified the intellisense snippets to prepend the prefix value to the auto complete return element closes issue #120

### DIFF
--- a/src/language/providers/utils.ts
+++ b/src/language/providers/utils.ts
@@ -17,6 +17,8 @@
 
 import * as vscode from 'vscode'
 
+const schemaPrefixRegEx = new RegExp('</?(|[^ ]+:)schema')
+
 // Function to insert snippet to active editor
 export function insertSnippet(snippetString: string, backpos: vscode.Position) {
   vscode.window.activeTextEditor?.insertSnippet(
@@ -31,15 +33,11 @@ export function checkLastItemOpen(
   position: vscode.Position
 ) {
   var lineNum = position.line
-  const wholeLine = document
-    .lineAt(lineNum)
-    .text.substr(0, document.lineAt(lineNum).range.end.character)
+  const wholeLine = document.lineAt(lineNum).text
   while (wholeLine.length === 0) {
     --lineNum
   }
-  const previousLine = document
-    .lineAt(lineNum)
-    .text.substr(0, document.lineAt(lineNum - 1).range.end.character)
+  const previousLine = document.lineAt(lineNum).text
   if (
     previousLine.includes('</') ||
     previousLine.includes('/>') ||
@@ -61,15 +59,14 @@ export function lineCount(
 ) {
   var lineNum = position.line
   var lineCount = 0
+  const nsPrefix = getXsdNsPrefix(document, position)
   while (lineNum !== 0) {
     --lineNum
     ++lineCount
-    const wholeLine = document
-      .lineAt(lineNum)
-      .text.substr(0, document.lineAt(lineNum).range.end.character)
+    const wholeLine = document.lineAt(lineNum).text
     if (
-      wholeLine.includes('<xs:element') &&
-      !wholeLine.includes('</xs:element') &&
+      wholeLine.includes('<' + nsPrefix + 'element') &&
+      !wholeLine.includes('</' + nsPrefix + 'element') &&
       !wholeLine.includes('/>')
     ) {
       return lineCount
@@ -83,10 +80,9 @@ export function nearestOpen(
   position: vscode.Position
 ) {
   var lineNum = position.line
+  const nsPrefix = getXsdNsPrefix(document, position)
   while (lineNum !== -1) {
-    const wholeLine = document
-      .lineAt(lineNum)
-      .text.substring(0, document.lineAt(lineNum).range.end.character)
+    const wholeLine = document.lineAt(lineNum).text
     if (wholeLine.includes('element') && !wholeLine.includes('/>')) {
       if (checkElementOpen(document, position)) {
         return 'element'
@@ -101,8 +97,8 @@ export function nearestOpen(
       }
     } else if (wholeLine.includes('group')) {
       if (
-        wholeLine.includes('<xs:group') &&
-        !wholeLine.includes('</xs:group') &&
+        wholeLine.includes('<' + nsPrefix + 'group') &&
+        !wholeLine.includes('</' + nsPrefix + 'group') &&
         !wholeLine.includes('/>') &&
         !wholeLine.includes('/')
       ) {
@@ -135,25 +131,24 @@ export function checkElementOpen(
   document: vscode.TextDocument,
   position: vscode.Position
 ) {
+  const nsPrefix = getXsdNsPrefix(document, position)
   var lineNum = position.line
   while (lineNum !== -1) {
-    const wholeLine = document
-      .lineAt(lineNum)
-      .text.substr(0, document.lineAt(lineNum).range.end.character)
+    const wholeLine = document.lineAt(lineNum).text
     if (
-      wholeLine.includes('<xs:element') &&
+      wholeLine.includes('<' + nsPrefix + 'element') &&
       (wholeLine.includes('>') ||
-        wholeLine.includes('</xs:element') ||
+        wholeLine.includes('</' + nsPrefix + 'element') ||
         wholeLine.includes('/>'))
     ) {
       return false
     }
-    if (wholeLine.includes('</xs:element>')) {
+    if (wholeLine.includes('</' + nsPrefix + 'element')) {
       return false
     }
     if (
-      wholeLine.includes('<xs:element') &&
-      !wholeLine.includes('</xs:element') &&
+      wholeLine.includes('<' + nsPrefix + 'element') &&
+      !wholeLine.includes('</' + nsPrefix + 'element') &&
       !wholeLine.includes('/>') &&
       !wholeLine.includes('>')
     ) {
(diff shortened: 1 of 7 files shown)


src/language/providers/utils.ts
```diff
@@ -223,15 +230,16 @@ export function checkSimpleTypeOpen(
...
   var lineNum = position.line
   while (lineNum !== 0) {
     const wholeLine = document
       .lineAt(lineNum)
       .text.substr(0, document.lineAt(lineNum).range.end.character)
 
     if (
-      wholeLine.includes('<xs:simpleType') &&
-      !wholeLine.includes('</xs:simpleType') &&
+      wholeLine.includes('<' + nsPrefix + 'simpleType') &&
```
mbeckerle, line 241: Since this is everywhere, why not define an object that can be called like
`wholeLineElt('simpleType')`
which does this 
`wholeLineIncludes('<' + nsPrefix + arg)`
So that the namespace prefix stuff is suppressed.

src/language/providers/utils.ts
```diff
@@ -283,6 +291,33 @@ export function checkSetVariableOpen(
...
+) {
+  var initialLineNum = position.line
+  var lineNum = 0
+  while (initialLineNum !== 0 && lineNum <= initialLineNum) {
+    const lineText = document
+      .lineAt(lineNum)
+      .text.substr(0, document.lineAt(lineNum).range.end.character)
+    // returns an empty prefix value
+    if (lineText.includes('<schema')) {
+      return ''
```
mbeckerle, line 306: This is fine. Basically following the lead of how the schema element is prefixed.

mbeckerle (approved):
+1

I have one suggestion to clean up and eliminate the detail about nsPrefix having to show up everywhere.

src/language/providers/utils.ts
stevedlawrence: Might be worth doing that as part of the fix for #165. I suspect that's going to require a decent refactor anyways.

src/language/providers/utils.ts
```diff
@@ -283,6 +291,33 @@ export function checkSetVariableOpen(
...
+export function getNsPrefix(
+  document: vscode.TextDocument,
+  position: vscode.Position
+) {
+  var initialLineNum = position.line
+  var lineNum = 0
+  while (initialLineNum !== 0 && lineNum <= initialLineNum) {
+    const lineText = document
+      .lineAt(lineNum)
+      .text.substr(0, document.lineAt(lineNum).range.end.character)
```
stevedlawrence, line 303: Doesn't this substring just getting the whole line? Why not just use this to get the whole line:

```js
const lineText = document.lineAt(lineNum).text`
```

src/language/providers/utils.ts
```diff
@@ -283,6 +291,33 @@ export function checkSetVariableOpen(
...
+    if (lineText.includes('<schema')) {
+      return ''
+    }
+    // returns the prefix attached to the schema element plus a colon
+    if (lineText.match(new RegExp('<.*:schema'))) {
+      return lineText.substring(
+        lineText.indexOf('<') + 1,
+        lineText.indexOf(':') + 1
+      )
+    }
```
stevedlawrence, line 314: In the other review I suggested a regex that matches both `<foo:schema` and `<schema` so you don't need to do two scans, as well as using a capture group so you don't need to use substring. Did that not work?

Also, maybe move this RegExp to some global state so it doesn't need to be recompiled for iteration everytime this function is called? Feels like taht could be cause a performance hit?

src/language/providers/utils.ts
```diff
@@ -283,6 +291,33 @@ export function checkSetVariableOpen(
   return false
 }
 
+export function getNsPrefix(
```
stevedlawrence, line 294: Please add a comment that this returns the prefix + a colon if the prefix is not the empty string. Based on the name I wouldn't normally expect this function to return the colon since  namespace prefix is usually just the stuff before the colon. But I understand that including the colon makes the callers of this function much cleaner, so I think it just needs to be documented.

stevedlawrence (commented):
minor suggestions

src/language/providers/utils.ts
rt320: document.lineAt() returns a TextLine.  The includes() property isn't available on a TextLine, It is available on a string, which is what is returned by the proprty text.substr().
image

stevedlawrence: but `TextLine` has a `.text` property, the [TextLine doc](https://code.visualstudio.com/api/references/vscode-api#TextLine) says:

> text: string
>
> The text of this line without the line separator characters.

So calling `document.lineAt(lineNum).text` should get the entire line as a string.

src/language/providers/utils.ts
rt320: added additional comments

stevedlawrence (commented):
Minor suggestions.

Also, in the future, please do not ammed updates and force puche. Just keep making new commits--we can squash them all into a single commit at the end of the review. It makes it easier to see what has changed by looking at the new commits.

This workflow is described in detail here: https://cwiki.apache.org/confluence/display/DAFFODIL/Code+Contributor+Workflow

src/language/providers/utils.ts
```diff
@@ -283,6 +293,31 @@ export function checkSetVariableOpen(
...
+  position: vscode.Position
+) {
+  var initialLineNum = position.line
+  var lineNum = 0
+  while (initialLineNum !== 0 && lineNum <= initialLineNum) {
+    const lineText = document
+      .lineAt(lineNum)
+      .text.substr(0, document.lineAt(lineNum).range.end.character)
+    // returns either empty prefix value or a prefix plus a colon
+    if (lineText.includes('schema')) {
```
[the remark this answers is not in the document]
rt320: I can test the code without the if statement, but I believe without this test, the code would pass back empty for the first line in the file, which in the examples I have seen is the <?xml version="1.0" encoding="UTF-8"?>.

src/language/providers/utils.ts
```diff
@@ -17,6 +17,8 @@
 
 import * as vscode from 'vscode'
 
+const schemaPrefixRegEx = new RegExp('<(.*?)schema')
```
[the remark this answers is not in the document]
rt320: I'll make the change and test it

src/language/providers/utils.ts
```diff
@@ -283,6 +293,31 @@ export function checkSetVariableOpen(
...
+  var lineNum = 0
+  while (initialLineNum !== 0 && lineNum <= initialLineNum) {
+    const lineText = document
+      .lineAt(lineNum)
+      .text.substr(0, document.lineAt(lineNum).range.end.character)
+    // returns either empty prefix value or a prefix plus a colon
+    if (lineText.includes('schema')) {
+      let text = lineText.match(schemaPrefixRegEx)
+      if (text != null) {
+        console.log(text[1])
```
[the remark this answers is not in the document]
rt320: Removed

src/language/providers/utils.ts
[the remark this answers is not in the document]
stevedlawrence: That shouldn't match the regex and should return null and so not return. Running a small test, this outputs `null` for me:

```js
const regex = new RegExp('</?(|[^ ]+:)schema');
const str = '<?xml version="1.0" encoding="UTF-8"?>';
console.log(str.match(regex))
```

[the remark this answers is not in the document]
rt320: It is working for me now.

rt320:
Sorry, I missed your comment about forced commits.

src/language/providers/utils.ts
stevedlawrence: I don't think this comment was adressed? Everything else looks good to me.

src/language/providers/utils.ts
```diff
@@ -283,16 +272,35 @@ export function checkSetVariableOpen(
...
   position: vscode.Position
 ) {
   var lineNum = position.line
 
   while (lineNum !== 0) {
-    const wholeLine = document
-      .lineAt(lineNum)
-      .text.substring(0, document.lineAt(lineNum).range.end.character)
+    const wholeLine = document.lineAt(lineNum).text
+    //.text.substring(0, document.lineAt(lineNum).range.end.character)
```
stevedlawrence, line 303: Remove comment

stevedlawrence (approved):
+1 with one minor fix

Please squash into a single commit, force push, and this is good to merge.

When you squash, make sure to fix the comments so they make sense for what this squash commit actually adds. For example, one of the comments in your commit says: "moved schema regex to a global variable". That was a change made based on review comments, but doesn't actually happen in this PR since the schema regex is completely new in this PR.

Also, you have `closes issue #120`, but based on githubs documentation I think it might need to be `Closes #120`? I'm not positive though. Regardless, `Closes #XYZ` is the convention we've been using

stevedlawrence:
The commit message isn't quite right. These lines should be removed:
```
-renamed function getNsPrefix to getXsdNsPrefix
-added additional comments to getXsdNsPrefix to notify that a colon is include with the returned prefix
-replaced if statement to test for "schema" with a regex
```
The commit does not actually do these things. Those were changes made due to review comments, but in the context of this single squashed commit do not happen.

There's also duplicate sentences in the commit message.

And you only need to include the related issue at the end, with "Closed #122". It does not need to be on the first line.

Also, I'm not sure if the meaning of "Resolves Remove Prefixes" is very clear. Should that be something like "Suggest namespace prefixes based on the schema element prefix"

scholarsmate:
In the commit message we have this:

> add function to parse prefix from schema tag and return the prefix **plas** a colon or return nothing

plus is misspelled as **plas**.

stevedlawrence:
There are conflicts that are preventing this from being merged. Can you rebase onto the latest master and fix those conflicts?

Also, while your at it, there's a type in the subject of the commit, there should be a space in `Suggestnamespace`

stevedlawrence:
FYI, what you merged changed all the spacing of package.json, and one comment was not addressed. Please don't be hasty about merging PRs and make sure to double check the differences when you force push changes.

stevedlawrence:
Also, ignoring the spaces in package.json  this also added a `__metadata` item, which I think we determined in another PR should not be there?

PR #305, opened 2022-09-22
Add semantic highlighting for XPath expressions

### LINKED ISSUE #123: Add XPath Color Segmentation
Update language extension to add XPath Color Segmentation:
- Stems from https://github.com/apache/daffodil-vscode/pull/117#discussion_r883745861

michael-hoke:
Closes #123 

This is a draft PR because the other suggestion of including the DeltaXML as an extension dependency and importing its code as needed has not been fully investigated yet. This seems like it should be doable.

The Rat checks are going to fail because there was some discussion on how these files should be licensed, but, as far as I can remember, no final solution had been decided on. Ideally, if we use the dependency solution, this becomes moot because we won't need to provide a license for these files.

I removed what I could from the original set of files so that we're only including code related to XPath (and not XSLT). This is mostly meant as a first pass over the code to remove what is obviously not being used by XPath.


michael-hoke:
I think this is the comment that referenced including DeltaXML as a dependency rather than including the code.

https://github.com/apache/daffodil-vscode/pull/117#discussion_r886867014

I thought there was a more concrete suggestion than that, but I can't find it.

michael-hoke:
I don't think that using the DeltaXML extension is going to be possible. From what I understand, for that to happen, our extension would need to be able to reference the API returned from the DeltaXML extension's activate function. However, it has no return value, so there will be no API.

I will make another pass through the code in this PR to see if anything else can be removed, but a lot of it will come down to the differences between XPath and XSLT, which I am not too familiar with.

michael-hoke:
@stevedlawrence @mbeckerle

We've got some questions on the licensing that we hope you can help clear up. The DeltaXML extension, which is where most of the code comes from, uses MIT licensing, which requires us to use the MIT license in all copied files. However, we have a couple of scenarios:

- We've copied the file from DeltaXML, but we have removed portions of the code as the extension will never use it. Is the MIT license appropriate here?

- We've made a small modification to the copied DeltaXML file. I think out of the two modifications, one was reverting where I took too much code out, and the other was adding in new code. Do we need to combine the MIT and Apache licenses somehow?

- We have a file that has both significant portions of copied sections of a DeltaXML file and sections written by us. Is a combined license appropriate here?

If we do need a combined license, what would that look like?

@NolanMatt Is this an accurate characterization of the changes you've made, and/or am I missing anything?

NolanMatt:
> @stevedlawrence @mbeckerle
> 
> We've got some questions on the licensing that we hope you can help clear up. The DeltaXML extension, which is where most of the code comes from, uses MIT licensing, which requires us to use the MIT license in all copied files. However, we have a couple of scenarios:
> 
> * We've copied the file from DeltaXML, but we have removed portions of the code as the extension will never use it. Is the MIT license appropriate here?
> * We've made a small modification to the copied DeltaXML file. I think out of the two modifications, one was reverting where I took too much code out, and the other was adding in new code. Do we need to combine the MIT and Apache licenses somehow?
> * We have a file that has both significant portions of copied sections of a DeltaXML file and sections written by us. Is a combined license appropriate here?
> 
> If we do need a combined license, what would that look like?
> 
> @NolanMatt Is this an accurate characterization of the changes you've made, and/or am I missing anything?

This is spot on, not missing anything.

stevedlawrence:
The MIT license does allow modifications to be licensed under different terms, you really just need to include the copyright header and MIT license. So you could in theory license your changes using ALv2. However, it adds a extra complexity, since you need to make it clear which parts are MIT and which are ALv2. And it doesn't really gain much having some parts be ALv2. It's recommended, and fine with the ASF, to license minor modifications under the original license. 

In fact, this page has some information about this exact thing in "Treatment of Third-Party Works": https://www.apache.org/legal/src-headers.html, and says exactly that in bullet 4:

> 4. Minor modifications/additions to third-party source files should typically be licensed under the same terms as the rest of the third-party source for convenience.

If the changes are "major" then we probably need to have a discussion to determine if mixing licenses is appropriate, and "major" vs "minor" is pretty subjective, but based on your description it sounds like they are pretty clearly minor changes.

So I'd just keep everything under the same MIT license terms.

michael-hoke:
Thanks, that sounds like it covers us for almost everything. We do have one file that has what I would consider to be major changes - I'd estimate it on the order of 50/50 with code that we pulled from the extension versus code we wrote ourselves, and, if it changes anything, there's probably more code in the original file that we didn't use than code we did use.

Their file: https://github.com/DeltaXML/vscode-xslt-tokenizer/blob/master/src/extension.ts
Our file: https://github.com/michael-hoke/daffodil-vscode-fork/blob/daffodil-vscode-semantic-highlighting/src/language/semantics/dfdlExt.ts

That is also assuming that removing significant portions of code isn't going to be considered a major change.

scholarsmate:
@michael-hoke, can we close this in place of #705?

michael-hoke:
Yes - closing in favor of #705

PR #331, opened 2022-11-01
omega-edit updates

### LINKED ISSUE #322: Upgrade Ωedit version
There have been some improvements to Ωedit that can benefit the extension.

Shanedell:
omega-edit updates:

- Update version from 0.9.20 to 0.9.23
- Update viewport subscribe to work for new version
  - "data" function not called immediately anymore, so we need to make sure the data is when setting up subscribtions.

Closes #322


yarn.lock
```diff
@@ -2,6 +2,13 @@
 # yarn lockfile v1
 
 
+"@cspotcode/source-map-support@^0.8.0":
```
mbeckerle, line 5: This file should not be checked in at all. Please add it to .gitignore so it doesn't get committed by accident again.

mbeckerle (approved):
+1. Just remove yarn.lock from the change set and add it to .gitignore

yarn.lock
mbeckerle: Actually. I take it back. It may be that you want to keep the yarn.lock file, as it captures existing dependencies so one can detect changes to dependencies.  It gets regenerated, so keep in the tree if you see fit.

Shanedell: @mbeckerle Yes I believe this was the reason we added it, was to ensure from developer to developer you make sure the dependencies are the same.

scholarsmate (approved):

stevedlawrence (approved):
+1

Changes look reasonable. Though it looks like there are a number of new dependencies, I assume pulled in by the new version of omega edit? Have those been verified to be ASF compatible and updates made to the `build/package/{NOTICE,LICENSE}` files?

Shanedell:
> +1
> 
> Changes look reasonable. Though it looks like there are a number of new dependencies, I assume pulled in by the new version of omega edit? Have those been verified to be ASF compatible and updates made to the `build/package/{NOTICE,LICENSE}` files?

@stevedlawrence I do need to update that, will do it in this PR before merging it. I know some were using Apache licensing without a notice in there repo. Can these possibly still be listed inside of the `build/package/NOTICE` and just do something like this

```
- '@grpc/grpc-js' packaged in extension/dist/ext/extension.js
  NOTICE not provided. See repo at https://github.com/grpc/grpc-node/tree/master
```

so that we know which dependencies are not known to be a part of the LICENSE or NOTICE? Also making sure we don't keep rechecking dependencies that we know are approved but just aren't marked down somewhere?

stevedlawrence:
The recommendation we've gotten in the past during incubation is to not include extra information in NOTICE files, even if it's to keep track of files that have no NOTICE.

This does adds difficulty when trying to keep track of which are missing and which just don't have a notice. Maybe we can figure out an alternative approach instead of keeping that information in the notice file.

mbeckerle:
Maybe create a NONOTICE file to list these.

Shanedell:
> Maybe create a NONOTICE file to list these.

I would be okay with doing this @stevedlawrence do you think this would be okay to do? Put it in the same location as the release NOTICE and LICENSE at `build/package`, however not actually bundle it inside of the release?

stevedlawrence:
Fine with me.

Shanedell:
@stevedlawrence @mbeckerle So a I think there is a good chunk of the `yarn.lock` file dependencies that are possibly not in the LICENSE or NOTICE. Would you guys be okay with me merging this PR in and creating another ticket to make sure all dependencies inside of `yarn.lock` are in the LICENSE, NOTICE or NONOTICE (will be created as part of that PR) files before a possible 1.2.0 release?

stevedlawrence:
Fine with me, please create a new issue so we don't forget about it.

Shanedell:
Create issue #333 to handle the tracking of the node dependencies.

PR #341, opened 2022-11-06
omega-edit updates

Shanedell:
omega-edit updates:

- Use version 0.9.24.
- Use omega-edit replace function instead of using work around.
- omega-edit added a NOTICE so removed it from NONOTICE and added it to the release NOTICE file.


mbeckerle (approved):
+1

build/package/NOTICE
```diff
@@ -280,3 +280,18 @@ The following NOTICE information applies to binary components distributed with t
   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
   See the License for the specific language governing permissions and
   limitations under the License.
+
+- 'omega-edit' in extension/dist/ext/extension.js
+  Copyright 2021 Concurrent Technologies Corporation.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
```
stevedlawrence, line 287: This is correct since this is what's in omega-edit's NOTICE file, but in case omega-edit devs see this, the ASF recommends to not put the license in the NOTICE file (that's what the LICENSE is for). This file usually about copyright information, code origins, and third-party NOTICES.

stevedlawrence (approved):
+1

build/package/NOTICE
Shanedell: Tagging @scholarsmate so he sees this comment about the NOTICE file for omega-edit.

Shanedell: @stevedlawrence Is this good to merge and I can make an update to the NOTICE for `omega-edit` if its `NOTICE` file is updated?

stevedlawrence: Yep, this is good to merge for the 0.9.24 version of omega-edit. If omega-edit changes the NOTICE in a later release then we can update our NOTICE/NONOTICE files as needed.

PR #349, opened 2022-11-18
Add support for JSON formatting of infoset

### LINKED ISSUE #343: Support Daffodil different infoset options in the debugger
The Daffodil Scala debugger needs to be updated to support different input/output options for the infoset that are defined inside of Daffodil. These include SAX, JSON, etc.

@stevedlawrence @mbeckerle would you guys mind leaving a comment about all the different types of input and/or output supported by Daffodil for the infoset that you would like to be supported by the debugger?

Shanedell:
Add support for JSON formatting of infoset

- Create new argument infosetFormat
- Update scala code to support infosetFormat:
  - Parse in value from extension.
  - Based on parse value decide what InfosetFormatter to use.
- Update extension code to support infosetFormat:
  - Add it is a launch arg inside of the launch.json.
  - Set default value as "xml".
  - Add selection box for infosetFormat on the Launch Wizard.
  - Send value over the debugger on launch
  - Update infoset view to select file type based on infosetFormat so they look correct.
  - Update LaunchArgs interface to include infosetFormat.
  - Update unit tests to use value and check that it is set properly
- Update yarn watch to generate src/version.ts before running actual watch command.

Closes #343

NOTE: 1.2.0 was released but holding off on merging this until the update for the daffodil website is merged and I send the release announcement out.

### DIFF
--- a/src/infoset.ts
+++ b/src/infoset.ts
@@ -46,8 +46,8 @@ async function openInfosetFilePrompt() {
 
     switch (action) {
       case 'Open':
-        let xml = await vscode.workspace.openTextDocument(uri)
-        await vscode.window.showTextDocument(xml, {
+        let infoset = await vscode.workspace.openTextDocument(uri)
+        await vscode.window.showTextDocument(infoset, {
           preview: false,
           viewColumn: vscode.ViewColumn.One,
         })
@@ -173,7 +173,7 @@ const fileInfosetProvider = new (class
 })()
 
 function tmp(sid: string): string {
-  return `${tmpdir()}/infoset-${sid}.xml`
+  return `${tmpdir()}/infoset-${sid}.${getCurrentConfig().infosetFormat}`
 }
 
 function ensure(path: string): string {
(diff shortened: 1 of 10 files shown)


scholarsmate (approved):
+1, LGTM for v1.3.0.

src/infoset.ts
```diff
@@ -46,8 +46,8 @@ async function openInfosetFilePrompt() {
 
     switch (action) {
       case 'Open':
-        let xml = await vscode.workspace.openTextDocument(uri)
-        await vscode.window.showTextDocument(xml, {
+        let data = await vscode.workspace.openTextDocument(uri)
```
mbeckerle, line 49: Do you really want to call this value "data" ? How about "infoset". Normally we think of data as the input to parse, and infoset as the output from parse, and most of your code follows that naming.

mbeckerle (approved):
+1
I made one variable rename suggestion. Otherwise looks ok.

Shanedell:
1.2.0 was released just need to update the page a bit before sending out the announcement so I am going merge this now.

PR #371, opened 2022-12-05
omega-edit updates

### LINKED ISSUE #334: Move omega-edit zip and code to run it to omega-edit node package
The `omega-edit` yarn package should package the zipped up server and the code necessary to unzip, run and stop the server instead of doing this inside of this repo. This is line with an issue in omega-edit https://github.com/ctc-oss/omega-edit/issues/447

Shanedell:
omega-edit updates:

- Moved omega-edit server related code to the omega-edit node module.
- Moved omega-edit scala server zip to omega-edit node package.
  - Delete build/scripts/omega_edit_download since the scala server is now bundled in the node package.
  - Remove omegaEditServerHash from package.json.
- Display information messages when starting and stopping the omega-edit server.
  - This is to give the developer a notice that it is starting/started or stopped.
  - This replace the terminal being opened displaying the omega-edit server binding.

Closes #334


src/utils.ts
```diff
@@ -173,6 +173,19 @@ export async function executeScript(
   return terminal
 }
 
+export async function stopJavaProcess() {
```
mbeckerle, line 176: Is there no decent os-lib type library for typescript/javascript?

mbeckerle (approved):
+1

I'd like to find an OS library rather than bit-by-bit inventing our own, if possible.

src/utils.ts
Shanedell: No there is not, I am not to sure how common it is to run scripts on the OS through TS/JS like we are doing. But in another PR I have coming it up it split out a lot the OS specific stuff into separate functions to allow for better use of those items. But yeah I couldn't find any package that would be what we want, the closest thing is `run-script-os` but that is for the `package.json` scripts to make OS specific commands that way but nothing like what we would need.

Shanedell:
@scholarsmate @stevedlawrence This should be good for a look over now. I apologize for all the commits it took me forever to figure out why the CI kept failing for windows. The issue seemed to be the path to the batch script was too long making it not ever run the script.

Shanedell:
@stevedlawrence @scholarsmate I created a new function inside of `omega-edit` for `setupServer` that will do everything except run the server. So I could technically still update the code to run the `omega-edit` server inside of the vscode terminal so the user can actually see it running as opposed to it running in the background. Do you think it would be preferred for it be seen or hidden?

Shanedell:
okay I will make that fix now

src/omega_edit/client.ts
```diff
@@ -72,8 +83,14 @@ export function activate(ctx: vscode.ExtensionContext) {
...
         if (!serverRunning && startServer) {
-          await getServer(ctx, omegaEditPackageVersion)
+          vscode.window.showInformationMessage('Starting omega-edit server!')
+          const [scriptName, scriptPath] = await omegaEditServer.setupServer(
+            rootPath,
+            omegaEditPackageVersion
+          )
+          serverTerminal = await runScript(scriptPath, scriptName)
           serverRunning = true
+          vscode.window.showInformationMessage('omega-edit server started!')
```
scholarsmate, line 93: This code is duplicated in lines 62 - 69.  Consider a refactor?

src/tests/suite/omegaEdit.test.ts
```diff
@@ -23,30 +23,30 @@ import { Artifact, Backend } from '../../classes/artifact'
 import * as omegaEditClient from '../../omega_edit/client'
 import { unzipFile, runScript, killProcess } from '../../utils'
 import { before, after } from 'mocha'
+import * as fs from 'fs'
 
+const PROJECT_ROOT = path.join(__dirname, '../../../')
```
scholarsmate, line 28: Instead of having this relative link, resolve it via `path.resolve()`, which should also shorten the path names.

src/utils.ts
```diff
@@ -173,6 +173,19 @@ export async function executeScript(
...
+  if (os.platform() === 'win32') {
+    // Windows stop debugger if already running
+    child_process.execSync('tskill java 2>nul 1>nul || echo "Java not running"')
+  } else {
+    // Linux/Mac stop debugger if already running and make sure script is executable
+    child_process.exec(
+      `kill -9 $(ps -ef | grep '${processName}' | grep 'jar' | awk '{ print $2 }') || return 0`
+    ) // ensure debugger server not running and
+  }
+}
```
scholarsmate, line 187: I'm not a fan of this technique to kill processes.  When we exec the server, we can store the process ID and send the kill signal to that PID.  I could be running some other java process or even some other Ωedit process on a different port and this code will just murder those processes as collateral damage.

Shanedell: Yeah this was mostly used for the daffodil-debugger to make sure another instance of it is running, but you are correct with the addition of omega-edit this could possibly trample that omega-edit instance if running. I would be also for removing this. @mbeckerle @stevedlawrence Do you think this would be good to actually remove?

src/daffodilDebugger.ts
```diff
@@ -120,18 +120,8 @@ export async function getDebugger(
...
-          'tskill java 2>nul 1>nul || echo "Java not running"'
-        )
-      } else {
-        // Linux/Mac stop debugger if already running and make sure script is executable
-        child_process.exec(
-          "kill -9 $(ps -ef | grep 'daffodil' | grep 'jar' | awk '{ print $2 }') || return 0"
-        ) // ensure debugger server not running and
-      }
+      // stop java program -- in this case the debugger
+      await stopJavaProcess('daffodil')
```
scholarsmate, line 124: Refactoring this as a utility function is a good idea.  Instead of sending in the process _name_, instead send it the process ID to be more targeted and safe about the process killing.

scholarsmate (changes requested):
Good refactoring effort.  Consider refactoring the code to reduce duplication (DRY principle).

Be more targeted about process murdering.  -1 right now for this reason.

src/utils.ts
scholarsmate: If the extension is _managing_ these server processes, what it starts up, it shall also shut down.

If there is _already_ a daffodil-debugger service running _on the same port_ as configured in VS Code, then consider it _unmanaged_ and use it as-is, and don't shut it down when the debugging session is done.  The same applies to the Ωedit server.  This way the extension _cooperates_ with different instances of the servers if they are running and listening to the configured port, otherwise it will use our internal versions as managed processes.

scholarsmate: Take a look at [spawnd](https://www.npmjs.com/package/spawnd).

Shanedell: we want to limit the amount of deps we add. This is done already whenever we start up our own instances however the above code was for ensure a daffodil-debugger was not running without knowing the pid. This was only ran in one place so removing it shouldn't have much of an effect

src/tests/suite/omegaEdit.test.ts
Shanedell: @scholarsmate Was able to remove most of these. However, it needed to be kept in the `src/tests/runTest.ts`

Shanedell: In another PR I have coming there will be a split out of some of these common items between test files as well

Shanedell: actually this change causes windows tests to fail so we will need to keep this how it is

src/tests/suite/adapter.test.ts
```diff
@@ -20,8 +20,11 @@ import * as path from 'path'
 import { DebugClient } from 'vscode-debugadapter-testsupport'
 
 suite('Daffodil Debug Adapter', () => {
-  const DEBUG_ADAPTER = path.resolve('out/adapter/debugAdapter.js')
-  const PROJECT_ROOT = path.resolve('.')
+  const DEBUG_ADAPTER = path.join(
+    __dirname,
+    '../../../out/adapter/debugAdapter.js'
+  )
+  const PROJECT_ROOT = path.join(__dirname, '../../../')
```
scholarsmate, line 27: I think we should still try to resolve these paths instead of using these relative paths, but I won't hold up this PR over it.  It's probably merits a "technical debt" ticket though.

scholarsmate (approved):
+1

src/tests/suite/adapter.test.ts
Shanedell: It worked okay on mac but seemed to cause issues in windows CI. So some more testing will need to be done on the other OSes, I believe one of the ubuntu CI's ran it fine so maybe just a thing to test out further on windows?

PR #378, opened 2022-12-08
1.3.0 snapshot

scholarsmate:
This is to prepare the repo for building 1.3.0 snapshots.


scholarsmate:
This is not what I intended.

PR #388, opened 2022-12-19
Run daffodil-debugger in CI.

### LINKED ISSUE #301: Suppress platform conditionals
Definitely make it a goal to use a library that suppresses all these platform conditionalizations before they proliferate. 

As people add new tests they should NOT copy this. 

We didn't do that soon enough in the daffodil-cli test rig, and the code-volume legacy of these sort of trivial conditionalization literring that test code is way too high.

_Originally posted by @mbeckerle in https://github.com/apache/daffodil-vscode/pull/240#discussion_r975803640_

### LINKED ISSUE #337: Java 8 compatibility
Attempt to revise our minimum Java dependency from 11 back down to 8 to accommodate other Apache projects like Apache Drill that are still on Java 8.

### LINKED ISSUE #377: Update test suite to support daffodil-debugger
Currently the test suite does not actually run the `daffodil-debugger`. We want to update the test suite to run the `daffodil-debugger` and then run tests against it, similar to how we do the `omega-edit` tests. Based on prelim looks this issue will most likely work in tandem with #301 as a lot of the code for running the `daffodil-debugger` will need to shared between the actual extension code and the suite.

Shanedell:
Run daffodil-debugger in CI.

- Update tests to run the daffodil-debugger in the CI, then run tests against it.
  - Created two simple tests to ensure the infoset files are created.
- Code needed updated to split out a lot of common code:
  - Create a new daffodilDebuggerUtils file that holds functions that are common to get the debugger running, stopped, etc.
  - Create a function for doing os checks to remove a lot of duplicated code and limit number of types it shows up in the code.
- Update the test schema to be a proper schema so it could be parsed by the debugger.
- Create a src/tests/common.ts for common variables that a few different tests use. Later on this include more vars and functions if needed.
- Support Java 8
  - Ignore com.microsoft.java for scala-steward as the version used is the last one that supports java 8

Closes #301
Closes #337
Closes #377

Notes:

- For some reason in the CI it seems that sometimes the two new TS tests for making sure an infoset is created will decide to fail. This seems to happen only on macos though.
- Currently for both tests to run we need to run a separate instance of the daffodil-debugger before hand. I think we should add an option to the daffodil-debugger to allow multiple connections and not automatically close after one connection is done debugging/running. This would make it so in the future if we have say ten tests that need the debugger, we can just run the debugger once and run all the tests as opposed to having to run the debugger before each test (in this scenario we would run it 10 times). I think this may be what causes the issue in the macos CI to sometimes fail, as I think sometimes the 1st instance of the debugger could have not been killed before the 2nd instance is trying to run. How does everybody else feel about this?
  - UPDATE: I updated the test to run two instances of the debugger, on different ports, at the start of the tests which I think lets the run a bit better/quicker. Also seemed to fix the issue I was seeing on macos
  - Also know normally we want 1 PR for an Issue but the 3 being closed were heavily needed in this PR for everything to get done properly


build.sbt
```diff
@@ -78,7 +78,9 @@ lazy val core = project
     name := "daffodil-debugger",
     libraryDependencies ++= Seq(
       "ch.qos.logback" % "logback-classic" % "1.2.11",
-      "com.microsoft.java" % "com.microsoft.java.debug.core" % "0.38.0",
+      // scala-steward:off
```
mbeckerle, line 81: Can you also add a comment about why (java 8) this scala-steward off is here.

src/tests/data/test.dfdl.xsd
```diff
@@ -16,15 +16,48 @@
...
+						</xs:appinfo>
+					</xs:annotation>
+				</xs:sequence>
+				<xs:sequence dfdl:hiddenGroupRef="hiddenGroup" />
+			</xs:sequence>
+		</xs:complexType>
+	</xs:element>
+
+  <xs:group name="hiddenGroup">
+		<xs:sequence>
```
mbeckerle, line 59: Add comment that this is intentionally an empty sequence (i..e., you didn't leave something out by accident.)

mbeckerle (approved):
+1 Looks like a pretty good cleanup as well as getting CI working, which is great.

scholarsmate (approved):

PR #402, opened 2023-01-06
Launch Wizard - Daffodil Debugger Classpath manipulation

### LINKED ISSUE #268: Ability to manipulate Daffodil Debugger Classpath is not adequate
The the ability to manipulate Daffodil Debugger Classpath is not adequate. It needs to display the list of entries on separate lines and they're long. I would have expected a "+" button, which when pressed creates an additional box to type or browse another entry into. With the browse feature this would be really pretty good.

Shanedell:
Update the launch config wizard manipulation of Daffodil Debugger Classpath:

- Remove replace and append check boxes.
- Replace single box that listed the classpath items with a UI list.
  - Browse will work the same but when you select a folder or file it will be added to the list.
  - List items will have a minus button beside the text. Both button and text are clickable and when clicked it will remove that specific item from the list.
- Save now will now write the daffodil debugger classpath to the launch.json.
- Update UI list to auto-populate the classpath items if set in the configuration.

Closes #268

### DIFF
--- a/src/launchWizard/launchWizard.ts
+++ b/src/launchWizard/launchWizard.ts
@@ -152,38 +152,7 @@ async function openFilePicker(description) {
       title: description,
     })
     .then(async (fileUri) => {
-      if (fileUri && fileUri[0]) {
-        let stats = fs.statSync(fileUri[0].fsPath)
-
-        if (stats.isDirectory()) {
-          let basePath = fileUri[0].fsPath
-          let fileString = ''
-
-          fs.readdirSync(basePath).forEach((file) => {
-            if (file.includes('.jar')) {
-              fileString += `${basePath}/${file},`
-            }
-          })
-
-          return fileString.substring(0, fileString.length - 1) // make sure to remove comma at the end of the string
-        }
-
-        if (fileUri.length === 1) {
-          return fileUri[0].fsPath
-        } else {
-          let files = ''
-
-          fileUri.forEach((file) => {
-            if (file.fsPath.includes('.jar')) {
-              files += `${file.fsPath},`
-            }
-          })
-
-          return files.substring(0, files.length - 1) // make sure to remove comma at the end of the string
-        }
-      }
-
-      return ''
+      return fileUri && fileUri[0] ? fileUri[0].fsPath : ''
     })
 
   if (chosenFile.includes(rootPath)) {
@@ -325,13 +294,22 @@ class LaunchWizard {
     let stopOnEntry = defaultValues.stopOnEntry ? 'checked' : ''
     let trace = defaultValues.trace ? 'checked' : ''
     let useExistingServer = defaultValues.useExistingServer ? 'checked' : ''
-    let daffodilDebugClasspathAction =
-      defaultValues.daffodilDebugClasspathAction === 'append'
-        ? defaultConf.daffodilDebugClasspath
-        : 'replace'
-    let replaceCheck =
-      daffodilDebugClasspathAction !== 'append' ? 'checked' : ''
-    let appendCheck = daffodilDebugClasspathAction === 'append' ? 'checked' : ''
+
+    let daffodilDebugClasspathList =
+      '<ul id="daffodilDebugClasspathTable" style="list-style: none; padding-left: 20px;">'
+    if (defaultValues.daffodilDebugClasspath) {
+      let itemArray = defaultValues.daffodilDebugClasspath.split(':')
+      for (var i = 0; i < itemArray.length; i++) {
+        daffodilDebugClasspathList += `
+          <li style="margin-left: -5px;" onclick="removeDebugClasspathItem(this)">
+            <p id="debug-classpath-li-${itemArray[i]}" class="debug-classpath-item">
+              <button id="remove-debug-classpath-li-${itemArray[i]}" class="minus-button" type="button">-</button>
+              ${itemArray[i]}
+            </p>
+          </li>`
+      }
+    }
+    daffodilDebugClasspathList += '</ul>'
 
     let infosetFormatSelect = ''
     let infosetFormatTypes = ['xml', 'json']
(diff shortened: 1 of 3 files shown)


src/launchWizard/launchWizard.ts
```diff
@@ -325,13 +294,22 @@ class LaunchWizard {
...
+    if (defaultValues.daffodilDebugClasspath) {
+      let itemArray = defaultValues.daffodilDebugClasspath.split(':')
+      for (var i = 0; i < itemArray.length; i++) {
+        daffodilDebugClasspathList += `
+          <li style="margin-left: -5px;" onclick="removeDebugClasspathItem(this)">
+            <p id="debug-classpath-li-${itemArray[i]}" class="debug-classpath-item">
+              <button id="remove-debug-classpath-li-${itemArray[i]}" class="minus-button" type="button">-</button>
+              ${itemArray[i]}
+            </p>
+          </li>`
```
scholarsmate, line 309: This feels much like the code in the `updateClasspathList` function.  Is there any way to consolidate it?

Shanedell: Not really. This is done for the initial setup of the list in the extension code if there are classpath items that need appended at the start, where as `updateClasspathList` requires using JS to get the list and append items to it. This can not be done inside of the extensions TS code as it does have have access to the DOM. The JS code could possibly on startup/load initialize the list, however doing it this way could cause issues rendering, such as the UI loading once then immediately getting refreshed causing the render to not seem smooth or possibly breaking. It would also cause the need to embed the classpath items into a hidden html element. There is some wiring that can be done between the JS and TS that could possibly do what you are saying, however even if it could in this cause it wouldn't be effective and would also probably cause more issues/confusion than it would fix.

scholarsmate: The Data Editor feature will be doing just as you described, sending data from the extension to the webview for rendering.  We'll see how it goes.  Either way we'll get smarter on this pattern.

Shanedell: Yeah so as I said this is not to bad to do but in this situation I don't think it makes to much sense, as I don't think it provides any benefits over the way it is currently being done. I believe changing it to try to have JS add items on load, can cause issues during the rendering making the launch config wizard not open smoothly or at all.

Shanedell:
@mbeckerle Would you mind taking a look at this PR? I believe you were the one who reported this was an issue in your 1.2.0 release vote. So it would be good to know if you like the updated way a user will manipulate the daffodil debugger classpath.

src/launchWizard/launchWizard.ts
```diff
@@ -152,38 +152,7 @@ async function openFilePicker(description) {
       title: description,
     })
     .then(async (fileUri) => {
-      if (fileUri && fileUri[0]) {
```
mbeckerle, line 155: Github won't let me comment on earlier lines. 

Line 139 - it's fine for the picker to start with the workspace folder, but it needs to be able to point at any folder on the machine. E.g., I've setup classpaths to refer to jars under ~/.ivy2/local/.... which is where they get put for a dependent schema by 'sbt publishLocal'. 

So navigating up and out of the workspace is needed.

mbeckerle (approved):
+1 Looks like the right kind of thing to me.

Just one note about the file/folder selection/picker.

src/launchWizard/launchWizard.ts
Shanedell: So you should be able to navigate out of the initial folder to any folder you want, I believe I have moved out of the initial folder many times. Are you not able to do that currently?

mbeckerle: I have not tried it in a while.

Shanedell: Oh okay it might worth while to try it out on this PR just to ensure if this is still an issue for you.

mbeckerle: To do that I need to up my game on daffodil-vscode.

I need to know how to build and deploy it locally based on a git sandbox, and how to tell it to use the local sandbox daffodil snapshot instead of using a released one. 

Presumably you guys who are actively developing on this do this all the time, but I have yet to do this. 

Instructions are where?

Shanedell: So some instructions for debugging it locally can be found here [Development Wiki Page](https://github.com/apache/daffodil-vscode/wiki/Apache-Daffodil%E2%84%A2-Extension-for-Visual-Studio-Code:-Development#apache-daffodil-extension-for-visual-studio-code-development). 

However, there is a decently quick way to get up running. After you meet the pre-reqs listed on the link above and get this PR pulled down, run these commands at the root of repo (this is described in the README.md under section [Package VSIX and Debugger](https://github.com/apache/daffodil-vscode#package-vsix-and-debugger) and [Installation From File](https://github.com/apache/daffodil-vscode#installation-from-file)):

```bash
rm -rf node_modules out || true
yarn install
yarn package
code --install-extension apache-daffodil-vscode-1.3.0-SNAPSHOT.vsix
```

So what this does is:

- Remove any hanging installed node package or compiled code
- Install the latest version of the node packages
- Compile the extension and debugger code, into an extension file (`.vsix`)
- Then installs the just built extension file into your VS Code editorer.

Now you should have the extension installed so you can open up VS Code, open the launch config wizard and test if you can navigate out of the initial folder (it should use your native os file explorer window). 

However, if this is also to much I could upload a VSIX to this PR. This would allow to download and install the vsix file made from this PRs code, then test out the navigation.

Shanedell: @mbeckerle Am I good to merge this or do you have updates on testing the browsing?

mbeckerle: I have setting this up on my todo list. I have not had any time to do this.

Shanedell: Oh okay no problem, do you want me to keep this open till you get a chance to do that?

mbeckerle: To be clear. You are good to merge.

Shanedell: Alright thank you!

PR #406, opened 2023-01-09
allow for self-signed certs when dowloading test artifacts over https

### LINKED ISSUE #405: Testing issues behind endpoint protection
When fetching test artifacts over SSL using endpoint protection, we see this error:

```
Error: self-signed certificate in certificate chain
```

Add an option to the `runTest.ts` script for allowing self-signed certs and/or a way to trust certain certificates so that we can run the test cases.

scholarsmate:
fixes #405


stevedlawrence:
What artifacts are being downloaded during tests?

PR #407, opened 2023-01-09
allow self-signed certs when downloading test files over https (#405)

### LINKED ISSUE #405: Testing issues behind endpoint protection

### LINKED ISSUE #413: Nightly CI tests against VS Code 'stable' and 'insiders'
CI is running against the earliest supported version of VS Code (currently 1.60.2).  Setup nightly CI by setting the `DAFFODIL_TEST_VSCODE_VERSION` environment variable to `stable` or `insiders` to run against the latest stable release or the latest (nightly) insiders release of VS Code, respectively.

scholarsmate:
Fixes #405
Closes #413

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -86,7 +86,7 @@ jobs:
         run: $SBT ratCheck || (cat target/rat.txt; exit 1)
 
   build-test-package:
-    name: Build, Test, and Package
+    name: 'Build, Test, and Package (OS: ${{ matrix.os }}, Node: ${{ matrix.node }})'
     strategy:
       matrix:
         os: [macos-11, ubuntu-20.04, windows-2019]
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
 	"author": "Apache Daffodil",
 	"license": "Apache-2.0",
 	"engines": {
-		"vscode": "^1.55.0"
+		"vscode": "^1.60.2"
 	},
 	"icon": "images/daffodil.ico",
 	"categories": [
@@ -55,7 +55,7 @@
 		"@types/glob": "^7.1.3",
 		"@types/mocha": "^9.1.1",
 		"@types/node": "^18.0.0",
-		"@types/vscode": "^1.55.0",
+		"@types/vscode": "^1.60.2",
 		"@vscode/test-electron": "^2.1.5",
 		"glob": "8.0.3",
 		"mocha": "10.0.0",
--- a/src/tests/runTest.ts
+++ b/src/tests/runTest.ts
@@ -18,13 +18,41 @@
 // All tests ran here are ones that require the vscode API
 import * as path from 'path'
 import * as cp from 'child_process'
+import { parseArgs } from 'util'
 import {
   runTests,
   resolveCliArgsFromVSCodeExecutablePath,
   downloadAndUnzipVSCode,
 } from '@vscode/test-electron'
 
+// The earliest version supported as indicated in package.json engines/vscode,
+// so that version is a reasonable default.
+const DEFAULT_DAFFODIL_TEST_VSCODE_VERSION =
+  require('../../package.json').engines.vscode.replace('^', '')
+
 async function main() {
+  const {
+    values: { disable_cert_verification },
+  } = parseArgs({
+    options: {
+      disable_cert_verification: {
+        type: 'boolean',
+        short: 'k',
+        default: false,
+      },
+    },
+  })
+  if (disable_cert_verification) {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
+  }
+
+  // The version of VS Code to use for running the test suite in.  Aside from
+  // actual version strings, 'stable' can be used for using the latest stable
+  // release and 'insiders' for using the latest insiders release.
+  const testVsCodeVersion = process.env.DAFFODIL_TEST_VSCODE_VERSION
+    ? process.env.DAFFODIL_TEST_VSCODE_VERSION
+    : DEFAULT_DAFFODIL_TEST_VSCODE_VERSION
+
   try {
     // The folder containing the Extension Manifest package.json
     // Passed to `--extensionDevelopmentPath`
@@ -34,7 +62,7 @@ async function main() {
     // Passed to --extensionTestsPath
     const extensionTestsPath = path.resolve(__dirname, './suite/index')
 
-    const vscodeExecutablePath = await downloadAndUnzipVSCode('1.68.0')
+    const vscodeExecutablePath = await downloadAndUnzipVSCode(testVsCodeVersion)
     const [cli, ...args] =
       resolveCliArgsFromVSCodeExecutablePath(vscodeExecutablePath)
 
@@ -48,16 +76,16 @@ async function main() {
       }
     )
 
-    // Download VS Code, unzip it and run the integration test
+    // Download VS Code, unzip it and run the integration tests
     await runTests({
       vscodeExecutablePath,
       extensionDevelopmentPath,
       extensionTestsPath,
     })
   } catch (err) {
-    console.error('Failed to run tests')
+    console.error('Failed to run tests: ' + err)
     process.exit(1)
   }
 }
 
-main()
+main().then()
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,57 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: Nightly Tests
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  build-test-package:
+    name: 'Build, Test, and Package (OS: ${{ matrix.os }}, Node: ${{ matrix.node }}, VS Code: ${{ matrix.vscode }} )'
+    strategy:
+      matrix:
+        os: [ macos-11, ubuntu-20.04, windows-2019 ]
+        node: [ '14', '16' ]
+        vscode: [ 'stable', 'insiders' ]
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+    env:
+      DAFFODIL_TEST_VSCODE_VERSION: ${{ matrix.vscode }}
+      SBT: sbt -J-Xms1024m -J-Xmx5120m -J-XX:ReservedCodeCacheSize=512m -J-XX:MaxMetaspaceSize=1024m ++2.12.15
+    steps:
+      - name: Setup Java 8
+        uses: actions/setup-java@v3.4.1
+        with:
+          distribution: temurin
+          java-version: 8
+      - uses: actions/checkout@v3.0.2
+      - name: Install Node.js
+        uses: actions/setup-node@v3.3.0
+        with:
+          node-version: ${{ matrix.node }}
+      - run: $SBT compile # used to initialize sbt, if not done sbt universal:packageBin ran by yarn package times out -- windows only
+        if: runner.os == 'Windows'
+      - run: yarn install
+      - run: yarn package
+      - run: xvfb-run -a yarn test
+        if: runner.os == 'Linux'
+      - run: yarn test
+        if: runner.os != 'Linux'
+      - run: $SBT test
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,19 +4,19 @@
 
 "@cspotcode/source-map-support@^0.8.0":
   version "0.8.1"
-  resolved "https://registry.yarnpkg.com/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz#00629c35a688e05a88b1cda684fb9d5e73f000a1"
+  resolved "https://registry.npmjs.org/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz"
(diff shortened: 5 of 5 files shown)


stevedlawrence:
What test artifacts are being downloaded?

scholarsmate:
@stevedlawrence VS Code.

src/tests/runTest.ts
```diff
@@ -34,7 +49,7 @@ async function main() {
     // Passed to --extensionTestsPath
     const extensionTestsPath = path.resolve(__dirname, './suite/index')
 
-    const vscodeExecutablePath = await downloadAndUnzipVSCode('1.68.0')
+    const vscodeExecutablePath = await downloadAndUnzipVSCode('stable')
```
scholarsmate, line 52: This is where we'd get a self-signed certificate in the certificate chain error when on the company VPN.

scholarsmate: Good question Steve.

For testing the extension, I think we ought to be floating on the latest stable, which this patch facilitates by using `'stable'` as the version instead of 1.68.0, which is pretty old now.  With 1.68.0, I was getting the following errors, which are fixed when using `'stable'`:

```
Extension is not compatible with Code 1.68.0. Extension requires: ^1.74.0.
Extension is not compatible with Code 1.68.0. Extension requires: ^1.74.2.
Extension is not compatible with Code 1.68.0. Extension requires: ^1.74.0.
```

Now pinning the version does allow us to _cache_ a local copy and not have to issue an HTTPS request to discover the current stable version, which was a problem for us before.  Since we were pinning to 1.68.0, we could hop off the VPN, download the artifact, then sign back into the VPN and continue testing without any HTTPS traffic.

stevedlawrence: Pinning feels perferable to me:

1. Disabling https doesn't feel great to me if it can be easily avoided.
2. Pinning to a particular version does help us to better define a minimum support version, or at least a known good version.
3. If a VS update breaks something, it won't immediately break all of our tests and gives time to figure out what changes are needed to succesfully update. Then we can pin to a certain version

Also, if this version came from an environment variable or something we could potentially have CI run with different VS Code versions if that was something that was of value.

scholarsmate: You make some valid points and appreciate this discussion.

You always want to pin your dependencies so you know precisely what is in your kit. This is a little different though as the extension is expected to run inside the users application where we have little control (we can have a "known good") over the version they have installed.  I imagine most users are running close to the latest stable with VS Code checking for updates and just the ease in which the update system works.  If the latest stable is breaking us, I think we ought to know as soon as practical because it _will_ affect our users and pinning it down to some old version will not allow us to "get out in front" of a potential issue, making us appear to be "sleeping behind the wheel".  That said, I like the idea of externalizing the version choice so we can test various versions as needed.  Perhaps a hybrid approach were we use a pinned "known good" version and also test the latest stable.  What factors shall we be considering when pinning down a VS Code version for testing?  Perhaps at the beginning of a release cycle we select the most recent stable and document that as the "known good" while also testing on the latest stable?

src/tests/runTest.ts
```diff
@@ -18,13 +18,34 @@
...
+    },
+  })
+  if (disable_cert_verification) {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
+  }
+
+  // The version of VS Code to use for running the test suite in
+  const testVsCodeVersion = process.env.TEST_VSCODE_VERSION
+    ? process.env.TEST_VSCODE_VERSION
+    : 'stable'
```
scholarsmate, line 47: @stevedlawrence, allow the VS Code version where we run the test suite to be externally configurable, with the latest stable release as the default.

src/tests/runTest.ts
stevedlawrence: > If the latest stable is breaking us, I think we ought to know as soon as practical

I agree, but defaulting to stable has some issues. For example, say VS Code makes an update to stable that breaks our extension. We can fix it then, but now if we ever try to run tests in a older commit that uses stable (e.g. to git bisect a bug) all those old tests will fail. If someone wants to run old tests, they should run exactly the same as it did back when those tests were committed. Version pinning is the right way to do this.

>  Perhaps a hybrid approach were we use a pinned "known good" version and also test the latest stable. 

Yeah, I think is is a good idea. I don't know enough about VS code to know what version to pin to or how often new releases are made. I think your suggestion is reasonable, but I'll leave it to the VS Code experts to decie.

I would suggest we create a GitHub action that runs our tests nightly and runs a test with stable (or maybe event nightly releases?) using your new environment variable and sends out some notification if it fails. This way tests are immutable and don't change depending on what stable happens to be at the time of running, but we also stay on top of changes to VS code and make sure our extension continues to work with new releases.

src/tests/runTest.ts
stevedlawrence: :+1: I like it. Though, I would just suggest we default to a pinned version and not stable, and some nightly CI can change this environment variable to stable or nightly to ensure we work with latest releases. So tests are still pinned but we still have a way to ensure we work with the latest and greatest.

src/tests/runTest.ts
```diff
@@ -18,13 +18,34 @@
...
+        default: false,
+      },
+    },
+  })
+  if (disable_cert_verification) {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
+  }
+
+  // The version of VS Code to use for running the test suite in
+  const testVsCodeVersion = process.env.TEST_VSCODE_VERSION
```
stevedlawrence, line 45: Should this be `DAFFODIL_TEST_VSCODE_VERSION` or something that makes it more clear this is Daffodil specific?

package.json
```diff
@@ -35,7 +35,7 @@
 		"package-setup": "node -e \"require('./build/scripts/package.ts').setup()\"",
 		"package-create": "node -e \"require('./build/scripts/package.ts').create()\"",
 		"pretest": "yarn compile && yarn webpack",
-		"test": "node ./out/tests/runTest.js",
+		"test": "node ./out/tests/runTest.js --disable_cert_verification",
```
stevedlawrence, line 38: As I rethink this, I'm a bit more concerned about disabling certificate validation. If a cert validation fails, it is a potential sign that you are under a MITM attack (which is sort of what's going on with your VPN, you just trust your man in the middle). I think it's fine to add this option, but I'm not sure it should be the default. Maybe just document this option somewhere?

src/tests/runTest.ts
scholarsmate: I think Daffodil is implied in the context, but I'm willing to rename this per your recommendation since it can be exported as a global environment variable (which I would _discourage_ using that way, but instead do it like `TEST_VSCODE_VERSION=1.68.0 node ./out/tests/runTest.js`).

package.json
scholarsmate: I pondered on this decision quite a bit, but finally decided to allow it as surgically as I can here.  My reason for allowing it is because it affects most of my development team, which are the main consumers of the test at this point.

src/tests/runTest.ts
stevedlawrence: Yeah, using is a global env is my concern. I could imagine a developer setting it globally to "stable" or "nightly" or something, and months later looking at their environment variables and not remembering that TEST_VSCODE_VERSION is actually configuring the Daffodil extension tests, but instead thinking it's something that VS Code itself uses or something.

package.json
stevedlawrence: Could this be made an environment variable too? Although most current VS Code extension devs run in to this, it feels dangerous to disable security by default. Someone should have to make an active choice to do that.

Also, as you mentioned in another comment, maybe version pinning makes this less of an issue since versions are cached and the pinned version won't be updated nearly as often as a stable would?

src/tests/runTest.ts
scholarsmate: I made this change.

package.json
scholarsmate: Funny you mention this, but it already _is_ an environment variable.  See lines 40-42 in runTest.ts.  Setting an environment variable is the mechanism used to communicate to node not to do cert verifications.  I'll change this back to doing cert verifications, then show my team how to do a surgical work-around because I don't want them setting `NODE_TLS_REJECT_UNAUTHORIZED=0` globally (e.g., via .bashrc or something) as that will affect _all_ node HTTPS connections.

scholarsmate: Certs are now being checked again by default.  If you're having problems with cert verifications, due to endpoint protection, there are ways to have node trust your endpoint protection certificate, or you can disable the cert verifications by either running the test script via `node ./out/tests/runTest.js --disable_cert_verification`, or via `NODE_TLS_REJECT_UNAUTHORIZED=0 npm run test`.  **Do not export NODE_TLS_REJECT_UNAUTHORIZED=0 into your environment as it will affect all node HTTPS connections done in that shell session.**

src/tests/runTest.ts
scholarsmate: Pinning done via CI.yaml using the `DAFFODIL_TEST_VSCODE_VERSION` environment variable.  Docs added for testing against a specific version of VS Code: https://github.com/apache/daffodil-vscode/wiki/Apache-Daffodil™-Extension-for-Visual-Studio-Code:-Development#testing-against-a-specific-version-of-vs-code

.github/workflows/CI.yml
```diff
@@ -96,6 +96,7 @@ jobs:
       run:
         shell: bash
     env:
+      DAFFODIL_TEST_VSCODE_VERSION: 1.74.2
```
stevedlawrence, line 99: This only pins in the GitHub CI, seems we should pin in runTests.ts and CI just tests with that version. Just like we don't want CI to randomly fail on stable updates, we don't want normal dev tests to fail on stable updates either.

src/tests/runTest.ts
```diff
@@ -18,13 +18,34 @@
...
+    },
+  })
+  if (disable_cert_verification) {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
+  }
+
+  // The version of VS Code to use for running the test suite in
+  const testVsCodeVersion = process.env.DAFFODIL_TEST_VSCODE_VERSION
+    ? process.env.DAFFODIL_TEST_VSCODE_VERSION
+    : 'stable'
```
stevedlawrence, line 47: I noticed package.json defines vscode engine at 1.55, should we be testing with that same version unless overridden by the env variable?

.github/workflows/CI.yml
scholarsmate: Why exactly don't we want normal dev tests to fail on stable updates?  I see this as something we should welcome and not avoid.  IMHO, we ought to be actively testing against the latest stable release.  CI can pin it against some "known good" so we don't regress against a particular historical version, which I believe is sufficient.  This project in general has not been great at keeping up to date with dependency versions and pinning the version inside the code will lead to yet another one of these versions that don't get updated with the times.  The VS Code version is under the end user's control, which is likely to be something more recent than what we'd be testing against.

src/tests/runTest.ts
scholarsmate: That said, I've never run the test suite against version 1.55.0, and if we're claiming that 1.55.0 is the minimum version we support, then perhaps we pin CI testing there.

.github/workflows/CI.yml
stevedlawrence: I should be able to test commit XYZ now and in some time in the future and should be guaranteed to should pass. Likewise, I should be able to download an old release and be guaranteed its tests pass no matter when I choose to run it and what stable happens to be at that time.

If we don't version pin and stable makes a breaking change, then that old commit or release that used to pass might fail and there's nothing you can do about it since we don't know what stable was at that point in time.

> IMHO, we ought to be actively testing against the latest stable release

100% agreed, and I would even claim we should be testing against beta/unstable/whatever if VS code has those so we catch any issues before they hit stable. But we should be doing so with automated builds/tests, and automated PRs that update the version to latest depenencies. We shouldn't need to rely on devs to run tests periodically enough to catch issues when a computer can do it on a much more frequent and reliable basis.

And If a dev wants to always test with the latest and greatest and have the chance for failure and catch any issues earlier, then they can set the environment variable to stable, and can even do so globally.

> This project in general has not been great at keeping up to date with dependency versions

Then this needs to be improved. There are now at least three PMC members that focus primarily on Daffodil VS Code (as compared to Daffodil itself, I definitely understand there are other non Daffodil related priorities) and can review and merge those changes. If the number is too overwhelming then we should make changes to only submit PRs for the most critical dependencies or only major releases or something, and ignore all the others.

src/tests/runTest.ts
stevedlawrence: Note that the yarn.lock file pins vscode to 1.71.0, so that's probably what we actually test with. I'm not familiar enough with VS Code to know what a reasonable minimum target it.

scholarsmate: 1.55.0 failed the tests, probably because the test suite wasn't in place at that time.  I upped the minimum version to 1.60.2, released October 4th, 2021 as this version works with the current test cases and has support for passing messages with array buffers that we need for the upcoming Data Editor.  I also made that the default for testing against rather than stable.

scholarsmate:
@stevedlawrence can you review the most recent changes when you get a chance?  Thanks.

package.json
```diff
@@ -55,7 +55,7 @@
 		"@types/glob": "^7.1.3",
 		"@types/mocha": "^9.1.1",
 		"@types/node": "^18.0.0",
-		"@types/vscode": "^1.55.0",
+		"@types/vscode": "^1.60.2",
```
stevedlawrence, line 58: Does yarn.lock need to be updated? It still references 1.55 as the minimum and 1.71 as the actual version.

src/tests/runTest.ts
```diff
@@ -18,13 +18,41 @@
...
 import {
   runTests,
   resolveCliArgsFromVSCodeExecutablePath,
   downloadAndUnzipVSCode,
 } from '@vscode/test-electron'
 
+// VS Code version 1.60.2, released October 4, 2021, is the earliest version
+// supported as indicated in package.json engines/vscode, so that version is a
+// reasonable default.
+const DEFAULT_DAFFODIL_TEST_VSCODE_VERSION = '1.60.2'
```
stevedlawrence, line 31: Can we pull this out of package.json or yarn.lock so we don't need to remember to update this file when one of those files is updated?

stevedlawrence (approved):
+1 :+1: minor comments

src/tests/runTest.ts
scholarsmate: Good idea!  I'll get that added.

scholarsmate:
@Shanedell, can you please review this when you get a chance?  Thanks!

Shanedell (approved):
+1 👍

PR #409, opened 2023-01-10
Rename the variable wholeLine to triggerText

### LINKED ISSUE #392: Rename the Wholeline variable to a more descriptive name
The variable called Wholeline contains the line of code where intellisense was triggered.  The value in this variable is looked at to try to determine the context in which intellisense was triggered. A more appropriate name might be TriggerText.

rt320:
-Renamed to variable wholeLine to triggerText to better describe it use. -Initially triggerText contains the text of the line where auto-complete is triggered

Closes #392

### DIFF
--- a/src/language/providers/attributeCompletion.ts
+++ b/src/language/providers/attributeCompletion.ts
@@ -63,7 +63,7 @@ export function getAttributeCompletionProvider() {
         document: vscode.TextDocument,
         position: vscode.Position
       ) {
-        const wholeLine = document
+        const triggerText = document
           .lineAt(position)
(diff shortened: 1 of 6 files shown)


stevedlawrence (approved):
+1 looks good :+1:

Though, just to clarify, I don't think this resolves issue #165, which has a similar title but is really more about fixing the logic for how we figure out the context for what to autocomplete.

scholarsmate (approved):
This is a simple rename of wholeLine to triggerText to be more indicative of its purpose.  I think this is a good refactor.

PR #414, opened 2023-01-11
Node and Scala Package updates

### LINKED ISSUE #372: Dependabot and scala-steward limit running
The MRs created by dependabot and scala-steward are becoming a bit much to keep up with. I think we should change how often these execute and open PRs.

Currently this how often they run:

```txt
- dependabot:
  * npm --> weekly
  * github-actions --> weekly

- scala-steward --> weekly
```

I propose updating all 3 to only run monthly or every other week.

Shanedell:
Node and Scala Package updates:

- Scala:
  - sbt-native-packager 1.9.9 to 1.9.11.
  - fs2-io 3.2.10 to 3.2.14.
  - log4cats-slf4j 2.4.0 to 2.5.0.
  - decline-effect 2.3.0 to 2.3.1.
  - scalafmt-core 2.7.5 to 3.6.1.
    - Updated .scalafmt.conf to use scala212 as the dialect.
  - sbt-scalafmt 2.4.6 to 2.5.0.
  - sbt-tpolecat 0.1.22 to 0.4.2.
  - sbt 1.7.3 to 1.8.2.
  - Removed sbt-json package as caused dependecy issues and was not completely necessary.
  - Update build.sbt to get daffodilVersion using regex instead of json.
  - Update formatting of scala and sbt files based on new scalafmt version.
  - Update scala-steward to ignore logback-classic:
    - Upgrading to 1.4.x does not support Java 8.
    - Upgrading to 1.3.x also was causing some issues about a class not being found.
- Node:
  - uuid 8.3.2 to 9.0.0.
  - xdg-app-paths 7.3.0 to 8.2.0.
  - @types/glob 7.1.3 to 8.0.0.
  - @vscode/test-electron ^2.1.5 to 2.1.5, going to a higher version required node 16.
  - vsce 2.9.2 to 2.15.0.
  - webpack 5.73.0 to 5.75.0.
  - Update how to check for arg -k/--disable_cert_verification.
    - Needed updated as parseArgs from 'util' is a node 19 thing, not sure how ci passed with this in.
- Updated dependabot and scala-steward bots to only run once a month.

Closes https://github.com/apache/daffodil-vscode/issues/372
Closes https://github.com/apache/daffodil-vscode/issues/374

### DIFF
--- a/project/Rat.scala
+++ b/project/Rat.scala
@@ -22,8 +22,8 @@ object Rat {
   lazy val excludes = Seq(
     // git files
     file(".git"),
-    /** Can't add license headers in JSON files.
-      * Adding a license attribute breaks things in some of these fiels as well.
+    /** Can't add license headers in JSON files. Adding a license attribute breaks things in some of these files as
+      * well.
       */
     file("language/dfdl.json"),
     file("language/syntax.json"),
--- a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
+++ b/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
@@ -173,7 +173,7 @@ class DAPodil(
       case extract(Command.PAUSE, _)                                     => pause(request)
       case extract(Command.DISCONNECT, args: DisconnectArguments)        => disconnect(request, args)
       case extract(Command.EVALUATE, args: EvaluateArguments)            => eval(request, args)
-      case _                                                             => Logger[IO].warn(show"unhandled request $request") *> session.sendResponse(request.respondFailure())
+      case _ => Logger[IO].warn(show"unhandled request $request") *> session.sendResponse(request.respondFailure())
     }
 
   /** State.Uninitialized -> State.Initialized */
@@ -475,7 +475,9 @@ object DAPodil extends IOApp {
       _ <- Logger[IO].info(s"disconnected at $uri")
     } yield done
 
-  /** Returns a resource that launches the "DAPodil" debugger given a DAP session, returning an effect that waits until the debugger stops or the session ends. */
+  /** Returns a resource that launches the "DAPodil" debugger given a DAP session, returning an effect that waits until
+    * the debugger stops or the session ends.
+    */
   def resource(
       session: DAPSession[Request, Response, DebugEvent],
       debugee: Request => EitherNel[String, Resource[IO, Debugee]]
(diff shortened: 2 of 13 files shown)


project/Rat.scala
```diff
@@ -22,8 +22,8 @@ object Rat {
   lazy val excludes = Seq(
     // git files
     file(".git"),
-    /** Can't add license headers in JSON files.
-      * Adding a license attribute breaks things in some of these fiels as well.
+    /** Can't add license headers in JSON files. Adding a license attribute breaks things in some of these fiels as
```
scholarsmate, line 25: fiels -> files

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -173,7 +173,7 @@ class DAPodil(
       case extract(Command.PAUSE, _)                                     => pause(request)
       case extract(Command.DISCONNECT, args: DisconnectArguments)        => disconnect(request, args)
       case extract(Command.EVALUATE, args: EvaluateArguments)            => eval(request, args)
-      case _                                                             => Logger[IO].warn(show"unhandled request $request") *> session.sendResponse(request.respondFailure())
```
scholarsmate, line 176: You probably want to keep the original formatting.

Shanedell: The updated version of scalafmt says that is the proper formatting

src/tests/runTest.ts
```diff
@@ -18,7 +18,6 @@
 // All tests ran here are ones that require the vscode API
 import * as path from 'path'
 import * as cp from 'child_process'
-import { parseArgs } from 'util'
```
Shanedell, line 21: This a node 19 thing only. The updated way should work for all nodejs version. Not completely sure how the CI didn't fail because of this, wondering if yarn.lock didn't get fully updated correctly.

scholarsmate: I like your change.  We don't need a very powerful argument parser here (yet).

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
scholarsmate: It also says to drive into a lake.  Not a big deal.  I'll allow it.

src/tests/runTest.ts
Shanedell: Yeah I mean if we need something more in depth it would possibly be best to do it ourselves.

scholarsmate (approved):
+1, LGTM.

scholarsmate:
@stevedlawrence, can you review this PR when you get a chance?  Thanks!

arosien:
@Shanedell needs to be squashed to 1 commit, but otherwise 👍.

arosien (approved):
@Shanedell needs to be squashed to 1 commit, but otherwise 👍.

Issue #76, opened 2022-02-02
Support loading and breakpoints of schemas on classpath


arosien:
(This issue is extracted from part of #66.)

Schemas may be included as resources on the classpath, which supports both file-system directories and jar files. The debugger extension, however, doesn't currently know about these resources in order to view them or set breakpoints within them.

One way to supply the classpath schemas to the editor, to view and to let the user set breakpoints there,  is to send [`LoadedSource` DAP events](https://microsoft.github.io/debug-adapter-protocol/specification#Events_LoadedSource) from our backend to VS Code when we notice any referenced schema outside of the initial schema. We can do this eagerly (scan the XML, etc.) or lazily (as the schema is processed we notify when a new schema is encountered). The editor then will request the source content, and the backend can send it (since we can access the data of the jar file on the classpath), and the editor will display the (read-only) content. Implementation-wise, we need to re-enable the `supportsLoadedSourcesRequest` initialization parameter that was disabled in #25, decide the schema source discovery strategy, and handle the appropriate events and requests.

Additionally, I suspect the default [`pathFormat` parameter of the `Initialize` request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Initialize) needs to change from `path` to `uri`, since the loaded source will have a `jar` URI scheme (because it is loaded from a jar on the classpath).

arosien:
@mbeckerle @stevedlawrence, are there any APIs within Daffodil that provide the tree of schema dependencies?

mbeckerle:
Hmmm. Not a published supported API, but I believe there is an
allSchemaFiles thing. Let me look.

mbeckerle:
ProcessorFactory has an sset member that gives access to the SchemaSet object, which has allSchemaFiles.
Each is a DFDLSchemaFile object, which has public member uriString.

arosien:
> ProcessorFactory has an sset member that gives access to the SchemaSet object, which has allSchemaFiles.
> Each is a DFDLSchemaFile object, which has public member uriString.

@mbeckerle one issue with the `sapi.ProcessorFactory` is that it has a much more restricted interface than `core.compiler.ProcessorFactory`. In particular, the `sset` stuff is only available on `core.compiler.ProcessorFactory`. I'm wondering what you would recommend; I'm thinking either the extension uses `core.compiler` directly, or perhaps some interface changes on Daffodil itself.

scholarsmate:
Medium priority, but needs developer availability to complete.

PR #417, opened 2023-01-12
Depend on another extension to load schema files from jars.

### LINKED ISSUE #265: XSD accessed via a jar file can be stepped into, not displayed
The server will need to provide a way to serve files over to the VSCode debugger, as currently an XSD accessed via a jar file, can be stepped into, but is not displayed. Interesting that you can continue to step and watch the progress of the building of the infoset. You just can't see the source.  However, when I put the ethernetIP/src/main/resources *directory* on the classpath, and the ethernetIP/target/classes directory on the classpath (in the launch.json), now I can step into the ethernet IP schema files just fine. The fact that this schema depends on a compiled scala class for a plugin caused no issues.

arosien:
It registers a content handler which provides virtual document support for jar URIs. Requires the Metals extension >=1.13.0, but the latest version is 1.21.0 anyway.

Fixes #265.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,8 @@
 		"webpack-cli": "4.10.0"
 	},
 	"extensionDependencies": [
-		"vincaslt.highlight-matching-tag"
+		"vincaslt.highlight-matching-tag",
+		"wmanth.jar-viewer"
 	],
 	"main": "./dist/ext/extension.js",
 	"activationEvents": [
--- a/src/tests/runTest.ts
+++ b/src/tests/runTest.ts
@@ -69,7 +69,13 @@ async function main() {
     // Install required extensions
     cp.spawnSync(
       cli,
-      [...args, '--install-extension', 'vincaslt.highlight-matching-tag'],
+      [
+        ...args,
+        '--install-extension',
+        'vincaslt.highlight-matching-tag',
+        '--install-extension',
+        'wmanth.jar-viewer',
+      ],
       {
         encoding: 'utf-8',
         stdio: 'inherit',


mbeckerle (approved):
+1 If scala's metals license is ok. 

Please verify their license allows us to depend on it.

arosien:
> +1 If scala's metals license is ok.
> 
> Please verify their license allows us to depend on it.

[Their license](https://github.com/scalameta/metals-vscode/blob/main/LICENSE) is Apache 2.0.

stevedlawrence:
I found https://github.com/wmanth/vscode-jar-viewer, which also has the ability to view files in jars. Seems it works by creating a new TextDocumentContentProvider:

https://vshaxe.github.io/vscode-extern/vscode/TextDocumentContentProvider.html

Creating a custom one for our use cases feels like it might be a better approach and gives us more control over how the file is opened--maybe we can eventually provide more useful features since we know it's a DFDL schema.

stevedlawrence:
> Creating a custom one for our use cases feels like it might be a better approach and gives us more control over how the file is opened--maybe we can eventually provide more useful features since we know it's a DFDL schema.

Actually, looking at that extension more, maybe that's unique to opening just a Jar file, and not a file:// URI that points inside a jar. Regardless, it feels like something along those lines might be a better approach than depending on the metals extension.

arosien:
> I found https://github.com/wmanth/vscode-jar-viewer, which also has the ability to view files in jars.

I replaced Metals with the jar-viewer plugin and it loads the jar-contained schemas the same way:
Screen Shot 2023-01-12 at 2 39 11 PM

It uses the [MIT License](https://github.com/wmanth/vscode-jar-viewer/blob/master/LICENSE.md).

arosien:
Also, since this extension allows browsing of jar files, it could perhaps also be used to fix #76 by allowing navigation into the schema files. I'll test that out also.

stevedlawrence (approved):
+1, sounds good to me.

I imagine someday we might want more control that what this provides. For example automatically search jars on the classpath containing DFDL schemas and list the schemas instead of having to use this extension to manually open a jar and select files like it looks like how this extension works. But I think this is a good interim solution to give us 90% of what we need without much extra stuff.

Shanedell:
@arosien I believe your issue is actually something that I fix in #414 , where the a parseArgs method is a nodejs 19 thing. If you look at the `yarn test` for  https://github.com/apache/daffodil-vscode/actions/runs/3907091146/jobs/6688638589 it actually fails because that reason. So I am thinking you will need that fix here

Shanedell:
@arosien Would you be able to approve #414? This will fix the `yarn test` failing in the CI (but still passing for some reason). Then once that is merged if you rebase and fix the format issue I will approve this.

arosien:
> @arosien I believe your issue is actually something that I fix in #414 , where the a parseArgs method is a nodejs 19 thing. If you look at the `yarn test` it actually fails because that reason, not sure why it turns green though. So I am thinking you will need that fix here.
> 
> (the line you added to install that extension in the test was a good catch also just believe it will still fail for the above reason)

@Shanedell I reran the failing check and it passed, so it was some nondeterministic effect going on. I'll look at #414 anyway. Thanks.

PR #427, opened 2023-01-19
Remove child_process as dependency, it is provided by nodejs when installed.

### LINKED ISSUE #260: Investigate switching from child_process
Based on discussion on PR https://github.com/apache/daffodil-vscode/pull/251, it seems the `child_process` is not actively maintained and that the package is available to actually be acquired. This should be investigated to why it is still the recommend package to use or investigate a different node package to use as an alternative.

Shanedell:
Remove child_process as dependency, it is provided by nodejs when installed.

Closes #260


scholarsmate (approved):
I did the same in the data_editor branch.  There are a few more that can be removed as well.

Shanedell:
@stevedlawrence The `@grpc/grpc-js` and `google-protobuf` can also be removed would that be okay to add to this PR?

Shanedell:
Removing this in favor of  #442

PR #429, opened 2023-01-20
	Modifications to try to ensure intellisense offers relevant

rt320:
suggestions when there are multiple tags on a single line and
	when one or more end tags have been removed

	-Logic was introduced to determine if multiple tags are on a
	single line
	-The nearestOpen funtion was modified to take in to account
	multiple tags on a single line
	-The logic for closeElement and cloesElementSlash was modified to
	supply the end tag for previous items whether on a single line
	or multiple lines where the end tag is missing or has been
	deleted. The triggers '/' or '>' should now provide end tags if
	they are missing

	Closes #309

### DIFF
--- a/src/language/providers/closeElement.ts
+++ b/src/language/providers/closeElement.ts
@@ -16,7 +16,12 @@
  */
 
 import * as vscode from 'vscode'
-import { getXsdNsPrefix, insertSnippet, nearestOpen } from './utils'
+import {
+  getXsdNsPrefix,
+  insertSnippet,
+  nearestOpen,
+  getItemsOnLineCount,
+} from './utils'
 
 export function getCloseElementProvider() {
   return vscode.languages.registerCompletionItemProvider(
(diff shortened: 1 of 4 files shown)


src/language/providers/closeElement.ts
```diff
@@ -27,98 +32,207 @@ export function getCloseElementProvider() {
         position: vscode.Position
       ) {
         var backpos = position.with(position.line, position.character - 1)
+        var backpos3 = position.with(position.line, position.character - 3)
```
scholarsmate, line 35: We should be using either `let` for mutable variables or `const` for immutable variables.

src/language/providers/closeElement.ts
```diff
@@ -27,98 +32,207 @@ export function getCloseElementProvider() {
...
+                  insertSnippet(
+                    '>\n\t$0\n</' + nsPrefix + 'simpleType>',
+                    backpos
+                  )
+                } else if (
+                  triggerText.endsWith('>') &&
+                  triggerText.includes('dfdl:defineVariable') &&
+                  nearestOpenItem.includes('defineVariable')
+                ) {
+                  var startPos = document.lineAt(position).text.indexOf('<', 0)
```
scholarsmate, line 115: Many / All of the `var` in here should be `const`.

src/language/providers/utils.ts
```diff
@@ -160,7 +274,7 @@ export function checkElementOpen(
       triggerText.includes('<' + nsPrefix + 'element') &&
       !triggerText.includes('</' + nsPrefix + 'element') &&
       !triggerText.includes('/>') &&
-      !triggerText.includes('>')
+      !triggerText.includes('/element>')
```
scholarsmate, line 277: Is this right?

src/language/providers/utils.ts
```diff
@@ -312,10 +528,13 @@ export function checkBraceOpen(
...
     //.text.substring(0, document.lineAt(lineNum).range.end.character)
 
+    if (!triggerText.includes('{')) {
+      return false
+    }
     if (
       triggerText.includes('"{') &&
       triggerText.includes('}"') &&
-      triggerText.includes('..') &&
+      (triggerText.includes('..') || triggerText.includes('.')) &&
```
scholarsmate, line 537: This can be simplified to just `triggerText.includes('.')`.

stevedlawrence:
What issue is this resolving? The ticket mentions #309, but that's a pull request and not an issue. Do you mean #165?

As to this change, I have trouble following the logic with all the nested if statemens, which makes me think this might not be best approach. We've been okay with it so far, but it's honestly too much to follow and ensure that we're doing the right thing, and as things get even more complex I don't see this approach being maintainable.

I think maybe a new algorithm is really needed to figure the correct context for autocompletion instead of just a bunch of if-statements. If you want to have a phone call or email thread to discuss some alternative approaches I'm happy to do so.

rt320:
Code fixes will be present in a new PR

PR #423, opened 2023-01-16
WIP: initial integration of the new data editor

scholarsmate:
**This is a work in progress**

New data editor webview extension.

closes #311
closes #364
closes #145
closes #143
closes #141
closes #140
closes #139
closes #138
closes #137
closes #132
closes #131
closes #130

PR #433, opened 2023-01-24
integrate vscode-extension-tester (#432)

### LINKED ISSUE #432: Integrate test mechanism to query webview DOM
A package that can be used to query the `webview` can be found here: https://github.com/redhat-developer/vscode-extension-tester.  It allows for many kinds of tests and automations, including `webview` DOMs (https://github.com/redhat-developer/vscode-extension-tester/wiki/WebView).  It may also be useful for testing code completion.  The bundled `extest` script appears to offer the same functionality as `src/tests/runTests.ts` but also integrates selenium web drivers.

scholarsmate:
This PR allows for testing strictly from the UI perspective (e.g., not using the `vscode` API) using a web driver (selenium driving chromium in this case).  This allows us to get to the entire DOM of the VS Code UI and interact with it the way an actual user would (users use the UI, not the API).

Access to the DOM is a required ingredient for writing tests for the Data Editor since that uses a custom VS Code webview that you cannot interact with otherwise (e.g., you can't interact with the webview DOM via the `vscode` API).

NOTE: When creating tests run by a web driver, sometimes the element that you want to select and interact with doesn't exist yet because the DOM hasn't been updated yet by VS Code.  This is where upping the timeout and adding pauses comes in.  The pauses are put in certain locations where it may take a bit for VS Code to change the DOM, and we need to wait a bit for the element you want to select to exist.  I tuned these pauses based on trial and error on my laptop, but may need to be upped further for CI to pass.

blocks #423 
closes #432


src/tests/suite-ui/omegaEdit-driver.test.ts
```diff
@@ -0,0 +1,171 @@
+/*
```
Shanedell, line 1: My main comment for this file is that I would like to see more whitespaces. I feel as those some of these lines can be broken up more

Shanedell:
@scholarsmate I also think you need to add `yarn test-ui` to the `.github/workflows/CI.yml` and `github/workflows/nighly.yml` after each `yarn test`.

src/tests/suite-ui/editor-driver.test.ts
```diff
@@ -0,0 +1,75 @@
+/*
```
Shanedell, line 1: I also don't understand the point of this file, couldn't we just point to the documentation/samples for the `vscode-extension-tester` testing framework instead of having these tests?

scholarsmate: It's good to have at least 2 test files in the suite to make sure each test file cleans up properly.  This was a generic test used for reference and to make sure things got cleaned up.  I guess its purpose has been served, and I removed it.

scholarsmate:
> @scholarsmate I also think you need to add `yarn test-ui` to the `.github/workflows/CI.yml` and `github/workflows/nighly.yml` after each `yarn test`.

@Shanedell, done.  Let's hope it turns out well.

src/tests/suite-ui/omegaEdit-driver.test.ts
scholarsmate: As you wish.

Shanedell:
This is still a work in progress so I am going to close this PR and reopen once we have it working properly.

PR #434, opened 2023-01-24
Use ClasspathJarPlugin for packaging.

### LINKED ISSUE #431: APP_CLASSPATH Length Issue
Currently the `APP_CLASSPATH` is set via one line that list all jars needed. However, when it comes to adding new Scala dependencies (eg like in the TDML PR), it can make this line very long. When the line is too long, at least in Windows, the script will fail to run.

So is there a way we can break up this one liner into multiple lines or make it so its not going to effect the script?

Ideas:

- @scholarsmate Did have an idea of creating a JAR that contains a manifest file whose `Class-Path` specifies the disk paths of all the jars. This way only that single JAR, with the manifest, would need to be added to the `APP_CLASSPATH`. Would this be a good solution?

Information:

- Tagging @stevedlawrence @mbeckerle @tuxji in case you guys have any input, since this maybe is something that happened in Daffodil as well.

arosien:
Avoids command-line length issues occurring on Windows. Fixes #431.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,8 +66,6 @@ lazy val ratSettings = Seq(
   ratFailBinaries := true
 )
 
-lazy val commonPlugins = Seq(BuildInfoPlugin, JavaAppPackaging, UniversalPlugin)
-
 lazy val `daffodil-debugger` = project
   .in(file("."))
   .settings(commonSettings, ratSettings)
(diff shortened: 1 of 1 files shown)


Shanedell (approved):
+1 👍 Saw this fixed an issue in the TDML PR.

scholarsmate (approved):

PR #436, opened 2023-01-26
Relay diagnostics in the "Parse" section of variables.

### LINKED ISSUE #269: Need a way to have the run do daffodil's 'limited validation'
Need a way to have the run do daffodil's 'limited validation'. Basically to see the see the stream of validation errors as they are produced. This is just another option to set on daffodil when creating the Daffodil Data Processor. There's a `withValidation(...)` call.

arosien:
Limited validation is enabled and diagnostics are separated into validations and errors.
Fixes #269.

### DIFF
--- a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
+++ b/server/core/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
@@ -33,6 +33,7 @@ import java.io._
 import java.net.URI
 import java.nio.file._
 import org.apache.commons.io.FileUtils
+import org.apache.daffodil.api.Diagnostic
 import org.apache.daffodil.debugger.dap.{BuildInfo => DAPBuildInfo}
 import org.apache.daffodil.debugger.Debugger
 import org.apache.daffodil.exceptions.SchemaFileLocation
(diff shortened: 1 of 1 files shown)


mbeckerle (approved):
+1

For the future... new tickets maybe,.... The thought struck me that we have the parse output, the various state of the parse including variables, and now the diagnostics. The only other things daffodil does are (a) log4J log output (b) crash dumps (when an assertion fails - i.e., it's a daffodil bug). 

We should probably have a way for the log to be inspected by the GUI/IDE, and if daffodil hits one of its crashes, that should be brought to the attention of the user without them losing their state or having the GUI stop responding, etc. Ideally if they crash daffodil we'd like a bug report, so the IDE has to not crash just because Daffodil did.

Shanedell (approved):
+1 👍

arosien:
> +1
> 
> For the future... new tickets maybe,.... The thought struck me that we have the parse output, the various state of the parse including variables, and now the diagnostics. The only other things daffodil does are (a) log4J log output (b) crash dumps (when an assertion fails - i.e., it's a daffodil bug).
> 
> We should probably have a way for the log to be inspected by the GUI/IDE, and if daffodil hits one of its crashes, that should be brought to the attention of the user without them losing their state or having the GUI stop responding, etc. Ideally if they crash daffodil we'd like a bug report, so the IDE has to not crash just because Daffodil did.

The log is printed to a terminal console, where it can be viewed during debugging. Do you mean have the log persisted to a file somewhere? We can do that.

I'm interested in hearing more about the crashing situation. Currently the Daffodil session runs within the same process as the DAP debug session; if it became a requirement to detect a crash we could potentially fork the parsing process.

PR #440, opened 2023-01-31
Update omega-edit to allow configuration of the server port.

### LINKED ISSUE #271: Make the Omega-Edit RPC server port configurable
Currently the Omega-Edit RPC server port is hard coded to 9000.  I think that is a reasonable default, but there is always a chance that port 9000 might not be available, in which case another port must be used instead.  Make it so that the end user can modify the port setting.

Shanedell:
Update omega-edit to allow configuration of the server port:

- Update let terminal to be const terminal.
- Change var client to be let client, and defined the type expected for client.
- Cast unzip as NodeJS.NodeJS.WritableStream
- Add information message for terminal exit code when it exists.

Closes https://github.com/apache/daffodil-vscode/issues/271

### DIFF
--- a/src/omega_edit/client.ts
+++ b/src/omega_edit/client.ts
@@ -20,7 +20,11 @@ import * as fs from 'fs'
 import * as omegaEditSession from 'omega-edit/session'
 import * as omegaEditViewport from 'omega-edit/viewport'
 import * as omegaEditVersion from 'omega-edit/version'
-import { startOmegaEditServer, viewportSubscribe } from './utils'
+import {
+  startOmegaEditServer,
+  viewportSubscribe,
+  initOmegaEditClient,
+} from './utils'
 import { OmegaEdit } from './omega_edit'
 import { v4 as uuidv4 } from 'uuid'
 import XDGAppPaths from 'xdg-app-paths'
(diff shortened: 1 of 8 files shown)


src/omega_edit/client.ts
```diff
@@ -50,17 +54,38 @@ async function cleanupViewportSession(
...
+      port
     )
   }
 }
 
+async function getOmegaEditPort(
+  port: number | undefined = undefined
+): Promise<number> {
+  if (!port) {
+    let portEntered = await vscode.window.showInputBox({
```
scholarsmate, line 74: `const` can be used here.

src/omega_edit/client.ts
```diff
@@ -69,8 +94,13 @@ export function activate(ctx: vscode.ExtensionContext) {
...
-      async (startServer: boolean = true) => {
-        await commonOmegaEdit(ctx, startServer, omegaEditPackageVersion)
+      async (
+        startServer: boolean = true,
+        port: number | undefined = undefined
+      ) => {
+        port = await getOmegaEditPort(port)
+        initOmegaEditClient('127.0.0.1', port.toString())
+        await commonOmegaEdit(ctx, startServer, omegaEditPackageVersion, port)
         return await omegaEditVersion.getVersion()
```
scholarsmate, line 104: This is invisible to the end-user using the GUI.  Should put the version in an info message as well.

src/omega_edit/utils.ts
```diff
@@ -26,7 +26,13 @@ import { getClient, ALL_EVENTS } from 'omega-edit/settings'
 import * as omegaEditServer from 'omega-edit/server'
 import { runScript } from '../utils'
 
-const client = getClient()
+var client
```
scholarsmate, line 29: `var` should be `let`.  Should specify the `client` type here as well.

src/omega_edit/utils.ts
```diff
@@ -138,14 +144,29 @@ export async function viewportSubscribe(
...
   const [scriptName, scriptPath] = await omegaEditServer.setupServer(
     rootPath,
     omegaEditPackageVersion,
     ctx.asAbsolutePath('./node_modules/omega-edit')
   )
 
-  let terminal = await runScript(scriptPath, scriptName)
+  process.env.OMEGA_EDIT_SERVER_PORT = port.toString()
+
+  let terminal = await runScript(
```
scholarsmate, line 158: `let` could be `const` here.  Would it be worth sending the UI an info/warn message about the server starting up successfully or not?

src/utils.ts
```diff
@@ -143,7 +143,13 @@ export async function unzipFile(zipFilePath: string, extractPath: string) {
   return await new Promise((resolve, reject) => {
     let stream = fs
       .createReadStream(zipFilePath)
+      /* Ignore error that type Extract is not WritableStream. If converted to
+       * unzip.Extract({ path: `${extractPath}` }).unzipStream it fails completely
+       * but keeping it this way works fine
+       */
+      // @ts-ignore
```
scholarsmate, line 150: Can you cast it `as WritableStream`?

scholarsmate (approved):

scholarsmate:
@mbeckerle or @stevedlawrence, will you review this PR at your earliest convenience?  Thanks.

mbeckerle (approved):
+1

PR #451, opened 2023-02-13
Support "variables" and "tunables" launch arguments.

### LINKED ISSUE #200: Need to be able to supply values for variables.
When running Daffodil from the command line, it is possible to supply values for variables using:
```
-D VARIABLE=VALUE 
```
A method to provide these values when working with the debugger is needed.

arosien:
- "variables" is a JSON object with string values that is passed as Daffodil "external variables" to the parse. Variables are validated at runtime when the parse begins, as they are schema-specific.
- "tunables" is a JSON object with string values that is passed as Daffodil tunables to the parse. Only the allowed tunable keys, specified at https://daffodil.apache.org/configuration/, are allowed.

Fixes #200.


package.json
```diff
@@ -1,491 +1,545 @@
...
+                "description": "Additional classpaths to be exported to the debugger",
+                "default": ""
+              },
+              "variables": {
+                "type": "object",
+                "additionalProperties": {
+                  "type": "string"
+                }
+              },
+              "tunables": {
```
mbeckerle, line 361: Can json files have comments? Keeping this list synchronized with any new tunables added in daffodil's tunables XML seems worth commenting on both here and in that XSD file. 

Another idea: We generate  scala code off that XSD file now, we could generate this chunk of JSON too.

mbeckerle (approved):
+1

package.json
stevedlawrence: Agreed on the second idea. The `dafext.xsd` file where all these tunables are defined (including descriptions, min/max values, defaults) is in the daffodil-lib jar. If possible, it seems like better idea to extract this information at extension initialization time nased on whatever version of daffodil-lib is being used. Having to manually update this file is a losing battle--new tunables are added all the time, we already new ones for the 3.5.0 release.

arosien: Unfortunately JSON doesn't have comments.

There's no mechanism to specify some kind of file inclusion in this `package.json` data format, so I think we would need, like you say, an external process to produce the valid tunables and inject them into this configuration file. That's a non-trivial amount of work, so how about I make a new issue for that along with (optionally) removing the explicit tunable names from this file?

stevedlawrence: Why inject them into the file at all? Can't this list be generated at extension load time and dynamically used where needed? Seems something that changes as often as the list of tunables does shouldn't be hardcoded anywhere.

Also, taking a second look, this is missing a large number of tunables. Is there a reason to not include the full list?

mbeckerle: Probably our tunables doc is also obsolete - because new ones get added, and the doc gets stale, which just makes the argument for generating this on the fly, at runtime, and the UI can pull it over from Daffodil at startup.

arosien: > Why inject them into the file at all? Can't this list be generated at extension load time and dynamically used where needed? Seems something that changes as often as the list of tunables does shouldn't be hardcoded anywhere.

The package.json file is the build file for the extension, not the runtime config for it, and I'm not sure there is any other mechanism to configure the allowed launch config schema.

After doing some searches, there are a few projects which seem to run an external program which modify the package.json itself. But for this PR in particular I'd rather separate that work as a new issue.

stevedlawrence: Alternative idea just don't validate the list of tunables and just let it be a key/value pair just like the variable property. Daffodil will throw an exception if you try to set a tunable that doesn't exist, so you can catch it that point and let the user know it's invalid (the same should happen for variables I believe?).

Fine with me if changes are in a separate PR.

arosien: > Alternative idea just don't validate the list of tunables and just let it be a key/value pair just like the variable property. Daffodil will throw an exception if you try to set a tunable that doesn't exist, so you can catch it that point and let the user know it's invalid (the same should happen for variables I believe?).

That's what I was going to suggest also as a compromise, excellent idea!

package.json
```diff
@@ -1,491 +1,503 @@
...
-			}
-		]
-	},
-	"__metadata": {
-		"id": "4f1304da-7e65-48f1-9126-0a143e2e5ef2",
-		"publisherDisplayName": "Apache Software Foundation",
-		"publisherId": "4dbc1d1a-d64b-46f8-8756-1c234855f645",
-		"isPreReleaseVersion": false
-	}
+  "name": "apache-daffodil-vscode",
```
arosien, line 2: A previous commit changed all spaces to tabs, this changes it back.

Shanedell (approved):
+1 👍

PR #453, opened 2023-02-15
conditionally require wait_port

scholarsmate:
Require wait_port only when it is needed.


Shanedell (approved):
+1 👍

arosien (approved):
👍 

But what's the motivation for this?

scholarsmate:
> 👍
> 
> But what's the motivation for this?

Just a code clean up.  Bring in `wait_port` only when it's being used.

Issue #438, opened 2023-01-31
Simplify intellisense logic to determine what tag the cursor is in when triggered


rt320:
Currently there are multiple nested if statements that attempt to determine which tag is triggering intellisense.  The logic attempts to determine if a closing tag for the item exists and uses separate logic if more than one tag is on the same line.  Separate but similar logic is used for each tag type (sequence, element, group, etc).  The logic should be simplified to determine if the cursor at the trigger position is after an opening tag (ex: <xs:sequence> [cursor is here] ), or if the cursor is after the open tag symbol and a tag type (tag type ex: sequence, group, element, etc), (ex: <xs:element name="foo" [cursor is here] )) and used in most scenarios.  The logic for triggering closing tags should also be looked at for simplification.

Issue #366, opened 2022-12-02
1.2.0 Intellisense issues


mbeckerle:
This is a long ticket description. Perhaps can be broken out into separate issues over time if that helps, but until someone with more experience of the intellisense code looks at this, I don't want to do that preemptively. 

- I was unable to figure out how to enable intellisense from the
    documentation here:
    https://github.com/apache/daffodil-vscode/wiki/Apache-Daffodil%E2%84%A2-Extension-for-Visual-Studio-Code
    The animation showing how to do that goes too quickly, and I cannot see how
    the features it is using are being activated.

    Intellisense does seem to work however, after I set a file association of
    "*.dfdl.xsd" to "dfdl".  Actually it may have just worked anyway, as I'm
    not sure this setting was needed.

    I created a new foo.dfdl.xsd schema file with it.

 -   The completion for a schema chooses elementFormDefault="qualified" which is
    incorrect. (a bug) The preferred style for DFDL schemas (or any XSD) should
    be elementFormDefault="unqualified".
    Alas, I wish I had known this when we created the TDML language. It would
    be so much nicer without the "tdml:" prefixes required everywhere.

 -   The intellisense completion for a schema provides this namespace prefix:
    xmlns:math="[www.w3.org/2005/xpath-functions/math](http://www.w3.org/2005/xpath-functions/math)"
    which contains functions not supported by the DFDL schema language.
    So I am not sure why this namespace prefix is provided. Perhaps these
    functions can be used in schematron annotations or other annotations than
    DFDL?

 -   I wanted to insert a dfdl:format annotation. I typed "for" and was prompted
    with dfdl:format. Clicking on that embeds a giant pile of code as if this
    was the start of an empty dfdl schema file. This is not helpful. I was
    hoping it would insert this:

```
<xs:annotation>
    <xs:appinfo source="http://www.ogf.org/dfdl/">
       <dfdl:format />
    </xs:appinfo>
</xs:annotation>
```

and leave me positioned after the dfdl:format to add properties or a ref.
If I type "ann" I am offered xs:annotation, and clicking that inserts most
of what I wanted.

```
    <xs:annotation>
    <xs:appinfo source="http://www.ogf.org/dfdl/">

    </xs:appinfo>
    </xs:annotation>
```

- In this context, if I type dfdl:format in the middle and click on the
offered "dfdl:format" choice it still inserts the entirely wrong "whole
file" contents, in the middle of my appinfo, so clearly the context is not
being taken into account.

[...]

PR #458, opened 2023-02-20
  Significant modifications made to intellisense results

rt320:
- context logic change to use line and cursor position at trigger
    - functions developed to:
      - determine the tag at the context position
      - determine if the tag is open
      - determine if the trigger occured between tags
      - determine if a close tag is missing

    - Additional modifications made to:
      - provide above functionality if line contains multiple tags
      - provide intellisense based on tag at context position
      - simplify the logic to determine context
      - minimize if/else logic
      - fixes close tag completion for schema when close tag is missing
      - fixes tag end symbol missing problem in issue 459

    - Also fixes a bug with context when intellisense is trigger at the
    start of the line of a multi-tag line, and also if trigger at the
    end of the line of a multi-tag line

    - Incorporates modifications made by Shane Dell

    Closes https://github.com/apache/daffodil-vscode/issues/438
    Closes https://github.com/apache/daffodil-vscode/issues/459
    Closes https://github.com/apache/daffodil-vscode/issues/460

### DIFF
--- a/src/language/providers/intellisense/elementItems.ts
+++ b/src/language/providers/intellisense/elementItems.ts
@@ -25,7 +25,7 @@ export const elementCompletion = (definedVariables, dfdlFormatString, nsPrefix) 
       },
       {
         item: nsPrefix + 'schema',
-        snippetString: '<' + nsPrefix + 'schema xmlns:xs="http://www.w3.org/2001/xmlSchema"\n\t\txmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"\n\t\txmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"\n\t\txmlns:fn="http:/www.w3.org/2005/xpath-functions"\n\t\txmlns:math="www.w3.org/2005/xpath-functions/math" elementFormDefault="qualified">\n$0\n</' + nsPrefix + 'schema>',
+        snippetString: '<' + nsPrefix + 'schema xmlns:xs="http://www.w3.org/2001/xmlSchema"\n\t\txmlns:dfdl="http://www.ogf.org/dfdl/dfdl-1.0/"\n\t\txmlns:daf="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext"\n\t\txmlns:fn="http:/www.w3.org/2005/xpath-functions"\n\t\t elementFormDefault="unqualified">\n$0\n</' + nsPrefix + 'schema>',
       },
       {
         item: nsPrefix + 'element name',
@@ -48,12 +48,12 @@ export const elementCompletion = (definedVariables, dfdlFormatString, nsPrefix) 
       },
       {
         item: 'dfdl:assert',
-        snippetString: '<' + nsPrefix + 'annotation>\n\t<' + nsPrefix + 'appinfo source="http://www.ogf.org/dfdl/">\n\t\t<dfdl:assert>"<$1>"</dfdl:assert>\n\t</' + nsPrefix + 'appinfo>\n</' + nsPrefix + 'annotation>$0',
+        snippetString: '<dfdl:assert>"<$1>"</dfdl:assert>$0',
         markdownString: 'dfdl assertion test',
       },
       {
-        item: 'dfdL:discriminator',
-        snippetString: '<' + nsPrefix + 'annotation>\n\t<' + nsPrefix + 'appinfo source="http://www.ogf.org/dfdl/">\n\t\t<dfdl:discriminator test="{$1}"/>\n\t</' + nsPrefix + 'appinfo>\n</' + nsPrefix + 'annotation>$0',
+        item: 'dfdl:discriminator',
+        snippetString: '<dfdl:discriminator test="{$1}"/>$0',
         markdownString: 'dfdl discriminator test',
       },
       {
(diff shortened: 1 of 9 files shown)


rt320:
Found a problem with context when intellisense is trigger at the start of the line of a multi-tag line, and also if trigger at the end of the line of a multi-tag line.

Shanedell:
@rt320 Some of code added is too much to be added in a single place. Consider breaking out some of these long flowing items into sub methods to break the code more to make it more modular.

arosien:
I'm not sure if this is a pre-existing bug or not, but I can't seem to close a tag, the editor keeps deleting the `>` symbol after I type it:
sampleWorkspace-1677188626641

arosien:
I think there's some kind of algorithmic or infinite loop problem in this code. I can get a single XML tag to be suggested, but any further attempt doesn't show any suggestions while at the same time the "Code Helper (Plugin)" process is using 99% of a CPU.

I don't see any logging or any other more direct signal about what's causing this.

Shanedell:
@arosien I am not seeing this when I checkout the code. Do you possibly have the extension already installed through the marketplace? If so maybe debugging locally and the other installed version are clashing? I am not sure if that's the case, just a thought since I am not seeing the same behavior.

hdalsania:
> I'm not sure if this is a pre-existing bug or not, but I can't seem to close a tag, the editor keeps deleting the `>` symbol after I type it: sampleWorkspace-1677188626641 [ sampleWorkspace-1677188626641 ](https://user-images.githubusercontent.com/31163/221037370-36f42a55-302e-46e2-ac45-f3f8b52e053e.gif) [ ](https://user-images.githubusercontent.com/31163/221037370-36f42a55-302e-46e2-ac45-f3f8b52e053e.gif)

I am not seeing this behavior with Visual Code version 1.74.3. We have seen some strange code IDE related auto-completion issues with Visual Code version 1.75.1.

arosien:
> I am not seeing this behavior with Visual Code version 1.74.3. We have seen some strange code IDE related auto-completion issues with Visual Code version 1.75.1.

I have 1.75.1. What's "strange"?

I'll see if what @Shanedell describes is happening: extension conflicts.

arosien:
I've removed any older Daffodil extension and I get the deleting-end-tag behavior on VSCode 1.75.1. I'd like some logging in the extension to see what's being triggered.

Shanedell:
@arosien I actually do something similar now if I try to type in `/>` it auto deletes the `/` for some reason

Shanedell:
I thinking we should not have a `slash` completion provider as it seems to break using a `slash` inside of the document all together.

Shanedell:
Is there a possibility we would want to move to making a Language Server at some point? Some documentation is here [Language Server Guide](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide). In the [RedHat vscode-xml](https://github.com/redhat-developer/vscode-xml/) extension repo they have this code:

```ts
  // Setup autoCloseTags
  const tagProvider = (document: TextDocument, position: Position) => {
    const param = languageClient.code2ProtocolConverter.asTextDocumentPositionParams(document, position);
    const text = languageClient.sendRequest(TagCloseRequest, param);
    return text;
  };
  context.subscriptions.push(activateTagClosing(tagProvider, { xml: true, xsl: true }, ServerCommandConstants.AUTO_CLOSE_TAGS));
```

It seems to setup a provider to easily close tags and seems just less complicated than what we are doing currently. Any thoughts?

rt320:
@Shanedell There is no snippet defined for "dfdl:format ref".  The only reference I could find for it was in the file DFDLGeneralFormat.dfdl.xsd. This file seems to have been added when the code originally in dfdl.ts was split out into multiple files.  The original dfdl.ts file had a snippet defined as "defaultDfdlFormat" which is similar to the beginning of this file, but that snippet was removed.  This file seems to cause one of the issues Mike Beckerle brought up in issue #366.  I was planning to replace the file as part of the solution to issue #366.
I downloaded the RedHat XML code and tried your example above in the code. This was the result when I pressed the slash key.
image
This was the result when I pressed the '>' key.
image
I also tried deleting an end tag in a multi-tag line and having the auto-complete re-insert the closing tag. When I pressed '>' another '>' was added to the end of the line. When I pressed slash I was given the option to add the missing end tag.
image
When I chose that option, the end tag was added, but the triggering slash was left in place.
image
I haven't looked at the RedHat code. Possibly it offers a simpler coding solution, but does not seem to completely handle all situations.

Shanedell:
@rt320 That makes sense but have you seen these issues identified and possibly know why they are happening? I had made a PR to your fork to break out code and I think fix a couple issues. But I am something similar to what @mbeckerle stated in #366 where intellisense just stops working.

If I have

```xml
<xsd:element name=""
```
then add a `>` the intellisense is now broken and if I do `Ctrl + Space` it just says `Loading...` forever

Shanedell:
@rt320 So I have got it nailed down to being for that function during the looping with `ic` and `io`. In my PR to your branch it is method `getItemsForLineLT2`, its stuck in an infinite loop.

rt320:
@Shanedell I think the infinite loop in getItemsForLineLT2 might be caused by the "<= 2" on line 397.

rt320:
I manually added Shane's modifications to my latest code.  I tested the resulting code in VS Code 1.74.3 and 1.76.0.  Everything I tested worked in both versions of VS Code.

Shanedell (approved):
+1 👍 @rt320 Only comment is in your commit message you need to update `Closes 438` to be `Closes #438` so that is actually links to that issue

rt320:
I'll modify the commit message and force push

Shanedell:
@rt320 Also update you PR description with your latest commit message and don't have it surrounded with the triple ticks (```).

rt320:
Not sure what you mean by the "triple ticks"

src/language/providers/intellisense/elementItems.ts
```diff
@@ -102,12 +102,42 @@ export const elementCompletion = (definedVariables, dfdlFormatString, nsPrefix)
...
+        item: 'dfdl:defineFormat',
+        snippetString: '<dfdl:defineFormat name="$1" >\n\t$2\n</dfdl:defineFormat>$0',
+        markdownString: 'dfdl format name and configuration',
+      },
+      {
+        item: 'dfdl:defineEscapeScheme',
+        snippetString: '<dfdl:defineEscapeScheme name=$1 >\n\t$0,/dfdl:defineEscapeScheme>',
+        markdownString: 'dfdl escape character definition',
+      },
+      /*TODO not sure how to make dfdl:element work without braking xs:element
```
Shanedell, line 121: I still approve of this PR just want to state that this should be added as a new issue if not done already.

arosien (approved):
👍

PR #472, opened 2023-03-13
 Change the behavior of inserting the format tag

### LINKED ISSUE #365: Broken link on release page for 1.2.0
The link for Intellisense Instructions, click on it and it takes you to a "create a new page" situation on the Wiki.

### LINKED ISSUE #366: 1.2.0 Intellisense issues

rt320:
* added additional attribute snippets
   * modified attribution completion for format to include most attributes
   * fixed a bug where xml version and schema tags did not autocomplete

  Closes #366


michael-hoke:
+1 👍 

This should close #366 instead of #365?

arosien (approved):
👍

PR #477, opened 2023-03-14
Fix Java11+ build issues by adding run/fork true

NolanMatt:
On Java11+ issues were found during compliation that this fixes.  Backtested that it doesn't break any Java 8 builds.


scholarsmate (approved):
+1, this at least doesn't break Java 1.8 and seems to fix newer builds as well.

Shanedell (approved):
+1 👍

PR #479, opened 2023-03-15
Catch arbitrary exception thrown during CI

stricklandrbls:
- Added `.on()` error callback for catching "Call cancelled" for `viewportSubscribe()` in `./src/omega_edit/utils.ts`.


Shanedell (approved):
+1 👍 This was needed in `omega-edit` as well. It seems to happen randomly when the `omega-edit` server is shutdown but isn't actually a big issue.

scholarsmate (approved):
+1

PR #490, opened 2023-03-16
upgraded omega-edit to v0.9.55 and removed pino dependency

scholarsmate:
Ωedit v0.9.55 has an important bug fix for notify changed viewports.  Also the logging in Ωedit no longer exposes pino so that dependency is removed.


Shanedell (approved):
+1 👍

arosien (approved):
👍

PR #495, opened 2023-03-17
Set debugServer to 4711 (default) if not specified in config.

### LINKED ISSUE #494: Extension doesn't connect to launched backend if debugServer launch property missing
If you don't include `"debugServer": 4711` in the `launch.json` config then the debug session doesn't connect to the launched debugger backend (which is listening on 4711, btw), and the backend stops itself after the default timeout of 10s is reached.

I'm not sure what the extension is getting stuck on when this situation happens. But it sounds like some default value isn't filled in, or the missing case is being handled incorrectly.

shanedell:
Set debugServer to 4711 (default) if not specified in config.

Closes #494

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -406,6 +406,10 @@ class DaffodilConfigurationProvider
       config = getConfig('Launch', 'launch', 'dfdl')
     }
 
+    if (!config.debugServer) {
+      config.debugServer = 4711
+    }
+
     if (!config.program) {
       return vscode.window
         .showInformationMessage('Cannot find a program to debug')


scholarsmate (approved):
+1, LGTM.

arosien:
I think this is a regression from https://github.com/apache/daffodil-vscode/pull/453. Tracing it...

src/adapter/activateDaffodilDebug.ts
```diff
@@ -406,6 +406,10 @@ class DaffodilConfigurationProvider
       config = getConfig('Launch', 'launch', 'dfdl')
     }
 
+    if (!config.debugServer) {
```
arosien, line 409: I need to try this, as this doesn't look like it will do anything given code like 

https://github.com/apache/daffodil-vscode/blob/main/src/daffodilDebugger.ts#LL211-L215

which doesn't even use the `serverPort`.

arosien: Ah ok, it's returning the (modified) config to the VS Code debug system, which uses the info to find the launched debug port. So if the `serverPort` is undefined, that system doesn't know to connect to some port.

arosien (approved):
👍

src/adapter/activateDaffodilDebug.ts
Shanedell: Yeah that will need update will make an issue

PR #500, opened 2023-03-20
address bar fix

### LINKED ISSUE #499: Data Editor:  Address bar needs to match the address radix
The line numbers respect the address radix, but the address bar at the top keeps its indexes in decimal.  It should match the selected address radix.

scholarsmate:
Fix the address bar so that it is sensitive to both the address radix and the byte display radix.

Fixes #499.


Shanedell (approved):
+1 👍

arosien (approved):
:+1:

PR #502, opened 2023-03-20
fix save over original file (fixes #501)

### LINKED ISSUE #501: Save is not overwriting the original file
When attempting to save over the file being edited, nothing happens.

scholarsmate:
fixes #501

### DIFF
--- a/src/omega_edit/dataEditWebView.ts
+++ b/src/omega_edit/dataEditWebView.ts
@@ -15,7 +15,14 @@
  * limitations under the License.
  */
 
+import * as omegaEditChange from 'omega-edit/change'
+import { getServerHeartbeat, IServerHeartbeat } from 'omega-edit/server'
+import * as omegaEditSession from 'omega-edit/session'
+import { CountKind } from 'omega-edit/session'
+import * as omegaEditViewport from 'omega-edit/viewport'
 import * as vscode from 'vscode'
+import { EditorMessage, MessageCommand } from '../svelte/src/utilities/message'
+import { serverPort } from './client'
 import { SvelteWebviewInitializer } from './svelteWebviewInitializer'
 import {
   dataToEncodedStr,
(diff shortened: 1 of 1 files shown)


src/omega_edit/dataEditWebView.ts
```diff
@@ -334,25 +304,17 @@ export class DataEditWebView implements vscode.Disposable {
           })
           .then(async (uri) => {
             if (uri && uri.fsPath) {
-              if (uri.path === this.fileToEdit) {
-                await this.saveSession(this.omegaSessionId, true).catch(() => {
-                  vscode.window.showErrorMessage('Failed to save')
+              await omegaEditSession
+                .saveSession(this.omegaSessionId, uri.path, true)
+                .then(async (fp) => {
+                  vscode.window.showInformationMessage(`Saved to file: ${fp}`)
```
arosien, line 310: Concern: I'll need to load this in a real editor, but I'm not sure `saveSession` returns the file path.

scholarsmate: It does.  It has to, because if you set overwrite to `false`, and give it a file path that exists, a new file path will be generated in Ωedit and returned.  You can search for `save` in the C API documentation for more (see: https://ctc-oss.github.io/omega-edit/index.html).  Also testing shows this is the case.

Shanedell:
@scholarsmate Is there supposed to be a `Save As` or something similar added? This seems like it will always overwrite the file so what if we want to make a new one? Or do you plan that being a separate PR?

scholarsmate:
@Shanedell, the `Save` button opens a file picker that allows you to save the session any way you please.  It can be saved over the file being edited, to a new file, or overwrite some other existing file.

Shanedell (approved):
+1 👍

scholarsmate:
@arosien, what else are you waiting on?

src/omega_edit/dataEditWebView.ts
arosien: Sorry, bad code navigation in Github.

arosien (approved):
:+1:

Issue #503, opened 2023-03-20
Unintended log and heartbeat sharing when more than one DE is running


scholarsmate:
Open a Data Editor on port 9000, then open another Data Editor on port 9001, and suddenly you'll notice that the heartbeats are both _saying_ that they are running on port 9001.  Also the logs are now all being written to the log file for 9001.  Both Data Editors appear to work just fine (they are communicating with their original servers), but the heartbeats and logs should be going into their respective locations.

scholarsmate:
The bug was addressed by only allowing a single Data Editor to be running at one time.  We plan to allow multiple Data Editors to share one server instance in an upcoming release.

stricklandrbls:
Closed with merge of #509

PR #509, opened 2023-03-21
Safeguard for multiple data editor instances.

stricklandrbls:
- In regards to [Issue #503](https://github.com/apache/daffodil-vscode/issues/503), implementing a sufficient solution would require a larger development effort to resolve the log & heartbeat sharing.

- This hotfix simply locks the data editor to a single instance w/ an omega edit connection to localhost:9000.


src/omega_edit/client.ts
```diff
@@ -26,6 +26,7 @@ import { DataEditWebView } from './dataEditWebView'
 import { initOmegaEditClient } from './utils'
 
 export let serverPort: number = 9000
+export let serverPortLock: boolean = false
```
scholarsmate, line 29: Instead of introducing another variable, we can use the serverPort above.  Just make it either 0, or 'undefined'.  The first instance will set it to a positive number and the second instance can check if the port number is positive, which is should be and lock the user out from creating a second DE.

scholarsmate (changes requested):

stricklandrbls:
> @stricklandrbls, if we're supporting only one DE instance, then we shouldn't even ask for a port to run the server on. It will use what is configured.

We're not prompting the user for a port, I left that function in because I figured we would eventually prompt the user for a port once issue #503 is resolved. Plus, that function only prompts the user for a port if it were undefined.

Shanedell:
I am confused why are we only allowing a single instance? Shouldn't we be able to have multiple `omega-edit` instances, as long as they use different ports? We should also allow for the user to set the port we don't want a hard coded port that is always used. What if the hard coded port is already in use and the user wants to bind to a different port so they could use editor?

stricklandrbls:
> I am confused why are we only allowing a single instance? Shouldn't we be able to have multiple `omega-edit` instances, as long as they use different ports? We should also allow for the user to set the port we don't want a hard coded port that is always used. What if the hard coded port is already in use and the user wants to bind to a different port so they could use editor?

Currently, listed as Issue #503, there are client/server implementations that cause multiple instances of the data editor, on different ports, to cause all the data editors to have the same port. Which corrupts the heartbeat validity and data editor logs.

I originally implemented locking the server port to whichever the user specified but if the user:
- Opened the data editor on port 9000
- Closed the editor
- Opened another data editor on port 9001
This would cause the heartbeats to fail again. Essentially, whichever port the user initially specifies could be the only port used, during the lifetime of the VSCode application, that is.

stricklandrbls:
> Is that because the heartbeat isn't using the proper port to get the heartbeat from? Like when you set it `9001` did you verify when you tried to get the heartbeart that the value of the `serverPort` was properly set to `9001` and didn't stay at `9000`?

Yes, I tested this. It's a mixture of too many static & widely scoped variables. I initially had the `DataEditorWebView` object create and hold its own version of Omega Edit server information, as another object, but still would cause port consolidations.

scholarsmate:
@Shanedell, no the problem is that the different Data Editors share the same instance of the Ωedit client.  The Ωedit client manages the client and log objects as singletons.  This will actually work in our favor because we can use one Ωedit server for any number of Data Editors since they can all share the same instance.  We just need to know when to start the server (this is easy, check to see if there is a server already listening) and when to shut it down (this shouldn't be hard since we can check the number of open sessions at dispose time to decide if this is the last DE (it which case it should shut it down) or not).

The server port is a configurable setting in launch.json, so if 9000 isn't allowed, it can be set to something that is.

scholarsmate:
What @stricklandrbls is attempting to do for now is just allow one DE at a time until we have the time to add the additional logic for startup and shutdown of a single shared Ωedit server.  Since the one TCP port is configured via `launch.json`, there is no need to ask for a port each time a DE is started up.  This way it will behave the same when we release the next version, but it will support more than one DE at the same time.  I imagine we can get that into v1.3.1.

src/omega_edit/client.ts
```diff
@@ -45,19 +45,21 @@ async function getOmegaEditPort(port: number | undefined = undefined) {
...
-      value: defaultServerPort.toString(),
-    })
 
-    if (portEntered) {
-      serverPort = parseInt(portEntered)
-    } else {
-      serverPort = 0
-      throw Error('Bad port entered')
-    }
+    serverPort = defaultServerPort
```
Shanedell, line 49: Update function `getOmegaEditPort` to be

```ts
async function getOmegaEditPort() {
  const defaultServerPort = vscode.workspace
    .getConfiguration('dataEditor')
    .get<number>('serverPort', serverPort)

  serverPort = defaultServerPort
}

```

This will always assign a port.

src/omega_edit/client.ts
```diff
@@ -145,7 +152,7 @@ export function activate(ctx: vscode.ExtensionContext) {
         port: number | undefined = undefined,
```
Shanedell, line 152: Remove this param

src/omega_edit/client.ts
```diff
@@ -145,7 +152,7 @@ export function activate(ctx: vscode.ExtensionContext) {
...
         fileToEdit: string = ''
       ) => {
-        await getOmegaEditPort(port)
+        await getOmegaEditPort(serverPort)
```
Shanedell, line 155: See previous comment to remove the port param from the `getOmegaEditPort` function since its not needed anymore.

src/omega_edit/client.ts
```diff
@@ -45,19 +45,21 @@ async function getOmegaEditPort(port: number | undefined = undefined) {
     const defaultServerPort = vscode.workspace
       .getConfiguration('dataEditor')
       .get<number>('defaultServerPort', serverPort)
```
Shanedell, line 47: Can we rename `defaultServerPort` to `serverPort`. Since it can be updated I don't think it should have `default` in the name as that makes me thing this value shouldn't be set.

Shanedell (changes requested):

src/omega_edit/client.ts
```diff
@@ -36,29 +37,21 @@ export function getOmegaEditPackageVersion(filePath: fs.PathLike) {
...
-      .get<number>('defaultServerPort', serverPort)
-    const portEntered = await vscode.window.showInputBox({
-      prompt: 'Enter port number to run omega-edit server on',
-      value: defaultServerPort.toString(),
-    })
-
-    if (portEntered) {
-      serverPort = parseInt(portEntered)
-    } else {
+      .get<number>('defaultServerPort', defaultServerPort)
```
Shanedell, line 48: I think this variable in `dataEditor` config should be changed to `serverPort`.

So in `package.json` I believe `dataEditor.defaultServerPort` needs to be changed to `dataEditor.serverPort`. Then here it should do `.get<number>('serverPort', defaultServerPort)`. The configuration variable having `default` in the name makes me think it should not be set.

Shanedell:
@stricklandrbls Also update `src/tests/suite/dataEditor.test.ts`, it needs to remove the passing of the port to the `extension.data.edit` command. That I believe is why the CI is failing. However, locally I am still getting an error locally about the heartbeat. Not sure if you are still seeing that as well?

Shanedell (approved):
+1 👍

PR #510, opened 2023-03-21
Tab name to be the filepath of the file being edited

### LINKED ISSUE #507: Name data editor tab according to the file name
Calling the tab "Data Editor" is confusing. All editors I know of put the filename in the tab, with the full path in the hover text. I imagine VS Code probably has some helpers for all of this.

scholarsmate:
The panel object has a title property that is writable, but I don't see a way to do that and hover text.

Here is what it looks like if we use the full path:

Screenshot 2023-03-21 at 4 43 47 PM

I chose to use the base name of the file, though it's not unique if we open say 2 different `package.json` files, then we have 2 tabs named the same thing even though the paths are different.  Here is what it looks like with this patch:

Screenshot 2023-03-21 at 4 49 18 PM

Fixes #507.

### DIFF
--- a/src/omega_edit/dataEditWebView.ts
+++ b/src/omega_edit/dataEditWebView.ts
@@ -34,6 +34,7 @@ import {
   setViewportDataForPanel,
   viewportSubscribe,
 } from './utils'
+import path from 'path'
 
 const VIEWPORT_CAPACITY_MAX = 1000000 // Maximum viewport size in Ωedit is 1048576 (1024 * 1024)
 const HEARTBEAT_INTERVAL_MS = 1000 // 1 second (1000 ms)
@@ -96,6 +97,7 @@ export class DataEditWebView implements vscode.Disposable {
         .then(async (fileUri) => {
           if (fileUri && fileUri[0]) {
             this.fileToEdit = fileUri[0].fsPath
+            this.panel.title = path.basename(this.fileToEdit)
             await this.setupDataEditor()
           }
         })


scholarsmate:
It would be great if we can figure out how to make the hover text use the full path name.

src/omega_edit/client.ts
```diff
@@ -142,8 +142,7 @@ export function activate(ctx: vscode.ExtensionContext) {
       'extension.data.edit',
       async (
         startServ: boolean = true,
-        port: number | undefined = undefined,
-        fileToEdit: string = ''
```
Shanedell, line 146: You will need to revert all these changes that deal with `fileToEdit`.

The code involving `fileToEdit` is essential for the API to function properly for the tests. With this change during the tests when the data editor is initially opened it will prompt to select a file which is not the desired behavior during testing. The test still currently passes since it just checks the webview is opened, which it is behind the file picking prompt. In future tests for the data editor we will want to use the content that is inside of the data editor. To do this we need to be able to specify the file we wish to edit so that the editor will contain content and not get stuck on a file prompt.

Shanedell (changes requested):

src/omega_edit/client.ts
scholarsmate: I'm not happy with having this testing cruft in the shipped project.  I've reverted it _temporarily_ and will revisit later.

Shanedell (approved):
+1 👍

scholarsmate:
@arosien, will you please review this?  Thanks.

PR #511, opened 2023-03-22
move error msg and add instrux in edit vpt in single byte mode

scholarsmate:
Relocates the error message in single byte mode closer to where the error is taking place.  Also adds an instruction message in the edit viewport in single byte mode.  Here is the screenshot with this patch applied:

Screenshot 2023-03-22 at 9 15 24 AM


scholarsmate:
@arosien, will you please review this?  Thanks.

arosien (approved):
👍

PR #515, opened 2023-03-22
Implement management for selected data.

### LINKED ISSUE #512: Clear selection and disable Replace when a selection is made in multiple byte edit mode
There can be some unexpected surprises when a user is in multiple byte editing mode, they have a selection, and they execute a Replace operation.  When they commit their selection, it might not be what and where they expected.

To solve this we need to do two things.  The first is that we need to disable the `Replace` button when a selection is _active_.  The second is that we need an intuitive way to _clear_ the active selection (which is different than clearing the content in a selection -- this is a commitable `delete`).

For the selection clear operation, I recommend a little `[X]` button next to the `Selection ...` text in the `Edit` viewport.  Click the `[X]`, clear the selection.  If the user has made changes to the selection, give them an `Are you sure...` modal to confirm.

stricklandrbls:
- Notation added while in single edit byte mode to notify that user of the main editor selection area being disabled and to press ESC to close the single edit byte window, due to the `X` button in that window being `delete` byte instead of close

- Svelte store added for simple lookup of whether selection data is populated even with no content within the editor area. (Multiple byte edit mode).

- Button added to clear the selection data

Closes #512

### DIFF
--- a/src/svelte/src/components/dataEditor.svelte
+++ b/src/svelte/src/components/dataEditor.svelte
@@ -76,6 +76,7 @@ limitations under the License.
     headerHidden,
     originalDataSegment,
     editedByteIsOriginalByte,
+    selectionActive,
   } from '../stores'
   import {
     radixOpt,
@@ -411,6 +412,8 @@ limitations under the License.
       selected.id === 'logical'
         ? Math.floor(selectionEnd / 2)
         : selectionOffsetsByRadix[$displayRadix].end
+
+    $selectionActive = true
   }
 
   /**
@@ -661,6 +664,7 @@ limitations under the License.
     $selectionOriginalEnd = 0
     $cursorPos = 0
     $editorSelection = ''
+    $selectionActive = false
     $editedDataSegment = new Uint8Array(0)
   }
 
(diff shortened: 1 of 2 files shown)


src/svelte/src/components/dataEditor.svelte
```diff
@@ -1662,10 +1684,24 @@ limitations under the License.
     display: flex;
   }
 
+  .dataEditor div.measure.selection {
+    flex-direction: column;
+    text-align: center;
+    justify-content: center;
```
scholarsmate, line 1690: Super minor, but I think ergonomically, speaking the selection text and clear button should be left justified and not centered because it's easier to get to that clear button if it's at a fixed location. Centering it makes it move around.

scholarsmate (changes requested):

src/svelte/src/components/dataEditor.svelte
stricklandrbls: Just pushed an update. Take a look and let me know what you think.

scholarsmate (approved):

scholarsmate (changes requested):
Something isn't working right.  If I select a segment, then delete it and commit it, then I'm in a state where I can't get the Replace button to enable again.  See screenshot:

Screenshot 2023-03-23 at 8 06 26 AM

PR #524, opened 2023-03-23
Removed Duplicate Infotext & Reverted Editor div to textarea

### LINKED ISSUE #519: Edit View to use HTML textarea instead of editable div
Since we don't need to highlight bytes in the Edit View from the Data View (which is why we selected an editable `div` instead of a `textarea`), we should use a `textarea` because it preserves things like whitespace and newlines.  It also sets us up later to replace the `textarea` with something more feature rich, like CodeMirror, without too much trouble.  If there are multiple adjacent whitespace, we get a situation like this with an editable div:

Screenshot 2023-03-23 at 1 47 14 PM

The multiple adjacent whitespace collapses into a single space.  Remove that one space removes them all.

### LINKED ISSUE #523: Remove Duplicate Info Text
There seems to have been a merge overlap between commits: https://github.com/apache/daffodil-vscode/commit/4a0105ffdfdf4711c00eae7cef8cb2d7ee82fb80 & https://github.com/apache/daffodil-vscode/commit/85ea375eb75953bc4737a6056318bc67d4b5d736. Two information texts now appear. 

The bottom info text needs to be removed.

### Screenshot
image

stricklandrbls:
Removed duplicate infotext & changed edit panel.

    - Removed the duplicate information texts in regards to the editing
      windows for each byte edit mode.
    - Reverted the editor area from `div` to `textarea` to properly render
      whitespace characters.

Closes #523 
Closes #519

### DIFF
--- a/src/svelte/src/components/dataEditor.svelte
+++ b/src/svelte/src/components/dataEditor.svelte
@@ -115,7 +115,7 @@ limitations under the License.
   let isScrolledToEnd = false
 
   const selectedContent = writable(
-    document.getElementById('selectedContent') as HTMLDivElement
+    document.getElementById('selectedContent') as HTMLTextAreaElement
   )
   const editByteWindow = writable(
     document.getElementById('editByteWindow') as HTMLDivElement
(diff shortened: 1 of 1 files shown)


Shanedell (approved):
+1 👍

PR #525, opened 2023-03-24
    add logic to handle tag attributes split on multiple lines

### LINKED ISSUE #461: Incorrect intellisense suggestions triggered at the schema closing tag
Incorrect intellisense suggestions triggered at the schema closing tag.  When the cursor is positioned at the beginning of the schema closing tag '</xs:schema> and control space is entered, the intellisense results are generic alphabetic results (as of PR #458) instead of the results that should be supplied by the element completion function.

### LINKED ISSUE #473: Intellisense for the assert tag doesn't supply any attributes.

### LINKED ISSUE #474: Multi-line tags not being handled properly by Intellisense
There are two issues with how Intellisense is handling multi-line tags.

The first issue is with the indentation of the filled in lines. For example, when having intellisense fill in the schema tag, the user is required to provide a closing angle bracket. When they do, an empty newline and a line with a closing tag are inserted into the document. The indentation of these pieces is determined by the line where the closing angle bracket was placed, and not by the indentation of the opening angle bracket for the multi-line tag. This is the reason that the elementFormDefault property is left-aligned in the schema tag's intellisense autocompletation

image

Expected view:
image

The second issue is with non-schema tags where the user puts properties on multiple lines. When the closing tag is supplied, the autocompletion will back up 2 positions and place the closing tag there. However, this will insert the closing tag inside of a property value.

image

Expected view:
image

### LINKED ISSUE #493: Multi-line self closing tag shows as an open tag
The self-closing element tags in this snippet show as open tags with a close tag to the "checkMissingCloseTag" function. This can result in a closing tag for this item showing up at the cursor position if a '>' symbol is typed at the end of an existing line.  This code from gif.dfdl.xsd can cause an element closing tag to be generated at the cursor position if a '>' is typed at the end of a previous line.
`<xs:element name="Header">
	<xs:complexType>
		<xs:sequence>
			
			
			<xs:sequence dfdl:hiddenGroupRef="hidden_GIF_Signature_Group" />
			<xs:element name='Signature' type='xs:string' dfdl:inputValueCalc='{
					if (xs:string(../Hidden_Signature) eq "474946") then "GIF"    
					else fn:error("gif", "fn:error called.", "Header")
					}' />
			
			<xs:sequence dfdl:hiddenGroupRef="hidden_GIF_Version_Group"/>
			<xs:element name="Version" type="xs:string" dfdl:inputValueCalc='{
					if (xs:string(../Hidden_Version) eq "383961") then "89a"
					else if (xs:string(../Hidden_Version) eq "383761") then "87a"    
					else fn:error("gif", "fn:error called.", "Header")
					}'/>
		</xs:sequence>
	</xs:complexType>
</xs:element>
`

rt320:
- Fixed the logic for determining self-closing tags when '/>' is
        on a separate line
    - Added missing assert attribute choices
    - Corrected a problem with incorrect intellisense suggestions at
      the schema closing tag
    - Fixed suggestions before and after a multi-line self-close tag
    - Fixed suggestions inside a multi-line self-close tag
    - Removed commented code
    - Fixed issue with ctrl+space inside a closing tag

    Closes #461 
    Closes #473 
    Closes #474
    Closes #493


arosien (approved):
👍

PR #529, opened 2023-03-24
Reordered OE server kill & added verbose vars

### LINKED ISSUE #528: Extension "Port Required" on Data Editor Activation Command
### Issue
When activating the Data Editor the `serverPort` variable, that pulls from the `package.json` configuration settings, becomes unset while starting up the server connection to Omega Edit.

This seems to be related to lingering omega edit server processes due to a failure to dispose of a nonexistent `DataEditorWebview`.

### Current Workaround
Simply run the Data Editor activation command again and the panel will appear as normal.

---

**Failure Dialog**
image

**Invalid Port on Omega Edit Server Init**
port-0

stricklandrbls:
- The Omega Edit server kill on DataEditWebviewPanel disposal now nullifies port
- Variables related to Omega Edit server connection were changed to be more verbose/specific, in the chance of potentially additional "serverPorts" that may be needed in the future.

Closes #528

### DIFF
--- a/src/omega_edit/client.ts
+++ b/src/omega_edit/client.ts
@@ -25,8 +25,11 @@ import XDGAppPaths from 'xdg-app-paths'
 import { DataEditWebView } from './dataEditWebView'
 import { initOmegaEditClient } from './utils'
 
-const defaultServerPort: number = 9000
-export let serverPort: number = 0
+const DEFAULT_OMEGA_EDIT_PORT: number = 9000
+const OMEGA_EDIT_MIN_PORT: number = 1024
+const OMEGA_EDIT_MAX_PORT: number = 65535
+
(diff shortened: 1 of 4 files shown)


src/omega_edit/client.ts
```diff
@@ -113,17 +121,19 @@ async function serverStart(ctx, version) {
     ++animationFrame
   }, animationInterval)
 
+  // if (oemgaEditServerPort === 0) throw 'Server port was never set'
```
Shanedell, line 124: remove this

stricklandrbls: Had to step out for a house showing, should be home to make this change soon.

Shanedell (approved):
+1 👍

PR #534, opened 2023-03-27
Updated byte symbol for undefinedCharStandIn

### LINKED ISSUE #531: Consecutive 'undefinedCharStandIn' Symbol Display Issues
## Issue
When consecutive _undisplayable_ byte values render in the logical display viewport[^1], the viewport will display these bytes inconsistently. 

A binary test file will be attached to this issue for anyone looking to recreate this exact scenario. 
[bin-data.zip](https://github.com/apache/daffodil-vscode/files/11080209/bin-data.zip)

**Note:** The `.zip` file extension is irrelevant and was only added to satisfy GitHub's attachment upload requirements. 

### Viewports Screenshot
image
The last byte on the top line has the value of `0x09`.[^2]

[^1]: The Logical Viewport is text encoded as Latin-1.
[^2]: The symbol used for _undisplayable_ byte data may not actually be monospace. ~~Another issue may need to be created for this.~~

#### Footnote[^2] Test Update
Changing the � symbol to an '_' resolves this issue except for in the typical _problem child_ data byte of `0x0A`, which has been documented to cause some rendering and editing issues.

## Issue Resolution
The symbol for the `undefinedCharStandIn` needs to be changed to a _monospacable_ symbol.

stricklandrbls:
- The current committed symbol is not a monospace symbol which was causing logical viewport display issues.

Closes #531

### DIFF
--- a/src/omega_edit/utils.ts
+++ b/src/omega_edit/utils.ts
@@ -72,7 +72,11 @@ export async function viewportSubscribe(
     })
     .on('error', (err) => {
       // Call cancelled thrown sometimes when server is shutdown
-      if (!err.message.includes('Call cancelled')) throw err
+      if (
+        !err.message.includes('Call cancelled') &&
+        !err.message.includes('UNAVAILABLE')
+      )
+        throw err
     })
 }
 
@@ -94,7 +98,7 @@ export function logicalDisplay(
   bytes: ArrayBuffer,
   bytesPerRow: number
 ): string {
-  const undefinedCharStandIn = '�'
+  const undefinedCharStandIn = 9617
   let result = ''
   if (bytes.byteLength > 0) {
     // TODO: How does this affect the simple editor?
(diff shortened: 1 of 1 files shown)


scholarsmate (approved):
+1, 👍

src/omega_edit/utils.ts
```diff
@@ -94,7 +94,7 @@ export function logicalDisplay(
   bytes: ArrayBuffer,
   bytesPerRow: number
 ): string {
-  const undefinedCharStandIn = '�'
+  const undefinedCharStandIn = '░'
```
Shanedell, line 97: Is there a way to use an ascii code that when converted to a string it will give this value? Just thinking if we used a ascii code or something then convert it to a string if it produces the same result it would look cleaner that having this weird symbol in there.

Shanedell: @stricklandrbls Can we instead change this variable to be `const undefinedCharCodeStandIn = 9617`? Then when you go to use it do `String.fromCharCode(undefinedCharCodeStandIn)`? I was able to get this char code value by doing `console.log('░'.charCodeAt(0))` in a nodejs shell

stricklandrbls: Sure thing

Shanedell (approved):
+1 👍

PR #535, opened 2023-03-27
Pin package rollup-plugin-multi-input to 1.3.2

Shanedell:


NolanMatt (approved):
👍 But in grey

scholarsmate:
Rollup-plugin-multi-input version 1.3.3 is a development dependency that required node 16+, so we pinned it at 1.3.2 so that node 14 builds will still work.

Shanedell:
@arosien Can you give this a look over when you have a chance? This is causing all of the current PRs to fail CI because as Davin mentioned it is trying to install a newer version that required nodejs 16.

arosien (approved):
👍

PR #539, opened 2023-03-29
Build schema issue

### LINKED ISSUE #521: Backend builds sometimes fail with duplicate generated schemas
Sometimes [jobs fail](https://github.com/apache/daffodil-vscode/actions/runs/4500462875/jobs/7919625386) like:
```
$ sbt test && node ./out/tests/runTest.js
[info] welcome to sbt 1.8.2 (Temurin Java 1.8.0_362)
[info] loading settings for project daffodil-vscode-build from plugins.sbt ...
[info] loading project definition from /Users/runner/work/daffodil-vscode/daffodil-vscode/project
[info] loading settings for project daffodil-debugger from build.sbt ...
[info] set current project to daffodil-debugger (in build file:/Users/runner/work/daffodil-vscode/daffodil-vscode/)
[info] Compiling 1 XSD file(s) to /Users/runner/work/daffodil-vscode/daffodil-vscode/server/core/target/scala-2.12/src_managed_cxf
parsing a schema...
Error:  'DFDLExpression' is already defined
  line 38 of file:/Users/runner/work/daffodil-vscode/daffodil-vscode/server/core/target/scala-2.12/resource_managed/xsd/org/apache/daffodil/xsd/DFDL_part1_simpletypes.xsd
...
```

Workaround was committed as #508.

More info:
- XSD schemas are extracted from the daffodil dependencies and used to generate Java sources via JAXB. 
- `sbt test` seems to non-deterministically regenerate the Java sources. If you run it 3 times starting from a clean repo, the sources are generated 2 times, so some dependency task is being invalidated and thus rerun.
- I believe CI disk state is persisted across builds.

cc: @scholarsmate @Shanedell

NolanMatt:
Should hopefully fix #521.  took Adam's suggestion and removed some lines that could cause test to duplicate artifacts from compile.  Also with this don't need to blow away target directory before package, which now comes after test.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -117,5 +117,4 @@ jobs:
         if: runner.os != 'Linux'
       - name: Package
         run: |
-          rm -rf **/target
           yarn package
--- a/build.sbt
+++ b/build.sbt
@@ -177,7 +177,7 @@ lazy val xjcSettings =
       //   foreach throws an error when the type of resources is a Seq[File] rather than an Array[File]
       moved_resources toSeq
     }.taskValue,
-    Test / xjc / sources := (Compile / xjc / sources).value,
-    Test / doc / sources := (Compile / doc / sources).value,
-    Test / sourceGenerators := (Compile / sourceGenerators).value
+//    Test / xjc / sources := (Compile / xjc / sources).value,
+//   Test / doc / sources := (Compile / doc / sources).value,
+//    Test / sourceGenerators := (Compile / sourceGenerators).value
   )
--- a/build.sbt
+++ b/build.sbt
@@ -20,6 +20,7 @@ import Classpaths.managedJars
 
 //Fixes build issues on java11+
 run / fork := true
+Global / lintUnusedKeysOnLoad := false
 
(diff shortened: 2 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -117,5 +117,4 @@ jobs:
         if: runner.os != 'Linux'
       - name: Package
         run: |
```
Shanedell, line 119: Can you update this from

```yml
run: |
  yarn package
```

to be
```yml
run: yarn package
```

Other than that I like this PR. Once this is updated I am willing to approve!

NolanMatt:
Going to use a new clean branch that doesn't have a bunch of merge commits in it so it is much easier to do a 1 commit PR.

Shanedell:
You could have still used this one and have squashed the commits, that is the normal workflow.

build.sbt
```diff
@@ -20,6 +20,7 @@ import Classpaths.managedJars
 
 //Fixes build issues on java11+
 run / fork := true
+Global / lintUnusedKeysOnLoad := false
```
arosien, line 23:
```suggestion
```
This isn't needed to fix the CI issue, so let's make a separate PR.

PR #540, opened 2023-03-29
Update build.sbt to not duplicate compile/test aritfacts and get rid …

NolanMatt:
…of unnecessary delete statements in CI

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -116,6 +116,4 @@ jobs:
       - run: yarn test
         if: runner.os != 'Linux'
       - name: Package
-        run: |
-          rm -rf **/target
-          yarn package
+        run: yarn package
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ import Classpaths.managedJars
 
 //Fixes build issues on java11+
 run / fork := true
-
+Global / lintUnusedKeysOnLoad := false
 val packageJsonStr = scala.io.Source.fromFile("package.json").mkString
 
 val daffodilVer = {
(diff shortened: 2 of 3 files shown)


NolanMatt:
Funny, it actually failed for the first time on this PR. I guess something is still up.

Shanedell:
You will want to have a shorter commit message or take what you have and break into multiple lines. You will also need to add all issues to commit message by adding `Closes #ISSUE_NUMBER` on a separate line for each issues. Here is a sample commit from what you now:

```bash
Update build.sbt and CI.yml:

- Stop duplicating compile/test artifacts.
- Get rid of unnecessary delete statements in CI.

Closes #521 
```

scholarsmate:
@arosien, @NolanMatt is having a difficult time investigating the problem since it works for him locally and intermittently in CI.  He indicated that you are able to reproduce issue locally, so I'll reassign it to you.

arosien:
I really don't understand what's going on here. Since the original message was referring to schema files that existed in a path that were only meant to be a temporary location, and later moved, I removed the need to move any extracted schema files at all. However, this still failed the build--only under Node 16--with a timeout during the headless extension tests, which doesn't seem to have anything to do with TDML.

Running the same tests locally, including using Node 16, passes for me.

Rerunning the failed job makes it pass, but that doesn't reassure me. Trying another complete rerun to see if I can trigger any failures.

arosien:
@NolanMatt I can't force push to your fork, so you'll have to squash everything to get to a single commit. (or give me perms somehow)

Shanedell (approved):
+1 👍

arosien (approved):
👍

PR #559, opened 2023-04-03
Data Editor UI has togglable color mode.

### LINKED ISSUE #536: Display themes don't change Data Editor background colors
Light themes are especially hard to read.

3acb2279-0784-4a5e-ad15-bf3e01177ab8

stricklandrbls:
- UI color theme pulls from user selected VSCode color theme.
- Some UI components have locked theme background to ensure readability across various themes. 

Closes #536

### DIFF
--- a/src/svelte/src/components/dataEditor.svelte
+++ b/src/svelte/src/components/dataEditor.svelte
@@ -95,6 +95,11 @@ limitations under the License.
     setSelectionOffsetInfo,
     radixBytePad,
   } from '../utilities/display'
+  import {
+    CSSThemeClass,
+    UIThemeCSSClass,
+    darkUITheme,
+  } from '../utilities/colorScheme'
   import { vscode } from '../utilities/vscode'
   import { MessageCommand } from '../utilities/message'
   import { writable } from 'svelte/store'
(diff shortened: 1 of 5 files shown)


src/svelte/src/components/dataEditor.svelte
```diff
@@ -774,658 +781,713 @@ limitations under the License.
...
-          >
-          <div class="col-item" id="disk_file_size">{$diskFileSize}</div>
-        </div>
-        <div class="two-row-item flex-container col">
-          <label class="col-item file-metrics" for="computed_file_size"
-            >Computed Size</label
-          >
-          <div class="col-item" id="computed_file_size">
-            {$computedFileSize}
+<body class={$UIThemeCSSClass}>
```
scholarsmate, line 792: Why use a `body` tag?  This is going to be injected into VS Code markup which will already have a body.

scholarsmate (commented):
This update looks pretty nice.  Just a concern about using a `body` tag.

src/svelte/src/components/dataEditor.svelte
stricklandrbls: I needed a way to set the global font colors per scheme but didn't have a way to inject any styling in the `SvelteWebviewInitializer` that contains the UI body. This was the quickest solution that I predict will change with the implementation of https://github.com/ctc-oss/daffodil-vscode/issues/110

Shanedell (approved):
+1 👍

PR #566, opened 2023-04-05
upgrade to omega-edit 0.9.63

### LINKED ISSUE #567: Upgrade to Ωedit 0.9.63
0.9.63 has a new package structure and the code is now packed and there is now an index that serves as the client entrypoint.  It also includes enhanced server management and an option to plumb a server-side logback configuration file.

scholarsmate:
Upgrades Ωedit to 0.9.63, and now includes server-side logging, log rotation, heartbeat enhancements, and defensive programming.

closes #567


Shanedell:
@scholarsmate I think you need to update how the tests try to run the server

package.json
```diff
@@ -42,10 +42,10 @@
     "sbt": "sbt universal:packageBin"
   },
   "dependencies": {
+    "@omega-edit/client": "^0.9.59",
```
Shanedell, line 45: Make sure to update this to be `0.9.59` instead of `^0.9.59`. It is preferred to pin packages at a specific version. Having `^` basically allows the install to the version listed or anything newer.

Shanedell: @scholarsmate This will still apply just with `0.9.60`

scholarsmate: Ah, good point!  Fixing...

Shanedell:
@scholarsmate Also update `src/tests/runTest.ts` to catch the output of `runTests` and use that value in `process.exit` something like this

```ts
const runTestsResult = await runTests({
  vscodeExecutablePath,
  extensionDevelopmentPath,
  extensionTestsPath,
})
process.exit(runTestsResult)
```

This is needed because it seems `yarn test` will sometimes hang without it.

Shanedell:
@scholarsmate 0.9.62 was released and finally all `omega-edit-native` jars have their library files. So fingers crossed once you bump to that this CI will pass without much of an issue.

Shanedell (approved):
+1 👍

Shanedell:
@scholarsmate You may need to rebase.

scholarsmate:
@arosien, please review.  Thanks!

arosien (approved):
👍

PR #570, opened 2023-04-05
Clear selection data when ephemeral windows close

### LINKED ISSUE #561: Close Edit window when Offset field or Search used to change location
Editing a larger file (i.e., package-lock.json) in single mode, click on a byte in the Physical window.
The Edit widget appears with value.
Click in the offset field, and enter an address a long way way - like 0x2002.

The Physical window changes as expected, but the selected byte remains on the screen with the contents of the originally selected byte still active - with nothing indicating what byte is actively being edited. 
This is also an issue with prev/next searches.

The Edit widget should be closed or adjusted to the location & value of the specified offset.

stricklandrbls:
- Lingering selection data, especially in SBM, would cause issues when trying to search / replace.

Closes #561

### DIFF
--- a/src/svelte/src/components/dataEditor.svelte
+++ b/src/svelte/src/components/dataEditor.svelte
@@ -220,6 +220,7 @@ limitations under the License.
         (targetLine == 0 ? 0 : targetLine - 1) * lineHeight
     }
     closeEphemeralWindows()
+    clearDataDisplays()
   }
 
   function goToEventHandler(_: Event) {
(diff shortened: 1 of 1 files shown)


src/svelte/src/components/dataEditor.svelte
```diff
@@ -220,6 +220,7 @@ limitations under the License.
         (targetLine == 0 ? 0 : targetLine - 1) * lineHeight
     }
     closeEphemeralWindows()
+    clearDataDisplays()
```
scholarsmate, line 223: Is there any time where we'd close the ephemeral window and not clear the data display?  If not, let's just clear the data display ni the close ephemeral window function instead to make it simpler and less error prone.

scholarsmate (approved):
+1, 👍

Shanedell (approved):
+1 👍

PR #590, opened 2023-04-25
omega-edit release fix

Shanedell:
Was preparing a release but testing locally I found the omega-edit data editor did not run when the extension was installed via VSIX. This update fixes that issue as the server binary could not be found when installed via VSIX because it was not bundled into the VSIX.


scholarsmate (approved):
+1, 👍

PR #618, opened 2023-05-03
Fix hex view duplication issue

### LINKED ISSUE #616: 1.3.0-rc1 - Hex view duplication issue
The hex view seems to not remake `datafile-hex` each time it is opened. It instead seems to append the data file hex into the file if it exists already. Seems like on close it also adding more data. So the hex view needs to be updated to reduce all the duplication it is causing.

Shanedell:
Fix hex view duplication issue

- Issue was caused by variable not being set to empty string on each new debug session being started.

Closes #616

### DIFF
--- a/src/hexView.ts
+++ b/src/hexView.ts
@@ -140,6 +140,7 @@ export class DebuggerHexView {
       this.closeHexFile()
       this.deleteFile(this.hexFile)
       this.decorator.dispose()
+      this.hexString = '' // make sure no data from previous runs is carried over
     }
   }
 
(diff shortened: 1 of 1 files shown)


scholarsmate (approved):
+1, 👍

Shanedell:
@stevedlawrence @arosien @mbeckerle Would one of you guys be able to give this look over. Believe once this PR is merged we can try to get an rc2 out for testing.

arosien (approved):
👍

PR #620, opened 2023-05-03
Launch wizard updates and rename data editor vars

### LINKED ISSUE #617: Add Ωedit server port setting to the launch wizard
The Ωedit server port is not currently configurable in the launch setting wizard, but should be.

### LINKED ISSUE #619: Launch Wizard Broken
Looks like with some of the updates for TDML configs the launch wizard was broken. Some of the styling also does not match up.

### LINKED ISSUE #621: Rename dataEditor.* config variables
To be consistent the `dataEditor.*` variables should be renamed.

eg

```json
{
    "dataEditor.omegaEditPort": 9000
    "dataEditor.logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
    "dataEditor.logLevel": "info"
}
```

should become something like

```json
{
    "dataEditor": {
        "omegaEditPort": 9000
        "logLevel": "info"
    }
}
```

Shanedell:
Launch wizard updates and rename data editor vars:

- Rename variables dataEditor.* to be under an object called dataEditorConfig.
  - The variables that can be inside of that object are omegaEditPort, logFile and logLevel
- Add missing config keys and values to parsed configs.
  - If not done this causes the launch wizard to not open.
- Fix a issue with some TDML settings not showing up properly.
- Update TDML config settings styling to better match the rest of the launch wizard.
  - When these items are hidden, the height and width are set to 0 so no large black space would be created in the UI.
- Add support for setting data editor settings. These were omegaEditPort, logFile and logLevel.
- Don't show TDML name field when action is set to "none".

Closes #617
Closes #619
Closes #621

### DIFF
--- a/package.json
+++ b/package.json
@@ -343,7 +343,7 @@
               },
               "infosetOutput": {
                 "type": "object",
-                "description": "Destination for final Infoset (file-path | 'console' | 'none')",
+                "description": "Destination for final Infoset ('file' | 'console' | 'none')",
                 "default": {
                   "type": "console",
                   "path": "${workspaceFolder}/infoset.xml"
@@ -413,27 +413,14 @@
                 },
                 "default": {}
               },
-              "dataEditor.omegaEditPort": {
-                "type": "integer",
-                "description": "Editor server default port",
-                "default": 9000
-              },
-              "dataEditor.logFile": {
-                "type": "string",
-                "description": "Path to log file for data editor",
-                "default": "${workspaceFolder}/dataEditor-${omegaEditPort}.log"
-              },
-              "dataEditor.logLevel": {
-                "type": "string",
-                "description": "Log level for data editor",
-                "enum": [
-                  "error",
-                  "warn",
-                  "info",
-                  "debug",
-                  "trace"
-                ],
-                "default": "info"
+              "dataEditor": {
+                "type": "object",
+                "description": "Configuration for Data Editor. Settings are port, logFile and logLevel",
+                "default": {
+                  "port": 9000,
+                  "logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
+                  "logLevel": "info"
+                }
               }
             }
           }
@@ -464,9 +451,11 @@
             "daffodilDebugClasspath": "",
             "variables": {},
             "tunables": {},
-            "dataEditor.omegaEditPort": 9000,
-            "dataEditor.logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
-            "dataEditor.logLevel": "info"
+            "dataEditor": {
+              "port": 9000,
+              "logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
+              "logLevel": "info"
+            }
           }
         ],
         "configurationSnippets": [
@@ -498,9 +487,11 @@
               "daffodilDebugClasspath": "",
               "variables": {},
               "tunables": {},
-              "dataEditor.omegaEditPort": 9000,
-              "dataEditor.logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
-              "dataEditor.logLevel": "info"
+              "dataEditor": {
+                "port": 9000,
+                "logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
+                "logLevel": "info"
+              }
             }
           }
         ],
@@ -560,7 +551,7 @@
           },
           "infosetOutputType": {
             "type": "string",
-            "description": "Destination for final Infoset (file | 'console' | 'none')",
+            "description": "Destination for final Infoset ('file' | 'console' | 'none')",
             "enum": [
               "file",
               "console",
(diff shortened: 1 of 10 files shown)


src/utils.ts
```diff
@@ -147,6 +150,15 @@ export function getConfig(
     daffodilDebugClasspath: daffodilDebugClasspath
       ? daffodilDebugClasspath
       : defaultConf.get('daffodilDebugClasspath', ''),
+    dataEditorOmegaEditPort: dataEditorOmegaEditPort
+      ? dataEditorOmegaEditPort
+      : defaultConf.get('dataEditor.omegaEditPort', 9000),
```
scholarsmate, line 155: Can't we pull the default port from package.json?

Shanedell: That is what this trying to do. The second value is the default value if nothing is found

package.json
```diff
@@ -627,17 +618,17 @@
             },
             "default": {}
           },
-          "dataEditor.omegaEditPort": {
+          "dataEditorOmegaEditPort": {
```
scholarsmate, line 621: Should this just be `omegaEditPort`?

package.json
```diff
@@ -627,17 +618,17 @@
...
             "type": "integer",
             "description": "Editor server default port",
             "default": 9000
           },
-          "dataEditor.logFile": {
+          "dataEditorLogFile": {
```
scholarsmate, line 626: Should this just be `logFile`?

package.json
```diff
@@ -627,17 +618,17 @@
...
             "type": "string",
             "description": "Path to log file for data editor",
             "default": "${workspaceFolder}/dataEditor-${omegaEditPort}.log"
           },
-          "dataEditor.logLevel": {
+          "dataEditorLogLevel": {
```
scholarsmate, line 631: Should this just be `logLevel`?

package.json
Shanedell: I believe they should all keep `dataEditor` so we are aware of what they are for. These ones are used for getting default values.

src/utils.ts
scholarsmate: Shouldn't something _always_ be specified for a default server port (debug, data, or otherwise) in `package.json`.  If not, that's a bug in my opinion and the code should error out indicating that no default value was specified in `package.json` when it should have been.

scholarsmate: Sometimes you want a failsafe (e.g., PID files to help recover from an unexpected crash), and other times you want to _reveal_ programming errors (e.g., asserts).  In this case, I'd prefer the latter.

Shanedell: I can see where you are coming from. But I believe maybe we should ask others for their input as well before deciding what to change to and do from now on. @stevedlawrence @mbeckerle @rt320 @arosien @stricklandrbls @michael-hoke @NolanMatt @nlewis05  what are your thoughts?

stevedlawrence: I don't think I have a strong opinion one way or the other in general--there's good arguments for both sides. It's probably easier to detect typos or configuration name changes if you don't do defaults, so I'd maybe lean towards no default.

That said, in this particular case it looks like all uses of `defaultConf.get(...)` provide a default value, so at least in this PR I suggest we stay consistent with that and provide a default value.

scholarsmate:
@Shanedell, make sure to also document the changes here https://github.com/apache/daffodil-vscode/wiki/Apache-Daffodil™-Extension-for-Visual-Studio-Code:-v1.3.0.

scholarsmate (approved):
+1, 👍

package.json
```diff
@@ -413,27 +413,14 @@
...
-                "description": "Log level for data editor",
-                "enum": [
-                  "error",
-                  "warn",
-                  "info",
-                  "debug",
-                  "trace"
-                ],
-                "default": "info"
+              "dataEditorConfig": {
```
arosien, line 416:
```suggestion
              "dataEditor": {
```
`Config` seems redundant.

package.json
```diff
@@ -413,27 +413,14 @@
...
+                "type": "object",
+                "description": "Configuration for Data Editor. Settings are omegaEditPort, logFile and logLevel",
+                "default": {
+                  "omegaEditPort": 9000,
```
arosien, line 420:
```suggestion
                  "port": 9000,
```
Maybe just `port`?

package.json
arosien, line 621: I'm unsure what the schema definitions in this `configuration` section are for. They aren't official sections listed under https://code.visualstudio.com/api/extension-guides/debugger-extension#anatomy-of-the-package.json-of-a-debugger-extension. 

Looking around, it seems to be a custom configuration section we use to provide default values(?). If so, it can probably be removed and the `contributes/debuggers/configurationAttributes` section can be the sole location of launch schema and defaults.

This should probably be a different ticket for a different release.

package.json
scholarsmate: Since it's namespaced, I think just `port` is a fine choice.

package.json
scholarsmate: Agreed.

Shanedell: @arosien Should this be changed everywhere `dataEditorConfig` is used or just inside of package.json? Meaning like the interface that was created, etc.

package.json
Shanedell: Yeah this what it is for. However, it was separated from `configurationAttributes` as we were using it for defaults and making sure the type is right initially. I don't think we want to move them into `configurationAttributes`, doing this allows the user to change their values, but they are only there for initial values of object attributes. I think we actually wouldn't want the user to have the ability to set these item, at least some of them since it really wouldn't do anything.

package.json
Shanedell: I just renamed the package.json settings. So that in TS code you still know its a dataEditorConfig not the actual dataEditor.

arosien (approved):
👍 code looks good, haven't tried it out though

PR #623, opened 2023-05-05
Save debugger output to a file

### LINKED ISSUE #614: 1.3.0-rc1 error erased from terminal window before it can be read
I am trying to configure the vscode extension to allow me to single step through a DFDL schema.

I am unable to debug what is wrong with the launch.json because it fails, and a backtrace or something like one is displayed on the terminal window, but it is erased a split second later before it can be read or captured.

I'm attaching a little video of this I created by doing a MS-Teams screen share in a meeting that was just me and recording it. 

https://user-images.githubusercontent.com/1524099/235781974-3e8b06a2-e4d2-4a78-8504-2e45c53053bf.mp4

The action occurs between 8 and 10 seconds into the video. 

The VSCode environment is then stuck for a minute perhaps (I didn't time it) with a run indicator at the top still moving until it times out and leaves me with a dialog to dismiss. 

Github, in its infinite wisdom, won't let me attach the launch.json file here. And even renaming it to launch_json.txt it still won't let me attach it. So I'm inlining it here. I already know there are things that need to be added to the daffodilDebugClasspath that are not listed. The complaint on the terminal is quite likely just telling me something is missing. 

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "dfdl",
            "request": "launch",
            "name": "vmf-pmmc-dfdl run",
            "program": "${workspaceFolder}/src/main/resources/com/owlcyberdefense/vmf-multi/xsd/vmf-noc.dfdl.xsd",
            "stopOnEntry": true,
            "data": "${workspaceFolder}/src/test/resources/data/owl/binary-file-zip-message.dat",
            "infosetFormat": "xml",
            "infosetOutput": {
                "type": "file",
                "path": "/tmp/vmf-pmmc-dfdl-infoset.xml"
            },
            "debugServer": 4711,
            "openHexView": false,
            "openInfosetView": false,
            "openInfosetDiffView": false,
            "daffodilDebugClasspath": "${workspaceFolder}/src/main/resources:${workspaceFolder}/src/test/resources:${workspaceFolder}/../../zipLayerTransformer/src/main/resources:${workspaceFolder}/../noc-header/src/main/resources:${workspaceFolder}/../../../../../../opensource/DFDLSchemas/mil-std-2045/src/main/resources:${workspaceFolder}/lib_managed/jars/com.owlcyberdefense/zip-layer-transformer/zip-layer-transformer-1.0.0.jar",
            "variables": {},
            "tunables": {},
            "dataEditor.omegaEditPort": 9001,
            "dataEditor.logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
            "dataEditor.logLevel": "info"
        }
    ]
}

```

Shanedell:
Save debugger output to a file

- Use "tee" for mac/linux to save debugger output to a file.

Closes #614

### DIFF
--- a/src/daffodilDebuggerUtils.ts
+++ b/src/daffodilDebuggerUtils.ts
@@ -52,11 +52,20 @@ export const stopDebugger = (id: number | undefined = undefined) =>
 export const shellPath = (scriptName: string) =>
   osCheck(scriptName, '/bin/bash')
 
-export const shellArgs = (scriptName: string, port: number) =>
-  osCheck(
+export const shellArgs = (scriptName: string, port: number) => {
+  const logPath = vscode.workspace.workspaceFolders
+    ? vscode.workspace.workspaceFolders[0].uri.fsPath
+    : '/tmp'
+
+  return osCheck(
     ['--listenPort', `${port}`],
-    ['--login', '-c', `${scriptName} --listenPort ${port}`]
+    [
+      '--login',
+      '-c',
+      `${scriptName} --listenPort ${port} | tee ${logPath}/debugger-${port}-output.txt`,
+    ]
   )
+}
 
 export async function runDebugger(
   rootPath: string,


Shanedell:
@mbeckerle So with the way the terminal exits it is based on the script being ran. So unfortunately I am not able to keep the terminal open after the script exits Instead I made the script use `tee` to also output the debugger data to a local file. By default it should go your workspace folder and be named `debugger-${port}-output.txt`, port being your debugServer port, but if for some reason it cant find your workspace folder it will go to `/tmp`. This should still allow you debug the issue better.

NOTE: The logging was only added for mac/linux since I know both have `tee`. I saw online Windows may also have `tee` but others said `type` would be better. So if anybody knows more on that let me know and I can add that in as well.

scholarsmate:
While this is a quick fix to the current problem, I think it will be better to use a logging configuration (e.g., logback.conf or something) and have that write to file and std{out.err} instead of piping the output to `tee`.  The Ωedit Scala server uses a logback config file that is generated in `client.ts` and read by the server on startup for server logging.

mbeckerle (approved):
The change is fine as a workaround.

But I don't get why this is needed. 

Clearing a terminal window when a script exits ... That's entirely broken for a terminal window. Isn't the whole point of a terminal window that commands/scripts are run and they output something, and after they return to the prompt you can still read what was output by the prior command/script.  

For the vast majority of command line tools, the whole point is for them to print something, then they exit, and the terminal continues to display what they had output so you can read it while pondering what command to issue next. A long-running command, even running in background isn't really any different.

Shanedell:
@scholarsmate @mbeckerle Closing this PR in favor of https://github.com/apache/daffodil-vscode/pull/626 as that is much better fix for the issue. Please give it a look when you can, thanks!

PR #625, opened 2023-05-05
Add conditional defaults to TDML properties

### LINKED ISSUE #622: TDML configuration causing server timeout issue when running from launch.json
With the following configuration:
``` 
"tdmlConfig": {
 "action": "none",
 "name": "${command:AskForTDMLName}",
 "description": "${command:AskForTDMLDescription}",
 "path": "${command:AskForTDMLPath}" },
```

The debugger is continuing to ask for TDML configurations even when the action is "none"
image

If the configuration are not given within a time limit, the debugger will terminate and crash.

image

When this configuration was used:
```
           "tdmlConfig": {
                "action": "none",
                "name": "tdml1",
                "description": "tdml1",
                "path": "${workspaceFolder}/target/foo.tdml"
            },
```
The debugger will continue to run and not ask for the TDML configurations because they are already provided.

michael-hoke:
This is an untested draft. Unsure if this is the appropriate fix - Creating a draft PR in order to get asynchronous feedback.

Closes #622

### DIFF
--- a/src/tests/suite/utils.test.ts
+++ b/src/tests/suite/utils.test.ts
@@ -35,12 +35,6 @@ suite('Utils Test Suite', () => {
       type: 'none',
       path: '${workspaceFolder}/infoset.xml',
     },
-    tdmlConfig: {
-      action: 'none',
-      name: '${command:AskForTDMLName}',
-      description: '${command:AskForTDMLDescription}',
-      path: '${command:AskForTDMLPath}',
-    },
     stopOnEntry: true,
     useExistingServer: false,
     trace: true,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -100,7 +100,18 @@ export function getConfig(
   openInfosetDiffView = false,
   daffodilDebugClasspath: string = ''
 ) {
-  return {
+  // Keep 'none' here as a default so that we don't always add the tdmlConfig block
+  let tdmlAction = defaultConf.get('tdmlAction', 'none')
+  let tdmlPropsNeeded = false
+  if (
+    tdmlAction === 'generate' ||
+    tdmlAction === 'append' ||
+    tdmlAction === 'execute'
+  ) {
+    tdmlPropsNeeded = true
+  }
+
+  let jsonConfig = {
     name: name,
     request: request,
     type: type,
@@ -120,17 +131,6 @@ export function getConfig(
             'infosetOutputFilePath',
             '${workspaceFolder}/infoset.xml'
           ),
-        },
-    tdmlConfig: tdmlConfig
-      ? tdmlConfig
-      : {
-          action: defaultConf.get('tdmlAction', 'none'),
-          name: defaultConf.get('tdmlName', '${command:AskforTDMLName}'),
-          description: defaultConf.get(
-            'tdmlDescription',
-            '${command:AskForTDMLDescription}'
-          ),
-          path: defaultConf.get('tdmlPath', '${workspaceFolder}/infoset.tdml'),
         },
     stopOnEntry: stopOnEntry
       ? stopOnEntry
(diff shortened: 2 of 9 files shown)


src/utils.ts
```diff
@@ -120,12 +130,16 @@ export function getConfig(
...
-          action: defaultConf.get('tdmlAction', 'none'),
-          name: defaultConf.get('tdmlName', '${command:AskforTDMLName}'),
-          description: defaultConf.get(
-            'tdmlDescription',
-            '${command:AskForTDMLDescription}'
-          ),
+          action: tdmlAction,
+          name: tdmlPropsNeeded
+            ? defaultConf.get('tdmlName', '${command:AskforTDMLName}')
+            : defaultConf.get('tdmlName', 'tdmlNamePlaceholder}'),
```
mbeckerle, line 136: Hmmm. Why create a different text placeholder string? 

Also, what is a tdmlName - there are too many kinds of names - the TDML file's name, the testSuite name, the test itself name,... so use a longer more explicit identifier. 

Watch out also for "config", as there is a testSuite defaultConfig and each test has a config also, and a testSuite can have multiple named defineConfig children. 

Suggest name "tdmlVSCodeConfig" would keep things clearer. 

Ideally, this should prompt you for a TDML file name, and then store it here, i.e., be sticky about the selection.

mbeckerle (approved):
Just a few naming suggestions.

src/utils.ts
michael-hoke: > Hmmm. Why create a different text placeholder string?
> 
The root issue of #622 is that VSCode is trying to resolve ${command:AskforTDMLName} even in cases where it shouldn't, i.e. when a user runs a launch config with TDMLAction set to 'none'. The idea of this placeholder is to make it so that configs that are created won't use the command as the default value, which would require users to modify the default launch config to make it work for the default use case. On the other hand, if we run into a config with a TDMLAction that isn't none, we want the command to be the default value.

At best, this only reduces user impact for #622, so we're still looking into ways to fix it.

Thanks for the other suggestions - I don't want to try to overcrowd 1.3.0, but these will be helpful when we look to improve upon the TDML functionality in upcoming releases.

arosien: I don't think there should be an action of type `none`; if you don't want a TDML action then the TDML config section should be optional. Then there won't be any need to suppress variable interpolations, and if you do want an action then the variables will get interpolated.

src/tests/suite/utils.test.ts
```diff
@@ -35,12 +35,7 @@ suite('Utils Test Suite', () => {
       type: 'none',
       path: '${workspaceFolder}/infoset.xml',
     },
-    tdmlConfig: {
-      action: 'none',
-      name: '${command:AskForTDMLName}',
-      description: '${command:AskForTDMLDescription}',
-      path: '${command:AskForTDMLPath}',
-    },
+    tdmlConfig: null,
```
Shanedell, line 38: Remove this line. This will break the testing because you specifically don't add the `tdmlConfig` if the tdml props aren't needed.

Shanedell:
The main fix that is really needed is a check that if the tdml action is `none` if it is then you set `name, description and path` to empty strings or null. Since that is not currently being done and the items are left being something like `${command:AskForTDMLName}`, those will get ran. If the debugger has a string `${command:.*}` it will always run that command. So those values need to overridden if you wish to not prompt.

michael-hoke:
Ideally, yes, that is how to fix it, but there seems to be separate code paths for running through the command palette versus running from a launch config. If I try to do that in getConfig, it works with the command palette but not with the launch config.

I don't know if we can change the behavior or where the code changes would have to be for the launch config to work, although I haven't had much time to look into it.

Shanedell:
So can you rebase this on `main` or make a new branch based off of `main` and try just adding the else I mentioned in issue #622 and see if that works? The launch config wizard was updated in a recent PR so the issue you are referring to might be resolved

Shanedell:
@michael-hoke Could you at least rebase this PR on main and delete the line mentioned. I would be fine approving this really anyways

michael-hoke:
@Shanedell working on a fix now - assuming everything passes, it will be up soon. As it is right now, it does not fix the issue that it claims to close, so it should not be merged in anyways.

michael-hoke:
As a note: the default value for TDMLConfig.Action in package.json needs to remain as 'none' even though that isn't a valid action anymore. The VSCode API function we use to get our default configuration (vscode.workspace.getConfiguration()) keys off of this when it gets the default configuration.

Shanedell (approved):
+1 👍 . I pulled down the code and ran a `yarn test` and everything seemed to run properly. Sometimes the one test for the data editor seems to fail but then sometimes rerunning will make it pass.

Shanedell:
@michael-hoke Can you please rebase off of `main` and add "Closes https://github.com/apache/daffodil-vscode/issues/622" to git commit message. Hopefully CI will run this time

PR #655, opened 2023-06-08
Always export env variables

### LINKED ISSUE #652: vscode extension must be isolated from user's shell environment
I was unable to get the daffodil vscode 1.3.0-RC2 to work when installed.

This is because the vscode extension is sensitive to the default environment. It currently picks up many things from the users default shell environment.

The vscode extension needs to be isolated from the user's shell environment.

One example is that the vscode extension requires Java 11. 

Starting the vscode extension needs to guarantee Java 11 is what is being used. It should not use what is configured via the JAVA_HOME environment variable by the user. (Mine is defaulting to Java 17 these days.)

To get vscode to work I have this script I use to launch vscode in a minimal/empty environment with just the few things it needs. This technique is also useful to determine if you are depending on the environment in any way:

```
#! /bin/bash
if [ "$1" = "emptyEnv" ] # is it the special first arg
then
  shift # discard special first arg, and continue with the rest of the script
else
  # execute this same script again in an empty environment
  CMD=$0
  exec -c $CMD "emptyEnv" $* # exec with an empty environment and special first arg
fi
# the environment is empty when we get here
# just add env vars we need to get vscode to work
export LANG=en_US.UTF-8
export HOME=/home/mbeckerle
export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64 # java 11 required
export PATH=$JAVA_HOME/bin:$PATH
export DISPLAY=:0
export DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus # vscode spits lots of errors without this
exec /usr/bin/code
# end
```

Then, within ~/.config/Code/User/settings.json, I had to create these settings for terminal:

```
    "terminal.integrated.inheritEnv": false,
    "terminal.integrated.profiles.linux": {
        "Custom Init": {
          "path": "/bin/bash",
          "args": [
             "--norc"
          ]
        }
      },
      "terminal.integrated.defaultProfile.linux": "Custom Init"
```

That instructs vscode not to create a login shell for the terminal. The --norc option means it does not run the .bashrc file so that the java version specified in the launch script sticks. 

I am not sure the inheritEnv setting actually does anything. That alone sounds like the right thing, but that alone didn't solve the problem.
You still have everything in the login shell environment where vscode was launched. Hence the need for the environment-emptying launch script.

With those two techniques combined, a terminal created by vscode does not inherit everything my typical shell has. For example there is no daffodil on the path, no DAFFODIL_CLASSPATH env var, etc. Rather it has only a small environment of just what vscode needs.

Given the above if I just create a terminal with Terminal > new Terminal: The environment is relatively simple:

```
bash-5.0$ env
COLORTERM=truecolor
TERM_PROGRAM_VERSION=1.78.2
JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
NO_AT_BRIDGE=1
PWD=/home/mbeckerle/dataiti/opensource/DFDLSchemas/PCAP
VSCODE_GIT_ASKPASS_NODE=/usr/share/code/code
HOME=/home/mbeckerle
LANG=en_US.UTF-8
GIT_ASKPASS=/usr/share/code/resources/app/extensions/git/dist/askpass.sh
CHROME_DESKTOP=code-url-handler.desktop
VSCODE_GIT_ASKPASS_EXTRA_ARGS=--ms-enable-electron-run-as-node
TERM=xterm-256color
VSCODE_GIT_IPC_HANDLE=/tmp/vscode-git-56a269c0f5.sock
DISPLAY=:0
SHLVL=1
VSCODE_GIT_ASKPASS_MAIN=/usr/share/code/resources/app/extensions/git/dist/askpass-main.js
GDK_BACKEND=x11
PATH=/usr/lib/jvm/java-11-openjdk-amd64/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
ORIGINAL_XDG_CURRENT_DESKTOP=undefined
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
TERM_PROGRAM=vscode
_=/usr/bin/env
```

If I run daffodil debugging, then c-C in the terminal it creates, then the environment has this added environment variable only. 

```
DAFFODIL_DEBUG_CLASSPATH=/home/mbeckerle/dataiti/opensource/DFDLSchemas/PCAP/../ethernetIP/src/main/resources:/home/mbeckerle/dataiti/opensource/DFDLSchemas/PCAP/lib_managed/jars/com.owlcyberdefense/dfdl-ethernetip/dfdl-ethernetip-1.2.0.jar
```

This works.

shanedell:
Always export env variables

- Close any opened daffodil-debugger terminal by default.
  - This makes sure any env we rely on get set every time a new debug is ran.

Closes #652

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -238,20 +238,34 @@ export const getTerminal = (
   createTerminal: boolean
 ) => {
   if (!createTerminal) {
-    if (vscode.window.activeTerminal) {
-      const activeTerminal = vscode.window.activeTerminal
-
-      // check allows for shell name or full path to shell in terminal name
-      if (activeTerminal.name.includes(terminalName)) return activeTerminal
-    }
+    vscode.window.terminals.forEach((terminal) => {
+      if (terminal.name.includes(terminalName)) terminal.dispose()
+    })
   }
 
   // If no good active terminal available create new one
-  return vscode.window.createTerminal({
+  const terminal = vscode.window.createTerminal({
     name: terminalName,
     hideFromUser: hideTerminal,
-    env: env,
   })
+
+  // Looping to manual set all env variables. Setting "env: env" inside of createTerminal won't override variables already set
+  for (var key in env) {
+    if (key !== null && key !== undefined) {
+      let workspaceFolder = vscode.workspace.workspaceFolders
+        ? vscode.workspace.workspaceFolders[0].uri.fsPath
+        : ''
+      let exportVar = `${osCheck('set', 'export')} ${key}=${env[key]}`
+
+      if (exportVar.includes('${workspaceFolder}')) {
+        exportVar = exportVar.replaceAll('${workspaceFolder}', workspaceFolder)
+      }
+
+      terminal.sendText(exportVar, true)
+    }
+  }
+
+  return terminal
 }
 
 export async function runScript(


mbeckerle (approved):
+1

scholarsmate (approved):
This change looks fine, but I don't see where we're doing anything with the environment.  Did everything you intend get checked in?

shanedell:
@scholarsmate Whenever a new terminal is created it sets `env: env`. So the change is pretty simple, by default we always destroy any old debugger terminal and make a new terminal. This will make it so the env variables we want to set are being set before the debugger is ran. I am still testing an issue where when I set the `DAFFODIL_DEBUG_CLASSPATH` inside of my `.bashrc` or `.zshrc` it is not being override but creating a new terminal before each debug will be vital in fixing the issue I believe.

shanedell:
@mbeckerle @scholarsmate The new update fixes the issues where if `DAFFODIL_DEBUG_CLASSPATH` was already set in some other way it wasn't being overriden. This was because the VSCode API method `createTerminal` doesn't seem to overwrite env variables when you specify `env: env`. So to get around it I have the code loop over all env variables and manually set them before running the debug.

PR #656, opened 2023-06-12
Fix issue with launch wizard parsing

shanedell:
Fix issue with launch wizard parsing

- Update launch wizard to parse the daffodilDebuggerClasspathTable items textContent instead of innerHTML.
  - Using innerHTML was causing an issue where the wizard would set the classpath as the css class name "debug-classpath-item" instead of its actual value.

### DIFF
--- a/src/launchWizard/launchWizard.js
+++ b/src/launchWizard/launchWizard.js
@@ -182,13 +182,21 @@ function save() {
 
   let daffodilDebugClasspath = ''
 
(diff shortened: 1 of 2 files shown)


shanedell:
@nlewis05 please make sure this PR does in fact fix this issue you were seeing an approve if it does.

mbeckerle (approved):
+1

nlewis05 (approved):

scholarsmate (approved):
+1, LGTM

PR #659, opened 2023-06-13
Remove debugger extension config from launch.json

shanedell:
Remove debugger extension config from launch.json

- The removed config from .vscode/launch.json is a config for running a debug session.
  - We don't want this in this .vscode/launch.json as it should be used for running the extension, running extension tests or running the debugger. It won't be able to launch an interactive debug session, that type of config needs to added to the workspaces .vscode/launch.json instead.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,24 +19,6 @@
 	"version": "0.2.0",
 	"configurations": [
 		{
-			"type": "dfdl",
-			"request": "launch",
-			"name": "DfdlDeb",
-			"program": "${command:AskForProgramName}",
-			"stopOnEntry": true,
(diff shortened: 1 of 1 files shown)


mbeckerle (approved):
+1

If I understand this change, this configuration stuff was just in the wrong spot, in the daffodil-vscode's own .vscode/launch.json.

So this has nothing to do with the launch.json that is part of any given schema project.

shanedell:
@mbeckerle Yeah that is absolutely correct. Yeah the configuration that was removed, is supposed to be used inside of a `.vscode/launch.json` for a schema when you have the extension running or installed.

scholarsmate (approved):
+1

PR #660, opened 2023-06-14
Bind the debug server to the loopback interface.

### LINKED ISSUE #658: Default to binding debugger to localhost interface vs. 0.0.0.0
Related to #639, we bind to "all interfaces" (0.0.0.0) vs. a more restricted localhost (127.0.0.1). This should be more friendly to restricted machine setups.

arosien:
A.k.a., 127.0.0.1. We think 0.0.0.0 is used by default, and want
to use a more secure local interface.

Fixes #658.

### DIFF
--- a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
+++ b/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
@@ -454,7 +454,7 @@ object DAPodil extends IOApp {
 
       state <- Ref[IO].of[State](State.Uninitialized)
 
-      address = new InetSocketAddress(options.listenPort)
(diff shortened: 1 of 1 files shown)


arosien:
On startup prints:
```
2023-06-13 21:50:38,002 [io-compute-4] INFO  o.a.d.d.d.DAPodil - waiting at tcp://localhost:4711 
```

FYI at the same time, when the local server isn't up and a launch config tries to connect, the VS Code extension reports `connect ECONNREFUSED 127.0.0.1:4711`.

mbeckerle (approved):
+1

The change looks correct in that I'm pretty sure it will work, and the 0.0.0.0 thing did look unusual to me. 

I cannot speak to whether this has improved security or not as I have no expertise in how client/server connections are supposed to be secured.

shanedell (approved):
+1 👍

scholarsmate (approved):
+1

Issue #650, opened 2023-06-05
launch config wizard doesn't help with very tricky daffodilDebugClasspath


mbeckerle:
the wizard that creates the launch.json, doesn't actually let you populate the daffodilDebugClaspath. You have to edit that by hand, and nothing verifies it, so any typo and you get tedious failures that are hard to figure out.

The wizard should help you put directories and corresponding jars on the daffodilDebugClasspath, and it should not let you put a non-existing directory or jar file on it. Ideally, at launch time the daffodilDebugClasspath should be checked, just in case something on it no longer exists in the file system.

PR #664, opened 2023-06-15
Launch Wizard Validation

nlewis05:
[Launch Wizard Validations:](https://github.com/apache/daffodil-vscode/pull/664/commits/83dc80003a3be1d960ed64385f1fbb3cf1340eb0) 

- Check that daffodil debug classpath items exist when debug is being
  started.
- Don't allow duplicate items inside of the daffodil debug classpath
  table

Closes https://github.com/apache/daffodil-vscode/issues/650

### DIFF
--- a/src/daffodilDebugger.ts
+++ b/src/daffodilDebugger.ts
@@ -189,7 +189,14 @@ export async function getDebugger(
       // Get daffodilDebugger class paths to be added to the debugger
       let daffodilDebugClasspath = ''
 
+      //check if each classpath still exists
       if (config.daffodilDebugClasspath) {
+        config.daffodilDebugClasspath.split(':').forEach((entry) => {
+          if (!fs.existsSync(entry)) {
+            throw new Error(`File or directory: ${entry} doesn't exist`)
+          }
+        })
+
         daffodilDebugClasspath = config.daffodilDebugClasspath.includes(
           '${workspaceFolder}'
         )
--- a/src/launchWizard/launchWizard.js
+++ b/src/launchWizard/launchWizard.js
@@ -18,12 +18,54 @@
 // Retrieve vscode api - Doing this multiple times causes issues with the scripts
 const vscode = acquireVsCodeApi()
 
+// Function to get config index
+function getConfigIndex() {
+  var configSelectionBox = document.getElementById('configSelected')
+  var configSelectedValue =
+    configSelectionBox.options[configSelectionBox.selectedIndex].value
+
+  if (configSelectedValue === 'New Config') {
+    document.getElementById('nameLabel').style =
+      'margin-top: 10px; visibility: visible;'
+  } else {
+    document.getElementById('nameLabel').style = 'visibility: hidden;'
+  }
+
+  return configSelectedValue === 'New Config'
+    ? -1
+    : configSelectionBox.selectedIndex
+}
+
+// Function get daffodil debug classpath string
+function getDaffodilDebugClasspathString() {
+  let childNodes = document.getElementById(
+    'daffodilDebugClasspathTable'
+  ).childNodes
+
+  return Array.from(childNodes)
+    .map(
+      (childNode) =>
+        childNode.textContent
+          .replace(/\B\s+|\s+\B/g, '') // remove any un-needed whitespace. will not remove spaces between characters
+          .replace('-', '') // remove initial - in front of every item
+    )
+    .filter((cp) => cp != '')
+    .join(':')
+}
+
 // Function to call extension to open file picker
 function filePicker(id, description) {
+  let extraData = {}
+  if (id === 'daffodilDebugClasspath') {
+    extraData['daffodilDebugClasspath'] = getDaffodilDebugClasspathString()
+  }
+
   vscode.postMessage({
     command: 'openFilePicker',
     id: id,
     description: description,
+    configIndex: getConfigIndex(),
+    extraData: extraData,
   })
 }
 
@@ -115,23 +157,9 @@ function updateTDMLAction() {
 
 // Function to update config selected, also display name input box if 'New Config' selected
 function updateSelectedConfig() {
-  var configSelectionBox = document.getElementById('configSelected')
-  var configSelectedValue =
-    configSelectionBox.options[configSelectionBox.selectedIndex].value
-
-  if (configSelectedValue === 'New Config') {
-    document.getElementById('nameLabel').style =
-      'margin-top: 10px; visibility: visible;'
-  } else {
-    document.getElementById('nameLabel').style = 'visibility: hidden;'
-  }
-
-  let configIndex =
-    configSelectedValue === 'New Config' ? -1 : configSelectionBox.selectedIndex
-
   vscode.postMessage({
     command: 'updateConfigValue',
-    configIndex: configIndex,
+    configIndex: getConfigIndex(),
   })
 }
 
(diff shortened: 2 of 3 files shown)


shanedell:
@nlewis05 This should fix your issues. Give it a test and then when ready squash commits with `git rebase -i upstream/main`. Update your final commit message to be something like the below:

```txt
Launch Wizard Validations:

- Check that daffodil debug classpath items exist when debug is being started.
- Don't allow duplicate items inside of the daffodil debug classpath table

Closes #650 
```

If there was an issue for the duplicate thing as well add a new line under the provided closes with `Closes #ISSUE_NUMBER`

src/daffodilDebugger.ts
```diff
@@ -189,7 +189,14 @@ export async function getDebugger(
       // Get daffodilDebugger class paths to be added to the debugger
       let daffodilDebugClasspath = ''
 
+      //check if each classpath still exists
       if (config.daffodilDebugClasspath) {
+        config.daffodilDebugClasspath.split(':').foreach((classpath) => {
```
shanedell, line 194: Also little typo here, just update `.foreach` to be `.forEach`.

shanedell (approved):
+1 👍

src/daffodilDebugger.ts
```diff
@@ -189,7 +189,14 @@ export async function getDebugger(
...
 
+      //check if each classpath still exists
       if (config.daffodilDebugClasspath) {
+        config.daffodilDebugClasspath.split(':').forEach((classpath) => {
+          if (!fs.existsSync(classpath)) {
+            throw new Error(`File or directory: ${classpath} doesn't exist`)
+          }
+        })
+
         daffodilDebugClasspath = config.daffodilDebugClasspath.includes(
```
mbeckerle, line 200: What does this line do? 

Is this setting a default value of the workspace dir as the only content of daffodilDebugClasspath? If so I would not expect this as the default.

src/launchWizard/launchWizard.js
```diff
@@ -18,12 +18,62 @@
...
+
+// Function get daffodil debug classpath string
+function getDaffodilDebugClasspathString() {
+  let list = document.getElementById('daffodilDebugClasspathTable')
+
+  let daffodilDebugClasspath = ''
+
+  list.childNodes.forEach((childNode) => {
+    let classpath = childNode.textContent
+      .replaceAll(' ', '') // remove any un-needed whitespace
```
mbeckerle, line 47: How do you know it is un-needed whitespace?

I hate having spaces in file/path names, but people do it.

src/launchWizard/launchWizard.js
```diff
@@ -18,12 +18,62 @@
...
+      .replace('-', '') // remove initial - in front of every item
+      .split('\n')
+      .filter((cp) => cp != '')
+      .join(':')
+
+    if (classpath != '') {
+      daffodilDebugClasspath +=
+        childNode === list.childNodes[list.childNodes.length - 1]
+          ? classpath
+          : classpath + ':'
```
mbeckerle, line 57: I thought you took care of putting the ":" with the join(':') at line 51. 

This code is a funny mix of imperative and functional. I suggest go with all functional, and compute the daffodilDebugClasspath from the childNodes without side effects.

src/launchWizard/launchWizard.js
```diff
@@ -18,12 +18,62 @@
...
+}
+
 // Function to call extension to open file picker
 function filePicker(id, description) {
+  let extraData = {}
+  if (id === 'daffodilDebugClasspath') {
+    extraData['daffodilDebugClasspath'] = getDaffodilDebugClasspathString()
+  }
+
   vscode.postMessage({
```
mbeckerle, line 71: Why is the extension posting a message to itself telling it to do something, which will happen by side effect. Why is that better than just calling the thing? Is this how non-blocking is achieved?

src/launchWizard/launchWizard.js
```diff
@@ -18,12 +18,62 @@
 // Retrieve vscode api - Doing this multiple times causes issues with the scripts
```
mbeckerle, line 18: Can you explain why there are both launchWizard.js and launchWizard.ts files? I thought vscode was typescript based, but I noticed some time ago that there is also a smaller portion of ".js". 

Any short explanation to motivate this would help me better understand the code. Up to you whether that is just a response to this comment, or becomes a comment in the code (somewhere, perhaps not here) or doc.

src/launchWizard/launchWizard.ts
```diff
@@ -188,10 +188,9 @@ async function createWizard(ctx: vscode.ExtensionContext) {
           panel.dispose()
```
mbeckerle, line 188: Comment about above code that isn't part of this change set.

This is really a suggestion for the future. 

The most common thing I want to do with a config is clone one, then update it. 

You currently offer create a new one, update an existing one. But I don't see copy one. 

Having gone through the tedium of configuring one config for a test I want to use while debugging, for the most part I want to clone it and just change the test data file to be used.

mbeckerle (approved):
Small suggestions.

Mostly I'm just asking questions which may motivate adding code comments, or not.

src/launchWizard/launchWizard.js
shanedell: So the reason there is both `launchWizard` .ts and .js is because the webview UI will directly communicate to JS scripts. However, we may need some data that only be gotten through the typescript, such as different npm packages. So the posting message stuff done in both JS and TS is how we communicate from the WebView JS to the TypeScript code.

src/launchWizard/launchWizard.js
shanedell: So the reason this happens is because this JS can't have access to some different items like node packages that the TS code will. So this is needed to make all functionality work. The JS basically interacts with UI for events, getting data, etc. Then it sends that data back to the extension TS which takes care of writing a file, calculation, etc. The TS code then sends back any needed information to the JS code where it can update the UI with that information.

src/launchWizard/launchWizard.js
shanedell: So the join on line 51 may not do anything because each childNode item should only have one classpath. As where this code this creating a single string that will have all classpath items in it separated via `:`.

src/launchWizard/launchWizard.js
```diff
@@ -18,12 +18,62 @@
...
+    document.getElementById('nameLabel').style = 'visibility: hidden;'
+  }
+
+  return configSelectedValue === 'New Config'
+    ? -1
+    : configSelectionBox.selectedIndex
+}
+
+// Function get daffodil debug classpath string
+function getDaffodilDebugClasspathString() {
```
shanedell, line 40: @nlewis05 you should be able to update this function to just be this

```ts
// Function get daffodil debug classpath string
function getDaffodilDebugClasspathString() {
  let childNodes = document.getElementById(
    'daffodilDebugClasspathTable'
  ).childNodes

  return Array.from(childNodes)
    .map(
      (childNode) =>
        childNode.textContent
          .replace(/\B\s+|\s+\B/g, '') // remove any un-needed whitespace. will not remove spaces between characters
          .replace('-', '') // remove initial - in front of every item
    )
    .filter((cp) => cp != '')
    .join(':')
}
```

src/launchWizard/launchWizard.ts
shanedell: Added this a new issue > https://github.com/apache/daffodil-vscode/issues/670

src/launchWizard/launchWizard.js
shanedell: So to further explain for any custom webview or anything we make we need to make a JS and TS file. This is because JS is used for talking to the UI/DOM and then the TS is used for calculating, importing packages to run tasks, etc. We need it this way as it how the API is designed that only JS can communicate to the UI/DOM and thats basically all it can do. Then TS is the opposite, where it can do most things except access the UI directly. I hope this makes sense, let me know if doesn't I can try to explain it better.

shanedell:
@mbeckerle If my explanation of some of your items is okay are you good with me merging this into main?

PR #666, opened 2023-06-16
Remove blocking dialogs for debugger errors

### LINKED ISSUE #649: Blocking dialog displays schema definition errors
If I modify my schema and make any mistake, then I get a blocking dialog showing the SDEs. 

This is basically intolerable, because you must dismiss it, and thereby lose the information you need from it, in order to look back at the schema to try to fix it. 

Back when the vscode extension was only a debugger this may have been more tolerable, because it was assumed that one would author the schema elsewhere. But now with the intellisense features coming along, one wants to author/edit the schema in the vscode extension also. So it needs to support editing the schema, getting SDEs, stepping through the SDEs so you can fix them. Try again, etc. until it is runnable. And then run in the debugger. 

The bare minimum fix would be to output the SDEs to a terminal window, or other persisting window that one can scroll back and forth in. Anything non-blocking really. 

You must be able to move through the SDEs as you move the editor to fix the schema. 

 I tested this on 1.3.0-RC2 of the vscode extension

shanedell:
Remove blocking dialogs for debugger errors

- When some errors happen in the debugger they would cause a large blocking dialog in the extension.
  - This didn't looks good and didn't allow you to quickly fix issues.
- The errors now show in the terminal and small message in the bottom right corner.
  - The message will state the debugger crashed with some information as to why. Then states to look at debugger output for more information.
- Created a new case class for error events to relay back to the extension.
- Update extension to catch error events relayed from the debugger.
- Create class for handling debugEvents so we don't have too many places changing onDidReceiveDebugSessionCustomEvent, on when absolutely necessary like in the hexview.

Closes #649

### DIFF
--- a/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
+++ b/server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
@@ -49,6 +49,8 @@ trait DAPSession[Req, Res, Ev] {
 
   def sendResponse(response: Res): IO[Unit]
   def sendEvent(event: Ev): IO[Unit]
+  def abort(event: DebugEvent): IO[Unit]
+  def abort(event: DebugEvent, logMessage: String): IO[Unit]
 }
 
 object DAPSession {
@@ -63,6 +65,14 @@ object DAPSession {
 
       def sendEvent(event: DebugEvent): IO[Unit] =
         Logger[IO].info(show"<E $event") *> IO.blocking(server.sendEvent(event))
+
+      /** Send DebugEvent back to extension and exit session, ending debug */
+      def abort(event: DebugEvent): IO[Unit] =
+        sendEvent(event) *> sendEvent(new Events.TerminatedEvent())
+
+      /** Log error then send DebugEvent back to extension and exit session, ending debug */
+      def abort(event: DebugEvent, logMessage: String): IO[Unit] =
+        Logger[IO].error(logMessage) *> sendEvent(event) *> sendEvent(new Events.TerminatedEvent())
     }
 
   def resource(socket: Socket): Resource[IO, DAPSession[Request, Response, DebugEvent]] =
@@ -179,9 +189,7 @@ class DAPodil(
         disconnect(request, args)
       case extract(Command.EVALUATE, args: EvaluateArguments) =>
         eval(request, args)
-      case _ =>
-        Logger[IO].warn(show"unhandled request $request") *> session
-          .sendResponse(request.respondFailure())
+      case _ => session.abort(ErrorEvents.UnhandledRequest, show"unhandled request $request")
     }
 
   /** State.Uninitialized -> State.Initialized */
@@ -205,8 +213,10 @@ class DAPodil(
         debugee(request) match {
           case Left(errors) =>
             state.set(DAPodil.State.FailedToLaunch(request, errors, None)) *>
-              Logger[IO].warn(show"error parsing launch args: ${errors.mkString_(", ")}") *> session
-                .sendResponse(request.respondFailure(Some(show"error parsing launch args: ${errors.mkString_(", ")}")))
+              session.abort(
+                ErrorEvents.LaunchArgsParseError,
+                show"error parsing launch args: ${errors.mkString_("\n")}"
+              )
           case Right(dbgee) =>
             for {
               launched <- hotswap.swap {
@@ -219,10 +229,7 @@ class DAPodil(
                     DAPodil.State
                       .FailedToLaunch(request, NonEmptyList.of("couldn't launch from created debuggee"), Some(t))
                   ) *>
-                    Logger[IO].warn(t)(show"couldn't launch, request $request") *>
-                    session.sendResponse(
-                      request.respondFailure(Some(show"Couldn't launch Daffodil debugger: ${t.getMessage()}"))
-                    )
+                    session.abort(ErrorEvents.RequestError, show"couldn't launch, request $request")
                 case Right(launchedState) =>
                   state.set(launchedState) *>
                     session.sendResponse(request.respondSuccess())
@@ -242,7 +249,7 @@ class DAPodil(
             )
           )
         }
-      case _ => session.sendResponse(request.respondFailure())
+      case _ => session.abort(ErrorEvents.SourceError)
     }
 
   def source(request: Request, args: SourceArguments): IO[Unit] =
@@ -252,11 +259,11 @@ class DAPodil(
           .sourceContent(DAPodil.Source.Ref(args.sourceReference))
           .flatMap {
             case None =>
-              session.sendResponse(request.respondFailure(Some(s"unknown source ref ${args.sourceReference}")))
+              session.abort(ErrorEvents.SourceError)
             case Some(content) =>
               session.sendResponse(request.respondSuccess(new SourceResponseBody(content.value, "text/xml")))
           }
-      case _ => session.sendResponse(request.respondFailure())
+      case _ => session.abort(ErrorEvents.SourceError)
     }
 
   def setBreakpoints(request: Request, args: SetBreakpointArguments): IO[Unit] =
@@ -278,8 +285,7 @@ class DAPodil(
           _ <- session.sendResponse(response)
         } yield ()
       case _: DAPodil.State.FailedToLaunch =>
-        Logger[IO].warn("ignoring setBreakPoints request since previous launch failed") *>
-          session.sendResponse(request.respondFailure())
+        Logger[IO].warn("ignoring setBreakPoints request since previous launch failed")
       case s => DAPodil.InvalidState.raise(request, "Launched", s)
     }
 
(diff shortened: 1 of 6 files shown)


server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -49,8 +49,11 @@ trait DAPSession[Req, Res, Ev] {
 
   def sendResponse(response: Res): IO[Unit]
   def sendEvent(event: Ev): IO[Unit]
+  def terminate(message: String): IO[Unit]
```
mbeckerle, line 52: There are so many reasons the server could terminate. Is this rich enough? Passing just a string back?

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -358,11 +363,11 @@ class DAPodil(
           _ <- data.stack
             .findFrame(DAPodil.Frame.Id(args.frameId))
             .fold(
-              Logger[IO].warn(
+              Logger[IO].error(
```
mbeckerle, line 366: Can you introduce a one-liner method call to replace all these 5 lines of error-checking and handling? 

This looks like an abend to me. You are logging the error. So it appears this is never supposed to happen, i.e., one is technically always supposed to find scopes for the frames.

src/adapter/daffodilEvent.ts
```diff
@@ -0,0 +1,38 @@
...
+export function handleDebugEvent(e: vscode.DebugSessionCustomEvent) {
+  switch (e.event) {
+    case daf.infosetEvent:
+      let update: daf.InfosetEvent = e.body
+      let path = ensureFile(tmpFile(e.session.id))
+      fs.copyFileSync(path, `${path}.prev`)
+      fs.writeFileSync(path, update.content)
+      break
+    case daf.errorEvent:
+      let error: daf.ErrorEvent = e.body
```
mbeckerle, line 33: Ok, so for now we're just treating all errors as fatal, and getting the output out to the user somehow.

But is "debugger crashed" really the only situation here? A parse error propagating to top level isn't a crash of anything. It's normal for a parse to end that way on bad data. 

I think you need to distinguish parse errors (and someday unparse errors) as well as schema definition errors (which can be detected at runtime) from other "real" crashes. 

For today, as part of getting 1.3.0 released, I'd be happy if the errors at least were distinguished by error types and not just "debugger crashed" when they are often "normal" ends to a parse that has simply detected that the data doesn't match the schema.

mbeckerle (changes requested):
Needs some changes to better clarify the code intent and distinguish invariant failures from user-errors, or normal error behavior of the parse.

scholarsmate (approved):
+1

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -63,6 +66,10 @@ object DAPSession {
 
       def sendEvent(event: DebugEvent): IO[Unit] =
         Logger[IO].info(show"<E $event") *> IO.blocking(server.sendEvent(event))
+
+      def terminate(message: String): IO[Unit] =
```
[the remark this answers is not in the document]
shanedell: @mbeckerle  So `terminate` is for stopping the debugger when an error has occurred. I can see why this should be renamed to abend or abort. Maybe should be moved somewhere instead of session since is causing the debugger to terminate or abort instead of the session?

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -382,10 +387,10 @@ class DAPodil(
             .variables(DAPodil.VariablesReference(args.variablesReference))
             .fold(
               Logger[IO]
-                .warn(
+                .error(
                   show"couldn't find variablesReference ${args.variablesReference} in stack ${data}"
                 ) *> // TODO: handle better
-                session.sendResponse(request.respondFailure())
+                session.terminate(s"couldn't find variablesReference ${args.variablesReference} in stack ${data}")
```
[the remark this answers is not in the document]
shanedell: So in this instance should this just be a warn in the log and not cause the debugger to exit?

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -399,7 +404,7 @@ class DAPodil(
         for {
           variable <- launched.debugee.eval(args)
           _ <- variable match {
-            case None => session.sendResponse(request.respondFailure())
+            case None => session.terminate("eval failed")
```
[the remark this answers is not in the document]
shanedell: This might be something we should support in the future. Because currently even without the terminate when it send `respondFailure` the session is ended and the terminal is stopped. So this might be something to look into.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -49,8 +49,11 @@ trait DAPSession[Req, Res, Ev] {
 
   def sendResponse(response: Res): IO[Unit]
   def sendEvent(event: Ev): IO[Unit]
+  def terminate(message: String): IO[Unit]
 }
 
+case class ErrorEvent(message: String) extends Events.DebugEvent("daffodil.error")
```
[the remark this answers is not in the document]
shanedell: Created `ErrorEvents` object that will have different error events that can be thrown. Made these more specific to the error that will get thrown.

shanedell:
@mbeckerle You commented a lot on the areas on where I run a `.terminate`. So even without that and using the `respondFailure` would cause the session to end and the debugger to exit. So here that is basically what `.terminate` is doing but without causing the extension to show a blocking dialog. I would be okay with renaming `terminate` to `abort` and having it take a `event: DebugEvent`. I also can make more specific debug event classes in both the debugger and extension so they are more verbose.

src/daffodil.ts
```diff
@@ -54,7 +54,14 @@ export interface BuildInfo {
   scalaVersion: string
 }
 
-export const errorEvent = 'daffodil.error'
+export const errorEvents = {
+  'daffodil.error.unexpected': 'unexpected error',
```
mbeckerle, line 58: These are aborts, so really this doesn't need to be this pretty. Just printing the original case class name coming over from the scala code would be fine.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/ErrorEvents.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package org.apache.daffodil.debugger.dap
+
+import com.microsoft.java.debug.core.protocol.Events.DebugEvent
+
+/** Case classes for errors that we want to relay back to the extension */
+object ErrorEvents {
+  case class UnexpectedError(message: String) extends DebugEvent("daffodil.error.unexpected")
+  case class UnhandledRequest(message: String) extends DebugEvent("daffodil.error.requestunhandled")
+  case class LaunchArgsParseError(message: String) extends DebugEvent("daffodil.error.launchargparse")
```
mbeckerle, line 26: DebugEvent really doesn't need this string arg. You can just do 
```
lazy val event = Misc.getNameFromClass(this)
```
since you just need to display these strings as part of the abort info on the client side. The class name will do just fine.

mbeckerle (dismissed):
A few more suggestions to eliminate redundancy in error situations.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -67,9 +65,9 @@ object DAPSession {
       def sendEvent(event: DebugEvent): IO[Unit] =
         Logger[IO].info(show"<E $event") *> IO.blocking(server.sendEvent(event))
 
-      def terminate(message: String): IO[Unit] =
-        sendEvent(new ErrorEvent(s"${message}. check debugger output for more information.")) *>
-          sendEvent(new Events.TerminatedEvent())
+      /** Send DebugEvent back to extension and exit session, ending debug */
+      def abort(event: DebugEvent): IO[Unit] =
```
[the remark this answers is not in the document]
shanedell: yeah that is correct. they all caused the program to stop. I will remove the message param from all ErrorEvents, that makes a look of sense. I will make case objects instead of case classes as well.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
```diff
@@ -247,7 +247,7 @@ class DAPodil(
             )
           )
         }
-      case _ => session.terminate("failed to load sources")
+      case _ => session.abort(new ErrorEvents.SourceError("failed to load sources"))
```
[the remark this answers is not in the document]
shanedell: I will remove the message param from all ErrorEvents, that makes a look of sense. I will make case objects instead of case classes as well.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/ErrorEvents.scala
shanedell: I will remove the message param from all ErrorEvents, that makes a look of sense. I will make case objects instead of case classes as well.

shanedell: @mbeckerle so would something like `vscode.window.showErrorMessage(`debugger ${e.event}`)` be enough? It would display something like `debugger daffodil.error.request` in the bottom right. then the error information is inside of the terminal.

shanedell: @mbeckerle The above comment is what I latest commit does, let me if you would prefer something different.

server/core/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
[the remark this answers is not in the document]
shanedell: @mbeckerle That makes sense to me, thanks for this information. I have created issue https://github.com/apache/daffodil-vscode/issues/671 from this comment. Feel free to add to the ticket if you thing anything is missing.

mbeckerle (approved):
+1

PR #677, opened 2023-06-23
Launch Wizard updates based on 1.3.0-rc3 issues

### LINKED ISSUE #672: launch wizard text layout - can't read some things
image

Look at the Daffodil Debugger Classpath in the image. This is not readable. 

(Tested 1.3.0-RC3)

### LINKED ISSUE #674: Command Palette context sensitivity is being over-used
Using 1.3.0-RC3

To find the launch configuration wizard, you have to NOT be in an execution context. Like you cannot be partway through debugging. 

You have to click right on the right part of a schema _when not debugging_, then choose command palette from the menu. Then one of the choices is the launch config wizard. 

But if you are finished learning something using the debugger, just haven't hit the stop button yet. You go over to the editor for the schema file and click right,... launch wizard is ... not available to select. It's not there and greyed out, it's not even there. 

This is just bad. The right thing here is to have a menu with all the commands in it, with the ones that are not context relevant greyed out. 

But I claim this launch wizard should NOT be greyed out, or unavailable in this context anyway. Why shouldn't I be able to mess with the launch configuration, tweek it, and then hit the stop button on the prior debugging, or even continue the prior debugging to learn a bit more. There is no reason for this modality.

### LINKED ISSUE #675: launch config strange by-hand fixups required to make it work
Testing 1.3.0-RC3

I created a launch config using the wizard. 

When I run it, I'm prompted that the infoset output type must be "file" with a little blocking dialog. 
I look at that by hand, and it contains "console", so I edit it to "file". Things now work. 

After a bit, I decide to make a new launch configuration. So I activate the launch config wizard again. I change it to new config, which is great because it looks like it will keep the existing settings, but creates a new configuration from them.
It does keep the Debug Classpath, but I did have to re-enter the schema file again, and the data file.  

However, when I run it, I'm prompted with a little blocking dialog that the tdmlConfig action must be "generate",... or a couple other options. 

This is from the new configuration just created. Note the action is empty string, and the others are "undefined"
```
            "tdmlConfig": {
                "action": "",
                "name": "undefined",
                "description": "undefined",
                "path": "undefined"
            }
```

My prior working launch configuration has:
```
            "tdmlConfig": {
                "action": "generate",
                "name": "null",
                "description": "null",
                "path": "null"
            },
```
Note the "null" not "undefined". 

I edited my new configuration to have "generate" for the action, and the configuration then works. 

These hand correction steps should not be needed.

shanedell:
Launch Wizard updates based on 1.3.0-rc3 issues:

- Update daffodil debug classpath items CSS to make the items height and width match the content.
- Make launch wizard always available. Before was only available whenever not debugging.
- Fix issue with tdml config throw errors if action was not set.
- infosetOutput bug fixed, was defaulting to type console when it was supposed to file.

Closes #672
Closes #674
Closes #675

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,7 +279,6 @@
         "command": "launch.config",
         "title": "Configure launch.json",
         "category": "Daffodil Debug",
-        "enablement": "!inDebugMode",
         "icon": "$(debug-configure)"
       },
       {
@@ -345,8 +344,8 @@
                 "type": "object",
                 "description": "Destination for final Infoset ('file' | 'console' | 'none')",
                 "default": {
-                  "type": "console",
-                  "path": "${workspaceFolder}/infoset.xml"
+                  "type": "file",
+                  "path": "${workspaceFolder}/target/infoset.xml" 
                 }
               },
               "tdmlConfig": {
@@ -436,7 +435,7 @@
             "infosetFormat": "xml",
             "infosetOutput": {
               "type": "file",
-              "path": "${workspaceFolder}/infoset.xml"
+              "path": "${workspaceFolder}/target/infoset.xml" 
             },
             "debugServer": 4711,
             "openHexView": false,
@@ -466,7 +465,7 @@
               "infosetFormat": "xml",
               "infosetOutput": {
                 "type": "file",
-                "path": "${workspaceFolder}/infoset.xml"
+                "path": "${workspaceFolder}/target/infoset.xml" 
               },
               "debugServer": 4711,
               "openHexView": false,
@@ -549,7 +548,7 @@
           "infosetOutputFilePath": {
             "type": "string",
             "description": "Path to output for Infoset file (req: infosetOutput=file)",
-            "default": "${workspaceFolder}/infoset.xml"
+            "default": "${workspaceFolder}/target/infoset.xml" 
           },
           "stopOnEntry": {
             "type": "boolean",
--- a/src/daffodilDebugger.ts
+++ b/src/daffodilDebugger.ts
@@ -71,6 +71,90 @@ export async function getDataFileFromFolder(dataFolder: string) {
     })
 }
 
+async function getTDMLConfig(
+  config: vscode.DebugConfiguration
+): Promise<boolean> {
+  // If not supported TDML action entered, delete tdml config so no errors are thrown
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -345,7 +344,7 @@
                 "type": "object",
                 "description": "Destination for final Infoset ('file' | 'console' | 'none')",
                 "default": {
-                  "type": "console",
+                  "type": "file",
                   "path": "${workspaceFolder}/infoset.xml"
```
mbeckerle, line 348: The default location for this shoudl be "${workspaceFolder}/target/infoset.xml" so that it doesn't clutter and end up committed by accident.

mbeckerle (approved):
+1 
I added one small suggested additional change.

This looks like it will fix the issues. Plus some refactoring of code to named methods that provide self-doc is always good.

package.json
shanedell: gotcha I like this idea. I am also still looking into the selecting a single JAR issue. It is working for me on Mac but you said it doesn't work for you (which I believe you use Linux) and believe @nlewis05 mentioned he saw that as well, possibly failing on windows as well. So I am currently getting up Linux VM up and running to get this fixed.

shanedell: @mbeckerle So after getting my ubuntu VM up and running I am able to get the same issue as you and @nlewis05 when it comes to picking out single JARs. It seems the VSCode API is limiting Linux so that you have to specify that the user can pick files OR folders and doesn't allow both. One way to get around it is when you open the file picker for the JARs bring up a selection box where you chose `Single File` or `Directory` before going to the file picker, this allow for us to set up the file picker properly. Or can you think of another, possibly better, way to get around it?

shanedell: Or maybe instead of having only the `Browse` button we make two buttons. One `Add Folder(s)` and one `Add File(s)` and then how is how the code will know how to show the file picker. I think this would be nicer then the other option I stated above.

shanedell: @mbeckerle @nlewis05 Please give this another look over. I made the daffodilDebugClasspath have two buttons like mentioned above to fix the issue.

shanedell:
@scholarsmate @arosien @stricklandrbls Can you give this a look?

scholarsmate (approved):
+1, LGTM.

PR #694, opened 2023-07-11
Multiple debugger updates:

### LINKED ISSUE #471: Debugging extension in dev environment doesn't unpack current backend build
Debugging the extension launches `yarn watch` which does a `sbt universal:packageBin` which builds the backend, but the newly built backend isn't installed into the local XDG directory, so whatever release was previously installed there (from an official release, etc.) will be used. This causes confusion if you are expecting the existing backend code to be run.

Workaround: delete your `daffodil-dap` XDG directory and the extension will look in the current directory to install the local backend. (FYI, the tests unpack the current backend build into a custom directory and point the extension to there.)

Fix: do this install during the extension build.

### LINKED ISSUE #679: Allow setting the log level of the debugger
> Another rough spot is that there is so much terminal output that it's hard
to find the error messages in there. Is there a way to set the debug level
to ERROR so that the info and debug messages aren't displayed? I had to
search for ".error" to find the assertion failure statement (which has
java.lang.error as a prefix of the message).

Mentioned by @mbeckerle in his vote 1.3.0-rc4.

There should an item to configure in the launch.json, such as `debuggerLogLevel`, that allows you to set the level the debugger will log too. This value then gets sent to the debugger on launch and based on the value provided it will log to that level.

### LINKED ISSUE #691: Log debugger to console and file
Make the debugger log to both the console and a file.

shanedell:
Multiple debugger updates:

- Always extract the debugger zip.
- Update debugger to log to both console and file.
- Allow the setting of the debugger's log level and log filepath from the launch.json.
  - the logback.xml uses environment variables LOG_LEVEL and LOG_FILE to do this. This get set when the debugger is ran.
  - default log level is INFO
  - default log filepath is /tmp/daffodil-debugger.log
- Add "devMode" variable to launch.json
  - When enabled and the extension is being ran locally, it will run "sbt universal:packageBin" everytime before running the debugger. It will display a little loading in the bottom bar.
  - This can be used later to enable other features that are only available during development.

Closes #471
Closes #679
Closes #691

Loading symbol when devMode is enabled

Screenshot 2023-07-10 at 7 57 26 PM

### DIFF
--- a/debugger/src/main/resources/logback.xml
+++ b/debugger/src/main/resources/logback.xml
@@ -27,7 +27,17 @@
       <pattern>%date{ISO8601} [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n</pattern>
     </encoder>
   </appender>
-  <root level="DEBUG">
+
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+      <file>${DAFFODIL_DEBUG_LOG_FILE:-/tmp/daffodil-debugger.log}</file>
+      <append>false</append>
+      <encoder>
+          <pattern>%date{ISO8601} [%thread] %-5level %logger{15} - %msg %n</pattern>
+      </encoder>
+  </appender>
+
+  <root level="${DAFFODIL_DEBUG_LOG_LEVEL:-INFO}">
     <appender-ref ref="STDOUT" />
+    <appender-ref ref="FILE" />
   </root>
 </configuration>
--- a/package.json
+++ b/package.json
@@ -29,8 +29,9 @@
     "compile": "tsc -p ./ && yarn sbt",
     "lint": "yarn prettier src -c && cd src/svelte && yarn lint",
     "lint:fix": "yarn prettier src -w && cd src/svelte && yarn lint:fix",
-    "prewatch": "yarn gen-version-ts && yarn sbt",
-    "watch": "concurrently --kill-others \"webpack --watch --devtool nosources-source-map --config ./build/extension.webpack.config.js\" \"yarn watch:svelte\"",
+    "prewatch": "yarn gen-version-ts",
+    "scalawatch": "nodemon --verbose -e sbt,scala,xml,xjb -w debugger -w project -w build.sbt --ignore \"**/target/**\" --exec \"yarn sbt\"",
+    "watch": "concurrently --kill-others \"yarn scalawatch\" \"webpack --watch --devtool nosources-source-map --config ./build/extension.webpack.config.js\" \"yarn watch:svelte\"",
     "watch:svelte": "cd src/svelte && yarn dev",
     "webpack": "webpack --mode production --config ./build/extension.webpack.config.js",
     "prepackage": "yarn install && yarn compile && yarn webpack",
@@ -39,13 +40,14 @@
     "package-create": "node -e \"require('./build/scripts/package.ts').create()\"",
     "pretest": "yarn compile && yarn webpack",
     "test": "sbt test && node ./out/tests/runTest.js",
-    "sbt": "sbt universal:packageBin"
+    "sbt": "sbt debugger/Universal/packageBin"
   },
   "dependencies": {
-    "@vscode/debugadapter": "1.59.0",
     "@omega-edit/client": "0.9.63",
+    "@vscode/debugadapter": "1.59.0",
     "await-notify": "1.0.1",
     "hexy": "0.3.4",
+    "nodemon": "^3.0.1",
     "unzip-stream": "0.3.1",
     "uuid": "9.0.0",
     "wait-port": "1.0.4",
@@ -345,7 +347,7 @@
                 "description": "Destination for final Infoset ('file' | 'console' | 'none')",
                 "default": {
                   "type": "file",
-                  "path": "${workspaceFolder}/target/infoset.xml" 
+                  "path": "${workspaceFolder}/target/infoset.xml"
                 }
               },
               "tdmlConfig": {
@@ -420,6 +422,16 @@
                   "logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
                   "logLevel": "info"
                 }
+              },
+              "dfdlDebugger": {
+                "type": "object",
+                "description": "Configuration for debugger. Settings are logging (level and file)",
+                "default": {
+                  "logging": {
+                    "level": "INFO",
+                    "file": "/tmp/daffodil-debugger.log"
+                  }
+                }
               }
             }
           }
@@ -435,7 +447,7 @@
             "infosetFormat": "xml",
             "infosetOutput": {
               "type": "file",
-              "path": "${workspaceFolder}/target/infoset.xml" 
+              "path": "${workspaceFolder}/target/infoset.xml"
             },
             "debugServer": 4711,
             "openHexView": false,
@@ -448,6 +460,12 @@
               "port": 9000,
               "logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
               "logLevel": "info"
+            },
+            "dfdlDebugger": {
+              "logging": {
+                "level": "INFO",
+                "file": "/tmp/daffodil-debugger.log"
+              }
             }
           }
         ],
@@ -465,7 +483,7 @@
               "infosetFormat": "xml",
               "infosetOutput": {
                 "type": "file",
-                "path": "${workspaceFolder}/target/infoset.xml" 
+                "path": "${workspaceFolder}/target/infoset.xml"
               },
               "debugServer": 4711,
               "openHexView": false,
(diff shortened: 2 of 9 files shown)


mbeckerle (approved):
I appreciate the dialing back of the message output to "info" level from 'debug'. 

I did have one question which is that many of these settings appear to be in package.json twice. Is that avoidable somehow?

package.json
```diff
@@ -625,6 +655,30 @@
               "trace"
             ],
             "default": "info"
+          },
+          "debuggerLogLevel": {
```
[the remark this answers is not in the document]
shanedell: So it gets defined twice because [contributes.configuration](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) are `Contribute configuration keys that will be exposed to the user.`, see link. Then [contributes.debuggers.configurationAttributes](https://code.visualstudio.com/api/references/contribution-points#contributes.debuggers) is the schema for launch configuration arguments specific to this debugger.`, see link.

package.json
```diff
@@ -420,6 +420,30 @@
...
+                "default": "INFO"
+              },
+              "debuggerLogFilePath": {
+                "type": "string",
+                "description": "Debugger log file path",
+                "default": "/tmp/daffodil-debugger.log"
+              },
+              "devMode": {
+                "type": "boolean",
+                "description": "Enable development mode, enables certain features",
```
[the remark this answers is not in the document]
shanedell: @mbeckerle So the reason for this phrasing is that in the long term I imagine this variable being used for other items that should only happen when debugger locally. Currently all it does is rebuilds the debugger using `sbt universal:packageBin` when this variable is set to `true`

debugger/src/main/resources/logback.xml
```diff
@@ -27,7 +27,17 @@
       <pattern>%date{ISO8601} [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n</pattern>
     </encoder>
   </appender>
-  <root level="DEBUG">
+
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+      <file>${LOG_FILE}</file>
+      <append>false</append>
+      <encoder>
+          <pattern>%date{ISO8601} [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n</pattern>
```
arosien, line 35: Concern: we don't want color highlighting in the files, they will just look weird when sent through normal tools like `less`. Or at least that's how it used to be.

package.json
arosien, line 1: Suggestion: make the logging configuration it's own hierarchy, similar to the data editor config, e.g.,
```json
"debugger": {
  "logging: {
    "level": "INFO",
    "filePath": "/foo/bar/baz"
  }
}
```

src/daffodilDebugger/utils.ts
```diff
@@ -33,32 +32,45 @@ export const daffodilArtifact = (version: string): Artifact => {
...
 }
 
-export async function buildDebugger(baseFolder: string, filePath: string) {
-  if (!filePath.includes('.vscode/extension')) {
-    if (!fs.existsSync(filePath)) {
-      let command = osCheck(
-        'sbt universal:packageBin',
-        '/bin/bash --login -c "sbt universal:packageBin"'
-      ) // Needed --login so it could resolve sbt command
+export async function buildDebugger(
```
arosien, line 35: Concern: this function seems misplaced (within the extension itself). Couldn't the `sbt` commands be added to the "extensionHost" launch config, so a local launch of the extension would (optionally) ensure the local debugger backend was updated?

debugger/src/main/resources/logback.xml
```diff
@@ -27,7 +27,17 @@
...
+      </encoder>
+  </appender>
+
+  <root level="${LOG_LEVEL}">
```
arosien, line 39: Oh! I found a way to specify defaults: https://logback.qos.ch/manual/configuration.html#defaultValuesForVariables

debugger/src/main/resources/logback.xml
shanedell: Gotcha that would explain the weird stuff I was seeing in the file

debugger/src/main/resources/logback.xml
arosien, line 39:
```suggestion
  <root level="${DAFFODIL_DEBUGGER_LOG_LEVEL:-INFO}">
```

debugger/src/main/resources/logback.xml
```diff
@@ -27,7 +27,17 @@
       <pattern>%date{ISO8601} [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n</pattern>
     </encoder>
   </appender>
-  <root level="DEBUG">
+
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+      <file>${LOG_FILE}</file>
```
arosien, line 32:
```suggestion
      <file>${DAFFODIL_DEBUGGER_LOG_FILE}</file>
```

src/daffodilDebugger/utils.ts
```diff
@@ -78,9 +90,11 @@ export async function runDebugger(
     scriptPath,
     artifact.scriptName,
     createTerminal,
-    shellArgs(artifact.scriptName, serverPort),
+    shellArgs(serverPort),
     {
       DAFFODIL_DEBUG_CLASSPATH: daffodilDebugClasspath,
+      LOG_LEVEL: debuggerLogLevel,
+      LOG_FILE: debuggerLogFilePath,
```
arosien, line 97:
```suggestion
      DAFFODIL_DEBUGGER_LOG_LEVEL: debuggerLogLevel,
      DAFFODIL_DEBUGGER_LOG_FILE: debuggerLogFilePath,
```
Env vars should be namespaced to avoid clobbering.

src/daffodilDebugger/utils.ts
shanedell: So we do have something like that already, however since we run `yarn watch` after the first the extension is ran the `sbt` command will get ran but after it won't unless you kill the `yarn watch` and relaunch the extension.

src/daffodilDebugger/utils.ts
```diff
@@ -78,9 +90,11 @@ export async function runDebugger(
     scriptPath,
     artifact.scriptName,
     createTerminal,
-    shellArgs(artifact.scriptName, serverPort),
+    shellArgs(serverPort),
     {
       DAFFODIL_DEBUG_CLASSPATH: daffodilDebugClasspath,
+      DAFFODIL_DEBUGGER_LOG_LEVEL: debuggerLogLevel,
```
arosien, line 96: Oops my suggestion doesn't match the classpath one, sorry... adjust to match?

package.json
shanedell: I like this. One question though will we then want to make an issue to move `debugServer` to be moved under this `debugger` hierarchy and a `port` section?

src/daffodilDebugger/utils.ts
arosien: Hmm/grrr. I will think more, but I still think this should be in build-space, not extension-space.

src/daffodilDebugger/utils.ts
shanedell: Yeah I will do that with some of the other updates

package.json
shanedell: I am going to do `dfdlDebugger` instead of just `debugger` as it causes some confusion in the code

src/daffodilDebugger/utils.ts
shanedell: @arosien @mbeckerle There is a node package called `nodemon` where I can add a custom watcher for scala,sbt related files and when they are changed run a `sbt universal:packageBin` and then could remove that from the extension. Would that be preferred in this PR?

shanedell: I am going to add it because `nodemon` will only be a dev dependency and makes debugging much easier. I was able to get to watch all the files related to the debugger files

package.json
```diff
@@ -625,6 +649,31 @@
               "trace"
             ],
             "default": "info"
+          },
+          "dfdlDebuggerLogLevel": {
```
arosien, line 653: Concern: I think these schema properties need to match the new hierarchical structure.

arosien: Or at least it is confusing to have them not match, I see these names used [here](https://github.com/apache/daffodil-vscode/pull/694/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R128).

shanedell: So those ones are just used for default values of the hierarchy items. It is the same thing we do with other hierarchy items. It might be something to make a new ticket for this, instead of making these items like we do currently for each hierarchy item individually we just make one object default item that sets all of them at once. Does sound good? Since, this is done in multiple places, it would be good to have a separate PR to fix this everywhere.

shanedell: @arosien Created issue https://github.com/apache/daffodil-vscode/issues/698 to fix these items in a different PR.

arosien (approved):
👍

Issue #594, opened 2023-04-27
Use daffodil-tdml library


michael-hoke:
Use official TDML library from https://mvnrepository.com/artifact/org.apache.daffodil/daffodil-tdml and remove scalaxjc.

arosien:
I think the daffodil-tdml library isn't really for programmatic manipulation of test cases. It is more of an interpreter of the data that executes the test cases.

Instead of using those classes, we can use the ones defined in the XSD files that we generate Scala data structures for, via JAXB, and then define extension methods to manipulate them (append test case, etc.). We can then [marshal](https://www.baeldung.com/jaxb#marshalling---converting-java-object-to-xml) these back to XML to pass to the daffodil-tdml `TDMLRunner` for execution.

nlewis05:
Work was being completed under this PR on the apache-daffodil repo https://github.com/apache/daffodil/pull/1057

hdalsania:
Once the https://github.com/apache/daffodil/pull/1057 merged on DFDL side, We need to import the latest DFDL version and then remove scalaxjc with official TDML library from https://mvnrepository.com/artifact/org.apache.daffodil/daffodil-tdml. This library will only be available in maven after DFDL cut release with above PR.

arosien:
Copied from https://github.com/apache/daffodil/pull/1057#issuecomment-2151100868:

Closing for multiple reasons:
- TDML XML object mapping support isn't used, or can be avoided, in downstream Scala projects like [daffodil-vscode](https://github.com/apache/daffodil-vscode). See [above](https://github.com/apache/daffodil/pull/1057#discussion_r1484667731) for more context.
- The original JAXB solution was deemed to have too particular build configuration given differences between JVM versions, so scalaxb was chosen instead. But since JVM 21 was introduced it also appears to have JVM version dependent build configuration, which is slightly annoying. It's not a blocker, however.
- Given the oddities of XML object binding, the need for an alternative was proposed [above](https://github.com/apache/daffodil/pull/1057#discussion_r1484660294), say, JSON, which may fit more downstream languages like Scala and Typescript.

PR #699, opened 2023-07-12
Allow JVM reflection to fix JDK 17 changes.

### LINKED ISSUE #676: Debugger issues in Java 17
As noted by @mbeckerle inside of issues:

- https://github.com/apache/daffodil-vscode/issues/648
- https://github.com/apache/daffodil-vscode/issues/649

So things are not working properly when using Java 17. This issue is for looking into why that is.

arosien:
Allow JVM reflection to fix JDK 17 changes.
    
Certain reflection (used by JAXB) isn't allowed by default in JDK 17. See https://docs.oracle.com/en/java/javase/17/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-7BB28E4D-99B3-4078-BDC4-FC24180CE82B.
    
Fixes #676.
    
The added JVM flags are ignored by earlier JVMs... except JDK 8 which complains. Ugh.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -106,6 +106,24 @@ lazy val debugger = project
     packageName := s"${name.value}-$daffodilVer"
   )
 
+lazy val javaMajorVersion: Int =
+  System.getProperty("java.version").stripPrefix("1.").takeWhile(_.isDigit).toInt
+lazy val isAtLeastJava17: Boolean = javaMajorVersion >= 17
+/* Workaround: certain reflection (used by JAXB) isn't allowed by default in JDK 17:
+ * https://docs.oracle.com/en/java/javase/17/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-7BB28E4D-99B3-4078-BDC4-FC24180CE82B
+ *
+ * While we can handle this JVM quirk at build time, at runtime we won't know
+ * a user's JVM version. We'll provide documentation and an extension setting
+ * to add these flags to the extension-launched debugger backend.
+ */
+lazy val extraXjcJvmOpts: Seq[String] =
+  if (isAtLeastJava17)
+    Seq(
+      "--add-opens",
+      "java.base/java.lang=ALL-UNNAMED"
+    )
+  else Seq()
+
 lazy val xjcSettings =
   Seq(
     libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -118,6 +118,10 @@ lazy val xjcSettings =
     xjcCommandLine += "-p",
     xjcCommandLine += "org.apache.daffodil.tdml",
     xjcBindings += "debugger/src/main/resources/bindings.xjb",
+    xjcJvmOpts ++= Seq(
+      // workaround: certain reflection (used by JAXB) isn't allowed by default in JDK 17: https://docs.oracle.com/en/java/javase/17/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-7BB28E4D-99B3-4078-BDC4-FC24180CE82B
```
mbeckerle, line 122: Just break the too-long comment lines here and below.

mbeckerle (approved):
+1

build.sbt
arosien: The URL is very long, but I'll make the rest break better.

arosien:
I suspect CI is failing because it uses JDK 8 and that JVM flag isn't recognized for that version.

arosien:
> I suspect CI is failing because it uses JDK 8 and that JVM flag isn't recognized for that version.

Yes, this is the case. Ugh. The flag is needed at build time, to generate the JAXB sources, and also at runtime, when dealing with XML via JAXB.
- At build time we could detect the current JVM version and conditionally add the flags.
- At runtime we can't know what JVM version the user has, so we can document the error and have an extension setting that adds the flags when set.

This adds more evidence that JAXB needs to be removed, a la #594.

scholarsmate:
If we can rip and replace the code that is causing the problems without loosing features, that would be the easiest and preferred path.

We had discussed the possibility of bundling a JRE, or compiling it native using Graal VM (https://lists.apache.org/thread/nh8t4c32ov0c0hrvrp2qo3q84179ngvv).  The discussion on JRE control was overtaken by events since we came up with a "lighter touch" solution to the problem at that time and we didn't want to hold up the release.

scholarsmate (approved):

shanedell (approved):
+1

build.sbt
```diff
@@ -106,6 +106,24 @@ lazy val debugger = project
...
+lazy val isAtLeastJava17: Boolean = javaMajorVersion >= 17
+/* Workaround: certain reflection (used by JAXB) isn't allowed by default in JDK 17:
+ * https://docs.oracle.com/en/java/javase/17/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-7BB28E4D-99B3-4078-BDC4-FC24180CE82B
+ *
+ * While we can handle this JVM quirk at build time, at runtime we won't know
+ * a user's JVM version. We'll provide documentation and an extension setting
+ * to add these flags to the extension-launched debugger backend.
+ */
+lazy val extraXjcJvmOpts: Seq[String] =
+  if (isAtLeastJava17)
```
stevedlawrence, line 120: This can be replaced with `scala.util.Properties.isJavaAtLeast("17")`, which has a lot of extra special case handling.

arosien: Wow, TIL!

PR #702, opened 2023-07-13
Fix hierarchy default values. Update how yarn scripts are ran.

### LINKED ISSUE #698: Fix hierarchy default values
Fix the `package.json` to no longer make a new variable for item that would be in an object variables hierarchy. Instead create one default variable that sets all hierarchy related variabes.

For example:

Data Editor variables to remove

```json
"dataEditorPort": {
  "type": "integer",
  "description": "Editor server default port",
  "default": 9000
},
"dataEditorLogFile": {
  "type": "string",
  "description": "Path to log file for data editor",
  "default": "${workspaceFolder}/dataEditor-${omegaEditPort}.log"
},
"dataEditorLogLevel": {
  "type": "string",
  "description": "Log level for data editor",
  "enum": [
    "error",
    "warn",
    "info",
    "debug",
    "trace"
  ],
  "default": "info"
},
``` 

Data Editor variable to make:

```json
"dataEditor": {
  "type": "object",
  "description": "Configuration for Data Editor. Settings are port, logFile and logLevel",
  "default": {
    "port": 9000,
    "logLevel": "info"
  }
},
```

shanedell:
Fix hierarchy default values. Update how yarn scripts are ran.

- Remove individual default values for all attributes of an object. Instead create a default object.
- Run yarn scripts by calling a function from a TS file instead of doing it inline in the package.json.
  - Created build/scripts/yarn-scripts.ts which holds different methods that are the scripts that can be ran.
  - Added package run-func so it is easier to run the yarn TS script functions and makes the yarn scripts in the package.json look cleaner.
- Add nodemon config file debugger/nodemon.json. nodemon running through TS script seemed to cause issues so using config file instead.

Closes #698

### DIFF
--- a/build/scripts/yarn-scripts.ts
+++ b/build/scripts/yarn-scripts.ts
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-nocheck <-- This is needed as this file is basically a JavaScript script
+//                 but with some TypeScript niceness baked in
+const fs = require('fs')
+const glob = require('glob')
+const nodemon = require('nodemon')
+const concurrently = require('concurrently')
+
+function rmFileOrDirectory(path) {
+  if (fs.existsSync(path))
+    fs.rmSync(path, { recursive: true })
+}
+
+function genVersionTS() {
+  const version = JSON.stringify(require('../../package.json').version).replace(/"/g, "'")
+  fs.writeFileSync('./src/version.ts', `/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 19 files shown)


build/scripts/yarn-scripts.ts
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env node
```
stricklandrbls, line 1: Should this go before or after the license blurb?

I've also never seen a _shebang_ for a `.ts` file. Is it necessary?

package.json
```diff
@@ -633,46 +627,23 @@
...
-              "TRACE",
-              "WARN"
-            ],
-            "default": "INFO"
-          },
-          "dfdlDebuggerLogFile": {
-            "type": "string",
-            "description": "Path to log file for debugger",
-            "default": "/tmp/daffodil-debugger.log"
+          "dataEditor": {
```
stricklandrbls, line 630: @scholarsmate I think we'll have to modify the `dataEditorClient.ts` to accommodate this.  

https://github.com/ctc-oss/daffodil-vscode/commit/257ce11c6181dfc097b5d953dc440dfda8323d0d - Looks like we changed the way logging variables were pulled in from the VSCode configuration. I'd assume `logLevel` and `logFile` are similar.

package.json
```diff
@@ -633,46 +627,23 @@
...
+            "type": "object",
+            "description": "Configuration for Data Editor. Settings are port, logFile and logLevel",
+            "default": {
+              "port": 9000,
+              "logFile": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
+              "logLevel": "info"
+            }
```
scholarsmate, line 637: I prefer the way this is done in the `dfdlDebugger` section, so make this:

```
           "default": {
              "port": 9000,
              "logging": {
                "level": "INFO"
              }
            }
```

So we don't have 2 different ways to do the same thing.

package.json
scholarsmate: Right.  If this gets merged, we'll need to update `dataEditorClient.ts` to match.

src/classes/vscode-launch.ts
```diff
@@ -17,20 +17,7 @@
...
-
-export interface InfosetOutput {
-  type: string
-  file: string
-}
-
-export interface LoggingConfig {
-  level: string
-  file: string
-}
```
scholarsmate, line 29: I think the `LoggingConfig` interface is generic enough and should stay here since it can be reused by the other components, like the Data Editor.

scholarsmate (dismissed):
The way logging is configured among the components ought to be standardized.

stricklandrbls (approved):
+1

scholarsmate (approved):
+1

PR #704, opened 2023-07-14
Multiple Launch Wizard updates

### LINKED ISSUE #678: Launch Wizard copying configuration items inconsistent
> One rough spot is creating the 2nd launch config (via the wizard) copies
some settings from the prior config, like daffodilDebugClasspath, but not
everything. I had to tweak the output infoset file to be "file", not
console, again. It should either copy nothing at all, or copy everything
from the prior config.

Mentioned by @mbeckerle in his vote for 1.3.0-rc4.

We want to make sure the launch wizard is consistent where it adds config values from. In this case, if say some part of the configuration is not provided it should always use a default value instead of grabbing from a different config.

### LINKED ISSUE #693: Add debugger config items to Launch Wizard
- Update the Launch Wizard to allow used to set the debugger config items. Currently only logging configs are supported. Below is what the debugger config default looks like:

```json
"dfdlDebugger": {
    "logging": {
        "level": "INFO",
        "file": "/tmp/daffodil-debugger.log"
    }
}
```

shanedell:
Multiple Launch Wizard updates:

- Rename dataEditorConfig attribute to dataEditor to be more consistent.
- Fix issue caused by the tdmlAction being an empty string. When its empty set it to 'none'
- Update setting of dataEditor config for launch wizard to match new config.
- Add dfdlDebugger (logging.level and logging.file) to the launch wizard.
- Clear out daffodil debugger classpath items when switching config.
  - This fixes an issue where the debugger classpath would have duplicate items or just wouldn't be proper for the config.

Closes #678
Closes #693

### DIFF
--- a/src/classes/vscode-launch.ts
+++ b/src/classes/vscode-launch.ts
@@ -30,7 +30,7 @@ export interface VSCodeLaunchConfigArgs {
   infosetFormat: string | null
   infosetOutput: InfosetOutput | null
   tdmlConfig: TDMLConfig | null
-  dataEditorConfig: DataEditorConfig | null
+  dataEditor: DataEditorConfig | null
   stopOnEntry: boolean
   useExistingServer: boolean
   trace: boolean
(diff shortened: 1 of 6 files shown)


mbeckerle (approved):
+1 Some changes suggested, but maybe just comments or renaming.

stricklandrbls (approved):
+1

PR #706, opened 2023-07-18
Un-normalize (capitalize) the drive letter from showOpenDialog if we are on Windows

### LINKED ISSUE #565: TDML Generation on Windows creates bad paths
When using the `TDML Generate` command on Windows, paths are sent to the server with different cases on the drive names in the path. This is causing the absolute to relative path conversion to fail because it is case sensitive.

The data and program paths have lowercase drive names. The data file is populated via an open file dialog, and the program file is populated automatically based on which dfdl schema is opened when the `TDML Generate` command is run.

The TDML path and Infoset path use the default values, which include ${workspaceFolder}. This path is being replaced by the actual workspace path somewhere, and it has an uppercase drive name.

Example TDML File - note the model and document paths:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<ns1:testSuite xmlns:ns1="http://www.ibm.com/xmlns/dfdl/testData" xmlns:ns2="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext" xmlns:ns3="http://www.ogf.org/dfdl/dfdl-1.0/" xmlns:ns4="http://www.ogf.org/dfdl/dfdl-1.0/extensions" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ns6="urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int" suiteName="Default Test Case" defaultRoundTrip="onePass">
    <ns1:parserTestCase name="Default Test Case" root="file" model="..\/c:/Users/nolanm/Documents/GitHub/daffodil-vscode/src/tests/data/txttest.dfdl.xsd" roundTrip="onePass" description="Generated by DFDL VSCode Extension">
        <ns1:document>
            <ns1:documentPart type="file">..\/c:/Users/nolanm/Documents/GitHub/daffodil-vscode/src/tests/data/test.txt</ns1:documentPart>
        </ns1:document>
        <ns1:infoset>
            <ns1:dfdlInfoset type="file">txttest-infoset.xml</ns1:dfdlInfoset>
        </ns1:infoset>
    </ns1:parserTestCase>
</ns1:testSuite>
```

As a workaround, you can use a launch config with ${workspaceFolder} in all 4 paths.

michael-hoke:
Closes #565

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -52,7 +52,19 @@ async function getFile(fileRequested, label, title) {
       })
       .then((fileUri) => {
         if (fileUri && fileUri[0]) {
-          return fileUri[0].fsPath
+          let path = fileUri[0].fsPath
+
+          if (
+            process.platform === 'win32' &&
+            path.length > 2 &&
+            path.charCodeAt(0) > 97 &&
+            path.charCodeAt(0) <= 122 &&
+            path.charAt(1) === ':'
+          ) {
+            path = path.charAt(0).toUpperCase() + path.slice(1)
+          }
+
+          return path
         }
 
         return ''


src/adapter/activateDaffodilDebug.ts
```diff
@@ -52,7 +52,18 @@ async function getFile(fileRequested, label, title) {
...
       .then((fileUri) => {
         if (fileUri && fileUri[0]) {
-          return fileUri[0].fsPath
+          let path = fileUri[0].fsPath
+
+          if (
+            process.platform === 'win32' &&
+            path.charCodeAt(0) > 97 &&
+            path.charCodeAt(0) <= 122 &&
+            path.charAt(1) === ':'
```
scholarsmate, line 61: I'd move this check up 2 lines.

scholarsmate (approved):
+1, 👍

src/adapter/activateDaffodilDebug.ts
michael-hoke: Wouldn't that open us up to a NullPointerException? If fileUri[0] is null, it'll get caught by the if, and this code won't be hit.

scholarsmate: That if statement is nested in another if statement (line 54), so you won't get an NPE.  It might be a good idea to ensure that the length of the path is greater than 2 before peeking at different indexes (e.g, `charCodeAt(0)` and `charAt(1)`) without knowing the size of the path.  I instinctively look at those fail fast conditionals with what filters things the quickest and order those first.  If the second character is not a colon, then it saves you from the next 2 checks.

michael-hoke: I see what you're saying now - adding the length check is definitely a good idea, but I'm not seeing the reason for moving the `charAt(1) === ':'` above the drive letter check.

The path that makes it here comes from a file dialog, which should mean that it has been validated. For the short circuit to save us from the other checks, that validated path would have to have a lowercase ascii character as the first character and something other than a colon as the second character. Is that any more likely than a validated path that has a non-alphabetical character as the first character and a colon as the second?

shanedell (approved):
+1 👍

PR #708, opened 2023-07-19
Multiple updates

### LINKED ISSUE #692: See if packaging scripting can be updated
Look into updating the packaging flow to try using different webpack plugins instead of manually writing TypeScript to do everything

shanedell:
Multiple updates:

- Use webpack for doing most packaging tasks instead of using a manual TS script for everything.
- Update webpack to automatically extract out the debugger package to ./daffodil-debugger-${daffodilVersion}-${extVersion}.
  - This allowed for the code to unzip the debugger in the extension to be removed.
  - This also allows for easier running of the debugger.
  - No longer extract the package out to somewhere else on the user system.
    - The extracted debugger folder with the bin and lib dirs are bundled with the extension source.
- Update scalawatch to build scala code then generate the version.ts, this will ensure the updated debugger package gets unzipped.
- Move location of webpack files as now there are two.
  - webpack/ext-dev.webpack.config.js - used when running locally.
  - webpack/ext-package.webpack.config.js - used when creating a package.
- Remove build/scripts directory.
- Move build/scripts/yarn-scripts.ts to build/yarn-scripts.ts
- Remove .vscodeignore, gets automatically created during the packaging to ignore node_modules.
- Remove Backend class and archiveUrl method from Artifact class
- Remove execSync commands from yarn-scripts.ts

Closes https://github.com/apache/daffodil-vscode/issues/692
Closes https://github.com/apache/daffodil-vscode/issues/711


src/classes/artifact.ts
```diff
@@ -36,14 +37,7 @@ export class Artifact {
       ? `${type}-${this.version}-${LIB_VERSION}`
       : `${type}-${this.version}`
     this.archive = `${this.name}.zip`
+    this.folder = `${this.name}`
     this.scriptName = osCheck(`${baseScriptName}.bat`, `./${baseScriptName}`)
   }
-
```
arosien, line 41: FYI I filed https://github.com/apache/daffodil-vscode/issues/711 which is related to these lines.

shanedell: @arosien Just pushed an update to also delete the `Backend` class and linked to close issue #711

build/yarn-scripts.ts
```diff
@@ -67,19 +83,51 @@ function scalaclean() {
...
+# limitations under the License.
+
+**/node_modules/**/*
+!node_modules/@omega-edit/server/bin
+!node_modules/@omega-edit/server/lib
+`
+  )
+
+  execSync('yarn install', { cwd: pkg_dir })
+  execSync('yarn vsce package --out ../../', { cwd: pkg_dir })
```
scholarsmate, line 124: `yarn` running `yarn-scripts` that then run `yarn` is headache inducing.  I'm not a big fan of this, even if it means we have a longer `yarn scripts` `package` command.  Having `yarn` call these functions that then in turn call `yarn` again doesn't feel right to me.  I'd prefer that we don't do this and instead have a more complex `package` command.  It would be easier to track what's going on there.

package.json
```diff
@@ -24,23 +24,22 @@
...
     "watch:svelte": "cd src/svelte && yarn dev",
-    "webpack": "webpack --mode production --config ./build/extension.webpack.config.js",
-    "prepackage": "yarn install && yarn compile && yarn webpack",
-    "package": "yarn package-setup && yarn package-create",
-    "package-setup": "run-func build/scripts/package.ts setup",
-    "package-create": "run-func build/scripts/package.ts create",
+    "webpack": "webpack --mode production --config ./webpack/ext-dev.webpack.config.js",
+    "webpack:pkg": "webpack --mode production --config ./webpack/ext-package.webpack.config.js",
+    "prepackage": "yarn install && yarn compile && yarn webpack:pkg",
+    "package": "run-func build/yarn-scripts.ts package",
```
scholarsmate, line 42: Put the `yarn` calls at the end of that `package` function here in this line instead of calling back to `yarn` and creating a cycle.

build/yarn-scripts.ts
shanedell: I mean this would depend on what others think, if others agree we should revert it back we can as this was approved before this PR. This would need to be a separate issue to resolve and would not be a part from this PR. @mbeckerle @arosien @stevedlawrence @nlewis05 @michael-hoke @stricklandrbls @rthomas320 what do you guys think?

package.json
shanedell: As mentioned https://github.com/apache/daffodil-vscode/pull/708#discussion_r1270653896 this may need to be discussed before changing it. Or are you only saying to change this for the package yarn commands and all the others are fine?

scholarsmate (changes requested):
Overall, I like the idea of the supplemental `yarn` scripts.  Just prefer them to be **acyclical** even if it means a longer command in the `scripts` section.

build/yarn-scripts.ts
scholarsmate: I'm not suggesting we toss the whole thing out, just avoid calling `yarn` in this supplemental scripts file.

package.json
scholarsmate: Just saying that fix the `package` `yarn` script to have **it** call the other `yarn` `commands` instead of calling the `yarn` commands inside of the `yarn-scripts.ts` file

build/yarn-scripts.ts
shanedell: That makes sense but I also think having it all together makes easier to follow what is going on. If split up between running some stuff in the file and other in the scripts section it might get confusing.

package.json
scholarsmate: I think we can then get rid of the `execSync` dependency once this change is done too.

scholarsmate (approved):
+1, 👍

PR #716, opened 2023-07-26
Multiple svelte updates:

### LINKED ISSUE #709: Merge extension and svelte package.json
All the dependencies used of for `src/svelte/package.json` are all dev dependencies. So, these should be added to the main `package.json` and then remove `src/svelte/package.json`. This will allow for us to have a single `package.json` instead of multiple. This will also probably include updating some other root files and removing them from `src/svelte` since they are identical. This will include adding yarn scripts to probably compile svelte code using the root package.json

### LINKED ISSUE #710: Use webpack for svelte instead of rollup
- Update svelte code to be built using `webpack`, like the extension uses, instead of using `rollup`

shanedell:
Multiple svelte updates:

- Merge svelte package.json with root package.json.
- Move svelete code to be built/watched using webpack instead of rollup.
- Remove svelete unneeded files like .prettierrc, yarn.lock, etc.
- Update root to properly lint svelte source.

Closes #709
Closes #710


src/svelte/webpack.config.js
stricklandrbls, line 1: @scholarsmate @shanedell - There's currently a webpack config implemented in the [ctc-oss/implement-highlevel-svelte-components](https://github.com/ctc-oss/daffodil-vscode/tree/implement-highlevel-svelte-components) branch. 

We may need to link up and determine which version to use or whether to merge / intersect the difference when the aforementioned branch is up for PR.

shanedell: @stricklandrbls I believe @scholarsmate mentioned he took a stab at the `webpack` for svelte. But he also mentioned it didn't seem to fully function as desired and had asked me too to look into it and see if I could get it working. So I guess it would depend on what @scholarsmate thinks?

stricklandrbls: Yeah I also tried to implement `webpack` modifications to resolve some of the long `import` file paths but ran into the same issue. Nothing being imported was resolvable. I was very confused ha

src/svelte/webpack.config.js
```diff
@@ -0,0 +1,117 @@
...
+ * limitations under the License.
+ */
+
+//@ts-check
+'use strict'
+
+//@ts-check
+/** @typedef {import('webpack').Configuration} WebpackConfig **/
+
+// const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
```
scholarsmate, line 24: If we're not using this, let's just remove the commented out code.

package.json
```diff
@@ -56,20 +56,31 @@
...
+    "css-loader": "^6.8.1",
+    "css-minimizer-webpack-plugin": "^5.0.1",
     "glob": "8.0.3",
+    "mini-css-extract-plugin": "^2.7.6",
     "mocha": "10.2.0",
     "prettier": "2.8.8",
+    "prettier-plugin-svelte": "^2.9.0",
     "run-func": "^3.0.0",
+    "sass": "^1.57.1",
+    "svelte": "^3.55.0",
```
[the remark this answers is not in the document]
shanedell: Yeah would be probably be something you would want to do in that branch to make sure it works with all the new stuff thats in it.

scholarsmate (approved):
+1, looks good overall.  Perhaps later we can merge together the `webpack.config.js` together and just have a single top-level one.

shanedell:
> +1, looks good overall. Perhaps later we can merge together the `webpack.config.js` together and just have a single top-level one.

This might be possible but also possibly not because the way things are done to built the svelte stuff is slightly different that we do for the extension so that would have to looked into more.

scholarsmate:
No worries, this is already a nice consolidation.  If the webpack consolidation comes at the expense of a bunch of additional complexity, then it's not worth it.

stricklandrbls (approved):
+1

PR #726, opened 2023-08-02
V1.3.1 RC1 Hotfix: Search Result Highlighting

stricklandrbls:
- Fixed data editor issue where search results would not properly highlight if within the current viewport data.


shanedell (approved):
+1 👍

One comment, in the future don't put `v1.3.1 RC1` as that implies a release candidate has been published and this issue is related to that. Which in this case there is no release candidate

scholarsmate (approved):
+1

PR #728, opened 2023-08-03
Data Editor Profiler & Selection Hotfix

stricklandrbls:
- Fixed the non-breaking NaN input detection for the profiler.
- Fixed the TS & Svelte errors / warnings in the profiler.
- Fixed the selection range offsets to match address radix.

### DIFF
--- a/src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
+++ b/src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
@@ -62,7 +62,8 @@ limitations under the License.
     'Selection',
     $selectionDataStore.startOffset,
     $selectionDataStore.endOffset,
-    $selectionSize
+    $selectionSize,
+    $addressRadix
   )
 
   function generate_offset_headers(
@@ -91,9 +92,12 @@ limitations under the License.
     from: string,
     start: number,
     end: number,
-    size: number
+    size: number,
+    radix: RadixValues
   ): string {
-    return `${from} [${start} - ${end}] Size: ${size} `
+    return `${from} [${start.toString(radix)} - ${end.toString(
+      radix
+    )}] Size: ${size.toString(radix)} `
   }
 
   function updateAddressValue(event: Event) {
--- a/src/svelte/src/components/DataMetrics/ByteFrequencyGraph.svelte
+++ b/src/svelte/src/components/DataMetrics/ByteFrequencyGraph.svelte
@@ -22,6 +22,8 @@ limitations under the License.
   import Input from '../Inputs/Input/Input.svelte'
   import { offsetMax } from '../DataDisplays/CustomByteDisplay/BinaryData'
   import { DATA_PROFILE_MAX_LENGTH } from '../../stores/configuration'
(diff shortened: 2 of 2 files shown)


src/svelte/src/components/DataMetrics/ByteFrequencyGraph.svelte
```diff
@@ -138,6 +138,7 @@ limitations under the License.
           const startOffsetTemp = parseInt(e.detail.value)
```
scholarsmate, line 138: the parseInt calls need to be done with respect to the address radix also.  For example `parseInt(e.detail.value, 16)` for hex, etc.

stricklandrbls: Ah so we're not locking the offset to decimal?

src/svelte/src/components/DataMetrics/ByteFrequencyGraph.svelte
```diff
@@ -324,8 +326,8 @@ limitations under the License.
           >&nbsp;&nbsp;End Offset:
           <Input
             id="end-offset-input"
-            placeholder={endOffset}
-            value={endOffset}
+            placeholder={endOffset.toString()}
+            value={endOffset.toString()}
```
scholarsmate, line 330: Again, toString needs the address radix.

src/svelte/src/components/DataMetrics/ByteFrequencyGraph.svelte
```diff
@@ -352,8 +355,8 @@ limitations under the License.
           >&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Length:
           <Input
             id="length-input"
-            placeholder={length}
-            value={length}
+            placeholder={length.toString()}
+            value={length.toString()}
```
scholarsmate, line 359: I suppose the size ought to use the same radix as the address radix.

src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
```diff
@@ -91,9 +92,12 @@ limitations under the License.
...
     start: number,
     end: number,
-    size: number
+    size: number,
+    radix: RadixValues
   ): string {
-    return `${from} [${start} - ${end}] Size: ${size} `
+    return `${from} [${start.toString(radix)} - ${end.toString(
+      radix
+    )}] Size: ${size} `
```
scholarsmate, line 100: The size here probably ought to be in the same radix as the address radix.

src/svelte/src/components/DataMetrics/ByteFrequencyGraph.svelte
scholarsmate: Nope.  We'll use whatever radix the address is in.

scholarsmate:
Data offsets and lengths need to be in the same radix as the Address throughout the Data Editor.

scholarsmate (approved):
+1, 👍

shanedell (approved):
+1

PR #743, opened 2023-08-08
Fixes Editor Encoding Affecting Search Results

### LINKED ISSUE #742: Search Traversal Changes Results on Encoding Change
### Overview
The editor encoding select is affecting the search traversal results when initial search results have been populated.

The editor encoding should not affect the next and previous results in this way. Along with the fact that searching should be **completely** disabled when the input contains an regex error, in accordance with the current edit encoding.

<details><summary><h4>Screenshot</h4></summary>

Below is a screen capture of this event happening. If the user searches for '400' in the `latin-1` encoding then changes the editor encoding to `hex`, the search previous & next buttons conduct a new search based off the `hex` value `0x40`. This occurs when when `400` is not a valid regex search query for `hex`. ( `/^[0-9a-fA-F]*$/` )

Peek 2023-08-08 09-15

</details>

stricklandrbls:
- Fixes issue where changing the editor encoding affects the search query.

Closes #742

### DIFF
--- a/src/svelte/src/components/Header/fieldsets/SearchReplace.svelte
+++ b/src/svelte/src/components/Header/fieldsets/SearchReplace.svelte
@@ -71,7 +71,6 @@ limitations under the License.
   let direction: SearchDirection = 'Home'
   let preReplaceHasPrev: boolean = false
   let justReplaced: boolean = false
-  let searchInputByteLen: number = -1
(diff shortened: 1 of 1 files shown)


rthomas320 (approved):

shanedell (approved):
+1

PR #750, opened 2023-08-08
Fix issue where functionality was missing when using useExistingServer

### LINKED ISSUE #747: Some functionality is lost when using `useExistingServer`
When the `useExistingServer` is set to true we currently don't test for different things like prompting for data and programs files, etc. This needs to be fixed, the only thing that shouldn't happen if the `useExistingServer` variable is set is running the debugger.

shanedell:
Fix issue where functionality was missing when using useExistingServer

- Whenever useExistingServer was set to true a bunch of code didn't get executed.
  - However, the only code that shouldn't have been executed was running the debubbger.
  - This updates fixes that so that only the debugger isn't ran and all other functionality is still available when useExistingServer is set to true.

Closes #747


scholarsmate (approved):
+1

PR #758, opened 2023-08-10
rename commit to apply

### LINKED ISSUE #759: rename commit to apply
Changes made should be "applied" not "committed".

scholarsmate:
closes #759

### DIFF
--- a/src/dataEditor/dataEditorClient.ts
+++ b/src/dataEditor/dataEditorClient.ts
@@ -385,7 +385,7 @@ export class DataEditorClient implements vscode.Disposable {
         }
         break
 
-      case MessageCommand.commit:
+      case MessageCommand.applyChanges:
         await edit(
           this.omegaSessionId,
           message.data.offset,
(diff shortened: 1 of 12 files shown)


src/dataEditor/dataEditorClient.ts
```diff
@@ -385,7 +385,7 @@ export class DataEditorClient implements vscode.Disposable {
         }
         break
 
-      case MessageCommand.commit:
+      case MessageCommand.apply:
```
stricklandrbls, line 388: Given that we're going to change the term, I think "Apply" would suffice for the button label but maybe we should make the source code a little more verbose?

`MessageCommand.applyEdits` or `applyChangesToViewport`

I just wonder if we'll end up having to change the term in the source if we choose a generic `apply`.

scholarsmate: It's not just applying edits to the viewport, but to the edit model in Ωedit.

scholarsmate: Your requested change has been made to the message enum.

shanedell (approved):
+1

PR #765, opened 2023-08-10
profile graph can be in linear or log scale

### LINKED ISSUE #763: data profile graph to support linear and logarithmic scales
Currently the data profiler graph is using a linear scale and the low frequency counts are hardly visible among the higher frequencies.  A logarithmic scale option would help visualize the low frequency counts as well.

scholarsmate:
closes #763


stricklandrbls (approved):
+1

shanedell (approved):
+1

PR #771, opened 2023-08-10
fix selected offsets displayed in the data editor

### LINKED ISSUE #769: data inspector sometimes shows the wrong offset
Once the underlying viewport gets refreshed, this offset in the data inspector is relative to the viewport, not from the start of the file like it ought to be.

Screenshot 2023-08-10 at 2 24 27 PM

scholarsmate:
fixes #769


scholarsmate:
@stricklandrbls is working on a fix so we're not double accounting for the viewport geometry.

scholarsmate:
This is OBE with #777.

PR #776, opened 2023-08-10
Support JDK 17 + JAXB reflection at runtime.

### LINKED ISSUE #745: JDK 17 not supported when running TDML actions
While #699 fixes _builds_ of the debugger backend, adding the JVM reflection flags is missing when the extension _launches_ the debugger and subsequent TDML actions trigger the JVM reflection bug.

The extension needs to either
- detect what JVM version will be used, and add the flags to the command line; or
- add an extension setting, say "Use JDK 17 settings", so the extension can provide the flags.

Until this is available we should have a FAQ entry, or something, that talks about JVM 17 issues like this.

arosien:
Fixes #745.

- JDK 17 is detected at build time to enable the jvm flags during tests.
- Use the locate-java-home npm module to detect the latest JVM version from within the extension, honoring `$JAVA_HOME` if set.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -47,6 +47,7 @@ lazy val commonSettings =
     dependencyOverrides ++= Seq(
       "org.apache.commons" % "commons-lang3" % "3.12.0"
     ),
+    fork := true, // needed to pass javaOptions to tests, for example
     licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
     maintainer := "Apache Daffodil <dev@daffodil.apache.org>",
     organization := "org.apache.daffodil",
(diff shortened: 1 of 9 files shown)


scholarsmate (approved):
+1

shanedell (approved):
+1

scholarsmate:
Conflicts resolved.  Now they need to be squashed.

shanedell:
@scholarsmate @stricklandrbls There seems to be an issue with `clearInterval` in this PR, do either of you know why?

scholarsmate:
@shanedell, there is a type disagreement on the timer.  Wonder why it's affecting this PR and not happening elsewhere?  I'll look more into this Timer.

arosien:
> 

I think it is from a bad merge with yarn.lock. I'll try to fix.

arosien:
> 

Not that, but I can fix the typescript types real quick `Timer => Timeout`.

PR #778, opened 2023-08-11
Use jsonc-parser instead of JSON.parse

### LINKED ISSUE #635: .vscode/launch.json is not strict JSON
Cannot have a this comment block:
https://github.com/apache/daffodil-vscode/blob/main/.vscode/launch.json#L1

Also cannot have trailing commas in arrays like we have here:
https://github.com/apache/daffodil-vscode/blob/main/.vscode/launch.json#L62

shanedell:
Use jsonc-parser instead of JSON.parse

- Update all files to that use JSON.parse to now use 'jsonc-parser'.parse instead.
  - jsonc-parser allows the json being parsed to not be strict json.

Closes #635

### DIFF
--- a/src/infoset.ts
+++ b/src/infoset.ts
@@ -29,7 +29,7 @@ import {
 // Function to display an infomation message that the infoset file has been created
 // If the user wishes to open the file then they may click the 'Open' button
 async function openInfosetFilePrompt() {
-  let config = JSON.parse(JSON.stringify(getCurrentConfig()))
+  let config = getCurrentConfig()
 
   if (config.infosetOutput.type === 'file') {
     let rootPath = vscode.workspace.workspaceFolders
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -57,7 +57,7 @@ export function runCommand(command: string) {
 // Function for checking if config specifies if either the
 // infoset, infoset diff or hex view needs to be opened
 export async function onDebugStartDisplay(viewsToCheck: string[]) {
-  let config = JSON.parse(JSON.stringify(getCurrentConfig()))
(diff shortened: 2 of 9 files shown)


src/utils.ts
```diff
@@ -21,6 +21,7 @@ import * as os from 'os'
 import * as child_process from 'child_process'
 import path from 'path'
 import { VSCodeLaunchConfigArgs } from './classes/vscode-launch'
+import * as JSONC from 'jsonc-parser'
```
scholarsmate, line 24: Looks like you're just using `parse` from from `json-parser`, why `import *` from it?

src/infoset.ts
```diff
@@ -25,11 +25,12 @@ import {
   ensureFile,
   tmpFile,
 } from './utils'
+import * as JSONC from 'jsonc-parser'
 
 // Function to display an infomation message that the infoset file has been created
 // If the user wishes to open the file then they may click the 'Open' button
 async function openInfosetFilePrompt() {
-  let config = JSON.parse(JSON.stringify(getCurrentConfig()))
+  let config = JSONC.parse(JSON.stringify(getCurrentConfig()))
```
scholarsmate, line 33: I don't see the benefit of using JSONC in cases where you're parsing the output of `JSON.stringify` because comments (if there were any) aren't going to survive that operation.

shanedell:
@scholarsmate updated import to only import `parse` from `json-parser`. Also only use it now when reading a file because when using an object that will always be strict json only files will have non strict json I believe.

arosien (approved):
👍

scholarsmate (approved):
+1

PR #779, opened 2023-08-14
Overload DAPSession.abort to log Throwable.

### LINKED ISSUE #746: When debugger can't compile a schema, no error information is shown
#666 fixes #649, but any failure to launch a schema, usually due to schema compilation errors, doesn't show any error information other than the launch request that failed, and the extension shows a "toast" dialog with the text `debugger daffodil.error.request`.

This is the handling line: https://github.com/apache/daffodil-vscode/commit/6c5b0dd09f5aa53acd431d582799a141e88a9921#diff-ec3efaf651af993da06d613b4b4147bbe406f60ad0a09f9dc1c26bb9633d983eR232 and previously the exception was logged and passed along to the client.

We could pass along the exception message to the event sent to the extension, but we should at least still log the entire exception stack trace.

arosien:
Fixes #746.


mbeckerle (approved):
+1

scholarsmate (approved):
+1 LGTM

PR #782, opened 2023-08-15
Fix Button Collapse Bug

### LINKED ISSUE #761: not all buttons are being re-expanded after being shrunk
In the image below, the save buttons and the apply button in the content controls field set are still icon only, though the other buttons have expanded to include their labels.

Screenshot 2023-08-10 at 11 25 37 AM

stricklandrbls:
- Fixes an issue where the button collapse determination was entering an "impossible" state.

Closes #761


scholarsmate (approved):
+1

shanedell (approved):
+1

PR #814, opened 2023-08-15
Implement Byte Insertion Keybinds Hotfix

### LINKED ISSUE #813: No Byte Insertion & Deletion Keybinds
_Related to #783_ 

There are currently no keybinds to handle the insertion of a single byte, other than replace, and deletion.

stricklandrbls:
- Fixes issue where byte insertions did not have associated keybinds.

Closes #813


src/svelte/src/utilities/elementKeypressEvents.ts
```diff
@@ -0,0 +1,64 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+type ElementKeypressEventKey = 'Enter' | 'Delete'
+const MappableKeys: { [k in ElementKeypressEventKey]: true } = {
+  Enter: true,
+  Delete: true,
+}
```
scholarsmate, line 21: So we have this elementKeypressEvents class that really isn't generic as it's specific to Enter and Delete.  Can it be made generic?

stricklandrbls: @scholarsmate I was going to make it completely generic but left it as `Enter` and `Delete` only as I wasn't sure if we wanted to ability to make _all_ keys mappable. 

I guess it doesn't matter much though, outside of the conditional in `dataEditor.svelte` where it checks if any key that triggers the event is a mappable key.

scholarsmate: Yeah, it feels like overkill right now, but I _could_ see it expanding to becoming something more generic that can be used to handle all of our keybinding.  We may also want to allow the user to be able to configure their keybindings at some point down the road like many other editors and products allow.

shanedell (approved):
+1

PR #822, opened 2023-08-21
Fixed Multiple Data Editor Editing / Display Items

### LINKED ISSUE #783: Implement More Intuitive First & Last Byte Insertion
Currently, in the data editor there is no "normal" way to insert a byte as the first or last byte in the displayable viewport. This is because the button overlays are injected into the wrapper position. Since there is no _before_ or _after_ wrapper for first and last, respectively, the user cannot use the associated buttons.

A workaround for this is to either:
- Create a copy of the byte with the available before/after insert functionality, then edit that desired position.
- Include the desired position in a multiple byte edit selection and then append the desired edits to in the data editor viewport.

### LINKED ISSUE #821: Seek / Search / Navigation Traversal Discrepancy
File & viewport traversal needs to have a single point of calculation. Currently, there are discrepancies in the algorithms that determine what the correct offset and line number is needed to be used in order to render the data in the viewports correctly.

The fix for this is to have all children of the top level `dataEditor.svelte` uses the event dispatcher the call `seek(offset)` to handle viewport and file traversal. All the children should be responsible for is determining what offset to pass.

stricklandrbls:
- Fixed issue where binary display radix would still display 16 bytes per row, instead of 8.
- Fixed inability to edit a "non-regular" sized file ( file size < 2 )
- Rerouted all file / viewport UI offset traversal to use toplevel `seek()` function.

Closes #821
Closes #783

### DIFF
--- a/src/svelte/src/components/DataDisplays/CustomByteDisplay/BinaryData.ts
+++ b/src/svelte/src/components/DataDisplays/CustomByteDisplay/BinaryData.ts
@@ -16,8 +16,16 @@
  */
 
 import { SimpleWritable } from '../../../stores/localStore'
-import type { RadixValues } from '../../../stores/configuration'
-import { radixBytePad } from '../../../utilities/display'
+import {
+  NUM_LINES_DISPLAYED,
+  type BytesPerRow,
+  type RadixValues,
+  VIEWPORT_CAPACITY_MAX,
+} from '../../../stores/configuration'
+import {
+  radixBytePad,
+  viewport_offset_to_line_num,
+} from '../../../utilities/display'
 
 export const BYTE_ACTION_DIV_OFFSET: number = 24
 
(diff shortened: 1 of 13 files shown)


shanedell (approved):
+1

arosien:
Binary display radix shows 8 bytes per row ✔️ 

Screen Shot 2023-08-21 at 12 18 17 PM

stricklandrbls:
> I have a 1-byte file with the ASCII character `x`. If I edit that byte, replacing the `x` with a `y`: Screen Shot 2023-08-21 at 12 04 16 PM then "Apply changes" (btw, I had no idea what the icon was), I get the context `yx`: Screen Shot 2023-08-21 at 12 05 15 PM I'm expecting to get only `y`.

Just pushed a commit that I believe fixes that issue and an issue w/ overwrite only while the file size if irregular.

arosien (approved):
> Just pushed a commit that I believe fixes that issue and an issue w/ overwrite only while the file size if irregular.
Confirmed.

PR #828, opened 2023-08-28
Fix Issue of Byte Value 0 to Display Null

### LINKED ISSUE #827: Byte Value Viewport Displaying Null on 0 Value
The conditional that checks whether the byte within the viewport is a _null-byte_ is interpreting the type guard for the byte value `0` as false. This is causing any value of `0` within the viewport to appear null and being inoperable.

stricklandrbls:
Closes #827


scholarsmate (approved):
+1. 👍

arosien (approved):
lgtm

shanedell (approved):
+1

Issue #826, opened 2023-08-28
Allow viewport reframing on selected byte


scholarsmate:
Select a byte, then apply a reframing function that will shift the viewport such that the selected byte is positioned at the first byte location of the row.  This can be used to "snap" fixed-length records into position such that each record begins on a row boundary in the viewport.

hdalsania:
Let's verify the relevancy of this issue as of v1.6.0

stricklandrbls:
I believe this is still relevant. IIRC, this came from a requirement discussed in a previous requirements meeting we had a while back.

lrbarber:
The "Seek to Offset" option ensures that the requested address is on the top line of the display (unless it's near the end of the file), but it doesn't bring it to the left edge of the window -- instead, the left most position remains a multiple of 16.

scholarsmate:
Punt this to v1.7.0 as it can come in with the OmegaEdit hex editor and save us from doing this work twice.

Issue #784, opened 2023-08-15
Continue improving persisting selections in viewports


nlewis05:
Create a MVP of different kinds of selections, ex: breakpoints, Search and replace, and multiple byte editing mode 

#148

PR #837, opened 2023-09-01
Implemented Relative Seek Offset Traversal

### LINKED ISSUE #799: Ability to navigate to a certain relative byte index
Move over a certain number of bytes over, rather than seek to offset

ex:
type 172 and then hit go and it will move the cursor that many bytes over (positive for seeking forward and negative for seeking backward)

stricklandrbls:
- Added Relative and Absolute seek offset determination and traversal.

Closes #799

### DIFF
--- a/src/svelte/src/stores/index.ts
+++ b/src/svelte/src/stores/index.ts
@@ -22,6 +22,7 @@ import { derived, writable } from 'svelte/store'
 import { SimpleWritable } from './localStore'
 import { ErrorComponentType, ErrorStore } from '../components/Error/Error'
 import {
+  line_num_to_file_offset,
   radixBytePad,
   regexEditDataTest,
   validateEncodingStr,
@@ -32,6 +33,7 @@ import {
   type ByteValue,
 } from '../components/DataDisplays/CustomByteDisplay/BinaryData'
 import {
+  OffsetSearchType,
   ReplaceQuery,
   SearchQuery,
 } from '../components/Header/fieldsets/SearchReplace'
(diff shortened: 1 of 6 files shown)


src/svelte/src/stores/index.ts
```diff
@@ -218,14 +232,38 @@ export const selectionSize = derived(
...
+    $addressRadix,
+  ]) => {
+    if ($seekOffsetSearchType === OffsetSearchType.ABSOLUTE) {
+      return $seekOffsetInput.length > 0
+        ? Math.max(0, parseInt($seekOffsetInput, $addressRadix))
+        : 0
+    } else
+      return (
+        Math.max(0, $dataFeedLineTopOffset) +
+        parseInt($seekOffsetInput, $addressRadix)
```
scholarsmate, line 263: So the relative offsets are relative to the top left offset?

scholarsmate (approved):
+1, 👍 

I think this will work nicely with the reframing feature that's coming up.  That will allow the selected byte to shift left such that the selected byte becomes the first byte of the line.  It ought to just magically work with this because the top left offset ought to shift left by the same amount and this appears to relative to that offset.

src/svelte/src/stores/index.ts
stricklandrbls: That's correct. It's the most concrete calculated offset to base relational offsets from but this will mostly likely be changed in #826 and then I intend to implement some category of highlighting in the in-progress #784 issue.

PR #838, opened 2023-09-06
Implementing Help Component

### LINKED ISSUE #823: Difficult to read data editor "Edit" instructions
Screen Shot 2023-08-21 at 11 58 43 AM

The "Edit" header has a scrollbar to allow reading of the entire text, but the "Edit" header still blocks the top-most text.

(Zoom-out shows the entire text.)

This is based on commit 7318f8 for the 1.3.1 release. I don't think a fix is required for 1.3.1.

stricklandrbls:
- Fixed issue where the editor edit instructions were difficult to read on smaller width/height windows.
- Implemented new `Help` layout for larger/enhanced `Tooltip` implementations.

Closes #823

### DIFF
--- a/src/svelte/src/components/dataEditor.svelte
+++ b/src/svelte/src/components/dataEditor.svelte
@@ -62,6 +62,7 @@ limitations under the License.
   } from './DataDisplays/CustomByteDisplay/BinaryData'
   import { byte_count_divisible_offset } from '../utilities/display'
   import { clearSearchResultsHighlights } from '../utilities/highlights'
+  import Help from './layouts/Help.svelte'
 
   $: $UIThemeCSSClass = $darkUITheme ? CSSThemeClass.Dark : CSSThemeClass.Light
 
@@ -317,6 +318,7 @@ limitations under the License.
     on:seek={seekEventHandler}
   />
 
+  <Help />
   <hr />
   <ServerMetrics />
 </body>
--- a/src/svelte/src/components/layouts/Help.ts
+++ b/src/svelte/src/components/layouts/Help.ts
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+type DescriptionBody = Array<string>
+
+type HelpInformation = {
+  title: string
+  sectionId: AvailableHelpSections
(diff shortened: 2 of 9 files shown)


src/svelte/src/components/layouts/Help.ts
```diff
@@ -0,0 +1,48 @@
...
+  title: 'Edit Modes',
+  sectionId: 'edit-instructions',
+  descriptionBody: [
+    '<h3>Multiple Byte Edits</h3>',
+    "To edit multiple bytes, highlight, by clicking and dragging over, a selection of bytes in either the physical or logical viewports. This will populate the \
+    'Edit' panel with the data to edit.",
+    '<h3>Single Byte Edits</h3>',
+    'To edit a single byte, simply click any singular byte within either of the viewports. This is replace the content with an input field to perform edits.',
+  ],
+}
```
scholarsmate, line 42: Why not export this and allow it to be externally configurable?

src/svelte/src/components/dataEditor.svelte
```diff
@@ -319,6 +320,7 @@ limitations under the License.
     on:seek={seekEventHandler}
   />
 
+  <Help />
```
scholarsmate, line 323: Would be cool if the desired help text could be bound into a generic Help component.

scholarsmate (approved):
+1, this looks good, but I'm getting confused over what we want to be externally configurable and what is specific to our application.  It's always nice to have some generic components that we can easily reuse without having to modify them (open / closed principle states that objects (here components) ought to be open for extension, but closed for modification).

src/svelte/src/components/layouts/Help.ts
stricklandrbls: I figured that instructional information would be pretty concrete so when a developer needs to add a different category for help information then it can be hard coded here.

I didn't see any reason to make it visible outside of this file and using the `AvailableHelpSections` type & object for external use that only allows defined `HelpInformation` objects as an index.

src/svelte/src/components/dataEditor.svelte
stricklandrbls: The `Help` component is "generic" but only displays content from available `HelpInformation` types that are defined in the `AvailableHelpSections`. 

This component uses the contents of the `currentHelpSectionDisplayed` store to determine what is displayed.

NolanMatt (commented):
Nothing Davin didn't already say, looks good

shanedell (approved):
@stricklandrbls rebase and fix the merge conflict and this is good to go. +1

stricklandrbls:
I'll need to test this with the recently merged viewport displays CSS changes to ensure that this still functions as intended.

stevedlawrence:
This PR is from a branch pushed in the main `daffodil-vscode` repo. We try to avoid added new branches since it adds adds alot of extra emails. Instead, please push branches to your fork and open PRs from that.

Also, it might be a good idea to set something like this in your "asf" remote in the daffodil-vscode `.git/config`:

```gitconfig
[remote "asf"]
	url = git@github.com:apache/daffodil.git
	pushurl = do not push to asf
```

This way you can't accidentally push to the asf remote at all since the "pushurl" isn't a valid URL. It's a bit of a hack, but there's no other way do this in git that I know of.

And if you ever do need to push to asf, you can always manually do it like:

```bash
git push git@github.com:apache/daffodil.git ...
```

stricklandrbls:
@stevedlawrence I just realized that this had happened as I was rebasing and pushing changes from the current `main` branch. 

If I remember correctly I tried to use the GitHub feature for creating a branch that automatically is linked to an issue but it must have kept the apache repository as it's base. I typically use the CLI and now will continue to do so to avoid this happening again.

PR #841, opened 2023-09-11
upgrade to Ωedit™ v0.9.73

scholarsmate:
Upgrade from Ωedit™ v0.9.70 to v0.9.73.  Changes made to facilitate Ωedit™'s updates to split mutable (e.g., memory and CPU usage, etc.) and immutable (e.g., server version, JVM version, etc.) server information into two separate calls.


scholarsmate:
This PR took a lot of tinkering to get the dependencies right.  This shifts us from Node 14, to Node 16.  Ωedit™ is now built using Node 16, and the minimum version of VS Code that we're supporting for v1.4.0 has been bumped to 1.67.2 which uses Node 16 internally (up from Node 14 in v 1.60.2).  This PR also pins the versions for all dependencies that we ship with and some of the devDependencies where upgrading caused problems.

shanedell (approved):
+1 👍

src/dataEditor/dataEditorClient.ts
```diff
@@ -94,8 +96,19 @@ const OMEGA_EDIT_MIN_PORT: number = 1024
 // file-scoped types
 // *****************************************************************************
 
+class ServerInfo implements IServerInfo {
+  serverHostname: string = 'unknown' // hostname
```
stricklandrbls, line 100: I've been reading a bit on *stringly-typed* objects in TypeScript and it seems to be best to avoid them
> See **Item 33: Prefer More Precise Alternatives to String Types** in *Effective TypeScript* by Don Vanderkam

Maybe `serverHostname: string` could be `serverHostname: Hostname` where <code>type Hostname = "unknown" | "localhost" | \`{number}.{number}.{number}.{number}.` </code>. Likewise for other

src/svelte/src/components/ServerMetrics/ServerMetrics.svelte
```diff
@@ -20,13 +20,19 @@ limitations under the License.
...
     latency: 0,
-    omegaEditPort: 0,
     serverCpuLoadAverage: 0,
     serverTimestamp: 0,
     serverUptime: 0,
     serverUsedMemory: 0,
-    serverVersion: 'Unknown',
     sessionCount: 0,
+    omegaEditPort: 0,
+    serverVersion: 'Unknown',
```
stricklandrbls, line 29: these `'Unknown'` values are different than the `'unknown'` values in `serverInfo`. Should they be the same?

scholarsmate: Nope.  One is info and the other is setup for displaying in the UI.  The difference in case can also be used to help troubleshoot what side a problem has occurred on (extension side vs. webview side).

src/dataEditor/dataEditorClient.ts
scholarsmate: Seems like that would add a lot of boilerplate to the code for very little benefit.

src/svelte/src/components/ServerMetrics/ServerMetrics.svelte
scholarsmate: serverVersion here could be typed as say "Unknown" | `v${number}.${number}.${number}`.  I say we look more broadly to see where adding these kinds of types would be beneficial as a different issue.

src/dataEditor/dataEditorClient.ts
stricklandrbls: I'd have to look more at the flow of data for this particular instance but just in general I like the idea of strict and defined types over `string` or `number`. There are also cases where you can "type define" yourself into a corner and this might be one of them.

stricklandrbls (approved):
+1

PR #849, opened 2023-09-12
upgrade to node 16 and the min VS Code to 1.66.2

scholarsmate:
Dependencies are dropping support for node 14.  This updates our build to use node 16 and 18 from node 14 and 16.  This also bumps the minimum supported VS Code version to 1.66.2 from 1.60.2.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -90,7 +90,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-11, ubuntu-20.04, windows-2019]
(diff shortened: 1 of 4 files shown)


scholarsmate:
@shanedell, any ideas why CI is failing on the Windows 2019 node 18 test when the node 16 test is working just fine?  The error is that VSCode doesn't work on that version of Windows, which seems bogus to me because the same executable worked on the node 16 test!?

scholarsmate:
Did this update as part of #841.

PR #854, opened 2023-09-20
fix yarn.lock so it correctly matches the deps in package.json

scholarsmate:
The yarn.lock file that is currently in main is incorrect and doesn't match the packages and versions in package.json.  This PR corrects that.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,14 +2,6 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


stricklandrbls (approved):
+1

I am able to build the package and debug successfully. Still trying to figure out why the configurable viewports issue branch won't successfully run debug sessions.

rthomas320 (approved):
Successfully built the package file

PR #856, opened 2023-09-22
update nightly CI to test under node 16 and 18

scholarsmate:
Nightly testing to use node 16 and 18 builds for testing against the latest VS Code stable and insider's builds.  It may not work with node 16 since the latest stable and insider's use node 18 internally.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -26,8 +26,8 @@ jobs:
     name: 'Build, Test, and Package (OS: ${{ matrix.os }}, Node: ${{ matrix.node }}, VS Code: ${{ matrix.vscode }} )'
     strategy:
       matrix:
-        os: [ macos-11, ubuntu-20.04, windows-2019 ]
-        node: [ '14', '16' ]
+        os: [ macos-11, ubuntu-20.04, windows-2019, macos-latest, ubuntu-latest, windows-latest ]
+        node: [ '16', '18' ]
         vscode: [ 'stable', 'insiders' ]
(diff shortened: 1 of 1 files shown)


stevedlawrence:
What do nightlies do differently than PR and Push checks? It looks basically the same as the existing actions. Can we just remove this since ever PR and merge is already checked?

scholarsmate:
It's not the same as the existing actions as the nightlies test the latest stable and insider's versions of VS Code, not just a pinned version (1.67.2).

The nightly workflow has a workflow_dispatch trigger so I was able to run it manually over at the CTC-OSS fork: https://github.com/ctc-oss/daffodil-vscode/actions/runs/6275784426/job/17044528000.

Screenshot 2023-09-22 at 11 22 47 AM

All the MacOS and Linux tests pass, and of course Windows is having some strange problem across all the Windows tests.

stricklandrbls:
This looks good but I'm also dealing with build issues on my Linux box on main. The difference between the CI jobs and my OS is that I'm running a derivation of Ubuntu 22.04 but the CI is _only_ running 20.04.

Is there any reason to not be running `-latest` versions? My build failure on the `main` branch seems like this should be something tested and pretty critical.

scholarsmate:
@stricklandrbls, great point.  We can cover the earliest _and_ latest versions of the OSes.

stricklandrbls (commented):
+1

stricklandrbls (approved):
+1

NolanMatt (approved):
Version updates, looks good

PR #858, opened 2023-09-27
remove debugger server from yarn watch

scholarsmate:
The hot-reload in `yarn watch` seems to be suffering from a race in the debugger server build that can cause file corruption in the server packaging (flagged by unzip).  We don't need to do hot-reloads of the debugger server, just the Svelte and {Java,Type}Script code.


arosien (changes requested):
Concern: I think the `debugger/nodemon.json` file needs to be removed, since it is not referenced anymore.

stricklandrbls (approved):
+1

I was successfully able to run `yarn watch` manually and run the `Extension` debug configuration and start an instance of the data editor. 

#### Minor Concern
Are we able to add some CLI annotation between the [0] & [1] `webpack` compilations? The [1] compilation process takes around 17 seconds and makes it difficult to tell when everything is compiled and ready for starting a data editor instance. _See below and look at the first and last lines of output w/ the milliseconds taken for compilation_

image

</details>

arosien (approved):
👍

PR #862, opened 2023-09-28
add Makefile to create docx documents from the wiki

scholarsmate:
Creates docx files from the wiki.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,0 +1,86 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Requires: make, git, sed, and pandoc
+
+# URL of the wiki repository containing the markdown files
+REPO_URL := https://github.com/apache/daffodil-vscode.wiki.git
+
+# Directory containing the wiki markdown files
+SRC_DIR := .wiki/
+
+# Filter out markdown files starting with underscore
+SRC_MD = $(filter-out $(SRC_DIR)_%, $(wildcard $(SRC_DIR)*.md))
+
+# List of docx and html files to generate from the markdown files
+DOCX = $(addsuffix .docx, $(basename $(notdir $(SRC_MD))))
+HTML = $(addsuffix .html, $(basename $(notdir $(SRC_MD))))
+
+# Function to sanitize filenames
+# 1. Replace non-alphanumeric (except dots and hyphens) with underscores
+# 2. Replace consecutive underscores with a single underscore
+# 3. Replace underscore followed by hyphen with just the hyphen
+sanitize = $(shell echo $(1) | sed -e 's/[^a-zA-Z0-9.\-]/_/g' -e 's/__*/_/g' -e 's/_\-/-/g')
+
+help:
+	@echo "Available targets:"
+	@echo "  all    - Clone/update the repo and generate docx and html files."
+	@echo "  docx   - Generate docx files from the wiki markdown files."
+	@echo "  html   - Generate html files from the wiki markdown files."
+	@echo "  update - Pull updates from the wiki repository (markdown source repo)."
+	@echo "  clean  - Remove generated docx, html files and the cloned wiki."
+
+# Target to clone the git wiki repository containing the markdown files
+$(SRC_DIR):
+	@if ! git clone $(REPO_URL) $(SRC_DIR); then \
+		echo "Failed to clone repository"; \
+		exit 1; \
+	fi
+
+# Target to generate docx from md
+%.docx: $(SRC_DIR)
+	@# Convert img tags to markdown syntax for pandoc
+	sed -r 's!<img[^>]* src="([^"]+)"[^>]*>!\![](\1)!g' "$(SRC_DIR)$(addsuffix .md,$(basename $@))" | pandoc -t docx -f gfm --resource-path=$(SRC_DIR) --toc=true -o "$(call sanitize,$@)" -
+
+# Target to generate html from md
+%.html: $(SRC_DIR)
+	@# Convert img tags to markdown syntax for pandoc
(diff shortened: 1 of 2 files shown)


doc/Makefile
```diff
@@ -0,0 +1,49 @@
...
+# Main target to update the repo and then generate the docx files
+all: update
+	${MAKE} $(DOCX)
+
+# Cleanup target
+clean:
+	rm -rf *.docx $(SRCDIR)
+
+.PHONY: update all clean
+
```
stevedlawrence, line 49: What are your thoughts on committing this to to the wiki repo instead? And treating it just like any other repo that you clone and build artifacts from? Benefit is people can make changes to wiki, clone their forks, and est building a doc without having to change the main wiki first. Not sure how important that is, but it using a normal workflow seems nice.

Also, the release candidate script does this:

```scala
    DOC_NAME="Apache-Daffodil-Extension-for-Visual-Studio-Code-$VERSION"
    git clone https://github.com/apache/daffodil-vscode.wiki.git .wiki
    pandoc -t docx -f markdown -o .wiki/$DOC_NAME.docx .wiki/*.md 
    pandoc -t html -f markdown -o .wiki/$DOC_NAME.html .wiki/*.md 
```
So it's missing the sanitizing step and image stuff. And the makefile doesn't have the ability to create htl.

 If incorporated into the wiki, the release candidate scrit could be updated to just be a `git clone` and `make`?

stevedlawrence: I don't feel strongly, but just some counter-points:

1. It's a Makefile so I would consider it less critical than things like code, or arguably even less critical than wiki content that is much more visible. And since it's hidden and not part of an official release, it shouldn't be a problem with ASF standards. If we feel PRs are needed, which might be good practice, we can always open a PR in this repo or the daffodil repo and copy it to the wiki repo once "approved" and just close the PR
2. That could be solved with documentation page about "Building Wiki to PDF/doc" or something.

I think at least adding HTML so it could be used by the release candidate script would be useful, especially to get the changes you've added and sanitzation and images.

>  just needed the .docx so the documentation can be available on air-gapped networks.

We do publish documentation for each release as part of the release candidate script. Are you try to get more frequent updates?

scholarsmate: I need offline copies for each release in docx format, so I'm not running often (about once every 2 - 3 months).

I've been maintaining this Makefile on my system, but it really needs to live somewhere else with other project assets and be properly maintained by the development team.  I feel this discussion over which repo it should reside is more academic than pragmatic at this point.  I think the most _pragmatic_ (as in least opaque and easily reviewable) place for it to reside in the main project repo under the `doc` directory.

I'll add a target for HTML generation so it can be used in the release CI process so that it can use some of the additional features (and possibly some other fixes/adjustments later on) in the Makefile.

stevedlawrence: > I need offline copies for each release in docx format, so I'm not running often (about once every 2 - 3 months).

Docx and html are already auto generated for each release and are linked at the top right of each release page.

For example, here's the top right of the 1.3.1 release:

https://daffodil.apache.org/vscode/1.3.1/ 

Links to this:

https://daffodil.apache.org/docs/vscode/1.3.1/Apache-Daffodil-Extension-for-Visual-Studio-Code-1.3.1.docx

stevedlawrence (approved):
+1 with HTML add. It would probably be a good idea to update the release candidate script to use this as well if this is going to be the standard way to build wiki docs

scholarsmate:
@stevedlawrence, agreed that we should use the Makefile to generate the HTML in CI.  The HTML generation took a bit of tinkering to get correct for use offline (works online too) viewing.  The `html` target has been developed and pushed.

doc/Makefile
scholarsmate: Yikes!  The documentation on the release page looks pretty gross.  These are just concatenated arbitrarily, broken images, no TOC, and the HTML must be online to fetch external resources.  I think we're better off just providing a link to the wiki documentation.

If we want to provide a single document in HTML and DOCX, we'll need to construct it with greater care than what's being done currently.  Here is a general sketch of how that could be accomplished:

Start with the Brief (Home on the wiki), then strip the header and concatenate the user documentation matching the release version, do the same for the roadmap, and probably not include the developer documentation.  Then use pandoc on that, creating a standone file, generating the TOC, adding a title, and embedding resources.

Doing this is not trivial, it's not impossible, but could be messy and may require updates each time we do a release.  IMHO, the juice isn't worth the squeeze at this point, and prefer the link to the wiki documentation instead of what we currently have.

stevedlawrence: I'm fine dropping the HTML doc. Agreed it doesn't provide much value over the existing wiki since it must still be online.

But are you also suggesting dropping docx from the websie, or does only HTML have the problems you mentioned?

scholarsmate: The docx has the problems too.

If we can just keep the different documents separate (e.g., Brief, User, Requirements/Roadmap, and Developer), that'll be great as that's what this Makefile produces.

Compiling into a single document is where the problems are because you just can't concatenate them like we're doing now.  The files will need to be filtered for the target version (e.g., 1.3.1), ordered (Brief, User, Requirements/Roadmap), transformed (HTML img tags transformed to Markdown image instructions), parsed (headers/footers conditionally kept or removed), then converted (markdown to target output type).

Markdown doesn't support including other Markdown files.  If it did, we could create a Markdown template and include the guts we want, in the order we want them.

stevedlawrence: > If we can just keep the different documents separate (e.g., Brief, User, Requirements/Roadmap, and Developer), that'll be great as that's what this Makefile produces.

What if the docs are just uploaded to separately to a directory and the site just link to a directory? It's not very convenient, but I guess it's better than what we have now.

Or another option, maybe for the future, would be to use use a tool designed to do this. A quick google finds a number of tools that look like they might work. Here's the first one I saw: https://github.com/yakivmospan/github-wikito-converter . Let's not reinvent the wheel if we don't have to.

scholarsmate: I've looked into some of these things already as I certainly agree that we don't want to reinvent anything we have to.

The `wikito` project doesn't handle the _Sidebar.md correctly in that it just includes it, and the links there are not to embedded resources, but are just links to the web.  Like `pandoc` it doesn't handle HTML img tags, so they still need conversion.  We don't want to just take the wiki and create an offline version of it.  We want to be able to generate docs for a given release.  We need more control over version-specific documentation than what `wikito` provides.  Also the last commit for that project was back in 2021, so good luck getting issues resolved or pull-requests merged.

I think a preprocessor would give us the best chance for success.  Something like [markedpp](https://github.com/commenthol/markedpp), but we still need some logic to generate version-specific docs for each release.

Another option is to ditch GitHub Flavored Markdown (gfm) for something more capable like [AsciiDoc](https://asciidoctor.org/news/2013/01/30/asciidoc-returns-to-github/).  This is a huge change and would require converting the existing wiki from `.md` to `.adoc`.

We can explore more comprehensive options, but for right now, I just need to get this Makefile under maintenance and document how offline docs can be generated.  We can tweak and adjust as we go, but it is already useful right now.

stricklandrbls (approved):
+1

PR #874, opened 2023-10-02
Adds hover functionality  for attribute items

rthomas320:
Adds hover text 'no definition available' if the hover item is not an attribute with a definition

### DIFF
--- a/src/language/providers/intellisense/attributeHoverItems.ts
+++ b/src/language/providers/intellisense/attributeHoverItems.ts
@@ -1,0 +1,231 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License Version 2.0
+ * (the "License"; you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export function attributeHoverValues(attributeName: string): string {
+  switch (attributeName) {
+    case 'name':
+      return 'specify name'
+    case 'ref':
+      return 'Specifies the name of an element in this schema'
+    case 'minOccurs':
+      return 'Minimum number of times element will occur'
+    case 'maxOccurs':
+      return 'Maximum number of times element will occur'
+    case 'dfdl:occursCount':
+      return 'dfdl:occursCount property takes an expression which commonly looks in the Infoset via an expression to obtain the count from another element.'
+    case 'dfdl:byteOrder':
+      return 'This property applies to all Number Calendar (date and time and Boolean types with representation binary'
+    case 'dfdl:bitOrder':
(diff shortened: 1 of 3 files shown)


src/language/providers/intellisense/attributeHoverItems.ts
```diff
@@ -0,0 +1,339 @@
...
+      break
+    case 'minOccurs':
+      hoverValue = 'Minimum number of times element will occur'
+      break
+    case 'maxOccurs':
+      hoverValue = 'Maximum number of times element will occur'
+      break
+    case 'dfdl:occursCount':
+    hoverValue = 'dfdl:occursCount property takes an expression which commonly looks in the Infoset via an expression to obtain the count from another element.'
+    break
```
scholarsmate, line 37: Fix indent.  Run `yarn lint:fix`.

scholarsmate (approved):
+1, this looks fine to me.  The `getAttributeHoverProvider` is very verbose.  I'd probably have used a map to lookup the `hoverValue` from the `attributeName` and loaded the map using an external resource file in JSON or CSV.  What you have here gets the job done, so I approve.

src/language/providers/intellisense/attributeHoverItems.ts
```diff
@@ -0,0 +1,339 @@
...
+ * limitations under the License.
+ */
+
+// prettier-ignore
+
+export function attributeHoverValues(attributeName: string): string {
+  let hoverValue: string = ''
+  switch (attributeName) {
+    case 'name':
+      hoverValue = 'specify name'
```
shanedell, line 24: Do we need all the `hoverValue = *` and `break`'s, can you return the string value instead?

Ex:

```
  case 'name':
    return 'specify name'
```

I think that would make this look a bit cleaner, and be less lines, as then each `case` statement only does a `return` instead of doing an assignment plus a `break`. So then you could remove `let hoverValue: string = ''` from the first line of the method and remove `return hoverValue` from the last line of the method.

src/language/providers/intellisense/attributeHoverItems.ts
rthomas320: Replaced "hoverValue=" and break; with return of the string.

rthomas320: When ""plugins": ["prettier-plugin-svelte"]," was added to the prettierrc file, prettier stopped working for typescript. At least on my machine. I removed that line from the prettierrc on my machine and prettier executed on the ts files that were reported as having format problems.

rthomas320: when I run yarn prettier fix -c, I get:
[warn] Code style issues found in 70 files. Run Prettier to fix.
when I run yarn prettier fix, it doesn't change anything.
VS Code show that prettier ran also
image

shanedell: @rthomas320 Don't run `yarn prettier fix`, run `yarn lint:fix`. You may be running a global version of `prettier` instead of the one for `daffodil-vscode` specifically. I also think `yarn prettier fix` is not a command I believe if you want to run the fix manually instead of using `yarn lint:fix` it would be `yarn prettier src -w`, `src` is the directory to look at and `-w` tells prettier to fix the issues it finds.

shanedell: Once all the CI is passing I will add a +1

scholarsmate:
Now run `yarn lint:fix`.

src/language/providers/intellisense/attributeHoverItems.ts
rthomas320: I checked-out the .prettierrc and prettierignore files from main and ran yarn lint:fix.  That fixed the CI check format issues.

shanedell (approved):
+1

PR #875, opened 2023-10-02
Add -no-header flag for consistent JAXB builds

### LINKED ISSUE #840: compiled source matches convenience binary where expected
- The JAXB generated class files in daffodil-debugger jar are different when I build from source. It's not clear to me what is different
   (javap shows no differences) but this makes reproducible builds difficult to verify and is a bit concerning. I don't know why these
   files would be different. Even if generated from JAXB, I would expect it to be deterministic and create the same .class files. Please look into this for the next major release, reproducible builds are important.

- UUID like element gets randomly generated at each build, which would cause the above behavior. Timestamps are another common culprit. 

Source of this ticket is approval VOTE email from @stevedlawrence for v1.3.1-rc2.

michael-hoke:
Closes #840 

Creating as draft because this is most likely only going to be temporarily a problem.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -135,6 +135,7 @@ lazy val xjcSettings =
     xjcCommandLine += "-nv",
     xjcCommandLine += "-p",
     xjcCommandLine += "org.apache.daffodil.tdml",
+    xjcCommandLine += "-no-header",
     xjcBindings += "debugger/src/main/resources/bindings.xjb",
     xjcJvmOpts ++= extraJvmOptions,
     xjcLibs := Seq(


stevedlawrence (approved):
+1 👍 I'm not 100% convinced this will fix the reproducible builds issue, but it certainly won't hurt.

NolanMatt (commented):
Agree with Steve

NolanMatt (approved):

scholarsmate (approved):
+1, LGTM.   I see no harm in merging this.

Issue #797, opened 2023-08-15
Configurable viewport geometries


nlewis05:
It is useful to be able to config the length of how much data when using hexadecimal and binary, screen real estate issue, but at minimum, you should have a divisor, be able to configure to change sizes within other boxes

scholarsmate:
In general, allow the user to modify the viewport geometries (number of rows and columns).

PR #876, opened 2023-10-02
Fixes Alignment of Viewport Header Byte Texts

### LINKED ISSUE #864: Unaligned Viewport Data Byte Offset Text
The viewport byte offset text displayed does not align with the individual bytes of data for systems with higher than average resolution / dpi (2K, 4K, etc..).

Each byte offset set should be individualized similar to how each byte of data, given that the viewport data table headers should always be the same width as the rows of data.

artifact-dfdl-dataeditor

</details>

**Note:** This only seems to happen if the data editor is initialized for data editor `<body>` widths that do not trigger the element collapsing effect (1,500px). This artifact is resolved if initially the window size > 1500px and is then shrunk to trigger the `shouldCollapse` condition. Even if the window is explanded to be larger than the width threshold.

artifact-dfdl-dataeditor-fix

</details>

stricklandrbls:
- Fixed alignment of viewport byte indexing text.

Closes #864

### DIFF
--- a/src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
+++ b/src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
@@ -37,21 +37,14 @@ limitations under the License.
   import { UIThemeCSSClass } from '../../../utilities/colorScheme'
   import { createEventDispatcher } from 'svelte'
   import { OffsetSearchType } from '../../Header/fieldsets/SearchReplace'
+  import { byteDivWidthFromRadix } from '../../../utilities/display'
 
   type ViewportDivSpread = '24px' | '28px' | '68px'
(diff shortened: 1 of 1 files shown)


scholarsmate:
@stricklandrbls, can you post screenshots of the fixed headers?

scholarsmate:
The alignment looks good.  IIRC, the header content used to be bold.  It would be good if the header stood out a bit more from the content.  Maybe a different background color?

shanedell (approved):
+1

stricklandrbls:
This PR had been up for too long and I needed the correction in order to complete development towards #797. That being said I'm going to close this PR as I rebased the branch for #797 on top of this branch so it includes this resolution.

PR #879, opened 2023-10-06
Implemented Configurable Data Viewport Geometry

### LINKED ISSUE #794: Togglable views between logical view and data view only
Either only want the logical view or only the data view in the view port, most of the time you will only need one of these views and data profiling can inform this decision (e.g., does the file only contain printable ASCII?) 

Logical view hide (toggleable view) capability, Mike mentioned that with binary data does not require logical view as it is not byte-oriented data and we can save some screen real estate by hiding logical view.

### LINKED ISSUE #797: Configurable viewport geometries

### LINKED ISSUE #864: Unaligned Viewport Data Byte Offset Text

stricklandrbls:
- Added the ability to hide/show viewports.
- Added the ability to change the amount of bytes per row shown and be independent of the current display radix.
- Added the ability to increment/decrement the amount of viewport lines being displayed.
- Reworked CSS for viewport data tables to move from `display:grid;` to `display:flex;`.
---
### Settings Panel
Two expandable button icons were placed in the _Settings_ fieldset in the main display header. The top icon corresponds to showing / hiding data viewports. The bottom icon corresponds to the amount of bytes in each data line.

<details><summary><h4>Screenshot</h4></summary>

image

</details>

###  Display Line Adjustments
Two buttons were appended to the file traversal button row and separated by a divider as a visual indication of different functionality while still remaining relevant to data traversal. The left is for incrementing the amount of data lines and the right is for decrementing. Both adjustments are by 1 line.
<details><summary><h4>Screenshot</h4></summary>

image

</details>

Closes #864
Closes #797
Closes #794

### DIFF
--- a/src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
+++ b/src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
@@ -18,7 +18,6 @@ limitations under the License.
   import {
     addressRadix,
     displayRadix,
-    editMode,
     seekOffset,
     seekOffsetInput,
     selectionDataStore,
@@ -26,9 +25,9 @@ limitations under the License.
     selectionSize,
     bytesPerRow,
     viewport,
+    visableViewports
   } from '../../../stores'
   import {
-    EditByteModes,
     ADDRESS_RADIX_OPTIONS,
     type RadixValues,
     type BytesPerRow,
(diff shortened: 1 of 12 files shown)


src/svelte/src/components/DataDisplays/Header/DisplayHeader.svelte
```diff
@@ -37,28 +36,20 @@ limitations under the License.
...
-  let offsetLine = []
+  let offsetLine: string[] = []
 
   $: {
     offsetLine = generate_offset_headers(
       $addressRadix,
       $displayRadix,
       $bytesPerRow
     )
+    console.log(offsetLine)
```
scholarsmate, line 52: Nuke this please.

stricklandrbls: I _knew_ that something would get left behind lol

scholarsmate (approved):
+1

shanedell (approved):
+1

scholarsmate:
@stricklandrbls, merge when you're ready.

PR #881, opened 2023-10-19
upgrade Ωedit™ to v0.9.75

scholarsmate:
This upgrades Ωedit™ to v0.9.75 and a few other dependency packages.  The gRPC services provided via Apache Pekko are not available the instant the server comes online.  This patch retries calling `getServerInfo()` each second for up to 60 seconds.  I've tested this quite a bit and most of the time it comes up _nearly instantly_ after the server starts listening, but _on occasion_ it can take a few seconds.  It's difficult to discern why it occasionally takes a few seconds for the services to be ready to handle requests given the asynchronous nature of Pekko in Scala on the JVM.  This is on Java 8 using Pekko 1.0.1.


yarn.lock
```diff
@@ -3760,6 +3792,11 @@ underscore@^1.12.1:
   resolved "https://registry.yarnpkg.com/underscore/-/underscore-1.13.6.tgz#04786a1f589dc6c09f761fc5f45b89e935136441"
   integrity sha512-+A5Sja4HP1M08MaXya7p5LvjuM7K6q/2EaC0+iovj/wOcMsTzMvDFbasi/oSapiwOlt252IqsKqPjCl7huKS0A==
 
+undici-types@~5.25.1:
+  version "5.25.3"
+  resolved "https://registry.yarnpkg.com/undici-types/-/undici-types-5.25.3.tgz#e044115914c85f0bcbb229f346ab739f064998c3"
+  integrity sha512-Ga1jfYwRn7+cP9v8auvEXN1rX3sWqlayd4HP7OKk4mZWylEmu3KzXDUGrQUN6Ol7qo1gPvB2e5gX6udnyEPgdA==
+
```
stevedlawrence, line 3799: Looks like this new dependency got pulled in. Is this a dev dep or does it need to be listed in a LICENSE file?

scholarsmate: It's a dev dep that comes in as part of `@node/types` (see: https://github.com/ctc-oss/omega-edit/blob/main/package.json#L28).

scholarsmate:
@stevedlawrence, just an FYI that I updated the patch to include handling some delay that I've observed between the time the server is listening on the port until the time the gRPC services become available for use (the gRPC services are provided to the Scala server via Apache Pekko).  If the service is unavailable an `UNAVAILABLE` exception is thrown.  This patch will catch that exception, wait a second and try again until it succeeds or it tries more than 60 times and bails out.

scholarsmate:
@stevedlawrence, I believe there are reporting issues on the Ωedit™ side.  There are runtime jars needed by the server (see: https://github.com/ctc-oss/omega-edit/blob/main/server/scala/build.sbt) and they, in turnururntu have transitive dependencies.

PR #883, opened 2023-10-20
Implemented Indexable ByteValue Indications

### LINKED ISSUE #784: Continue improving persisting selections in viewports

stricklandrbls:
- Added indexable indication array for `ByteValue`s within the viewport displays.
- Created categorical byte indiciations values & CSS selectors.

<details><summary><b>Screenshot: <code>ByteValue</code> Search Result Indication</b></summary>

image

</details>
<details><summary><b>Screenshot: <code>ByteValue</code> Replace Result Indication</b></summary>

image
<i>Replacing <code>0x 00 00 00</code> -> <code>0x FF 01 FF</code></i>

</details>
Closes #784

### DIFF
--- a/src/dataEditor/dataEditorClient.ts
+++ b/src/dataEditor/dataEditorClient.ts
@@ -459,11 +459,17 @@ export class DataEditorClient implements vscode.Disposable {
       case MessageCommand.undoChange:
         await undo(this.omegaSessionId)
         await this.sendChangesInfo()
+        this.panel.webview.postMessage({
+          command: MessageCommand.clearChanges,
+        })
         break
 
       case MessageCommand.redoChange:
         await redo(this.omegaSessionId)
         await this.sendChangesInfo()
+        this.panel.webview.postMessage({
+          command: MessageCommand.clearChanges,
+        })
         break
 
       case MessageCommand.profile:
@@ -525,6 +531,9 @@ export class DataEditorClient implements vscode.Disposable {
         ) {
           await clear(this.omegaSessionId)
           await this.sendChangesInfo()
+          this.panel.webview.postMessage({
+            command: MessageCommand.clearChanges,
+          })
         }
         break
 
--- a/src/svelte/src/components/DataDisplays/CustomByteDisplay/DataLineFeed.svelte
+++ b/src/svelte/src/components/DataDisplays/CustomByteDisplay/DataLineFeed.svelte
@@ -31,6 +31,8 @@ limitations under the License.
     seekOffsetInput,
     visableViewports,
     dataDislayLineAmount,
+    replaceQuery,
+    searchResultsUpdated,
   } from '../../../stores'
   import {
     EditByteModes,
@@ -62,10 +64,8 @@ limitations under the License.
     type CSSThemeClass,
   } from '../../../utilities/colorScheme'
   import {
-    selectionHighlights,
-    searchResultsHighlights,
-    updateSearchResultsHighlights,
-    searchResultsUpdated,
+    viewportByteIndicators,
+    categoryCSSSelectors,
   } from '../../../utilities/highlights'
   import { bytesPerRow } from '../../../stores'
   export let awaitViewportSeek: boolean
@@ -169,7 +169,7 @@ limitations under the License.
     bytes: Array<ByteValue>
     highlight: 'even' | 'odd'
   }
-
+  
   enum ViewportScrollDirection {
     DECREMENT = -1,
     NONE = 0,
@@ -181,10 +181,13 @@ limitations under the License.
   let viewportDataContainer: HTMLDivElement
   let selectedByteElement: HTMLDivElement
   let themeClass: CSSThemeClass
-  let activeSelection: Uint8Array
   let lineTopFileOffset: number
-  let searchResults: Uint8Array
+  let makingSelection = false
 
+  $: {
+    makingSelection =
+      $selectionDataStore.startOffset >= 0 && $selectionDataStore.active === false
+  }
   onMount(() => {
     viewportDataContainer = document.getElementById(
       CONTAINER_ID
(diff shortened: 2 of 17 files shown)


stricklandrbls:
There seems to be some unintended merge changes between this branch and main that I did not change myself. These following items are marked as being removed:
- `src/language/dfdl.ts`
- `src/providers/attributeHover.ts`
- `src/providers/attributeHoverItems.ts`

I'm currently working on resolving this.

NolanMatt (approved):
Looks good

stricklandrbls:
I've come across a bug in the `on:click` event propogation while re-basing another ticket to include this PR's changes.

I'm currently working on a fix for this.

scholarsmate:
I hope to get to this review tomorrow.

stricklandrbls:
The issue I ran into was due to the fact that I moved the listener events for the highlight selection. Previously, there were listeners for each `ByteValue` but it was less load on the extension to have a single listener at the `ByteValue` container level, then parse which element fired the event.

There are two separate `on:mousedown` and `on:mouseup` events that are distinct but the child components during a single byte selection only have `on:click` events. This meant that the parent container was still registering the _mousedown_ and preventing a full click/release cycle needed for a _click_.

scholarsmate:
This is a nice new feature.  We can see replaced sequences.

Screenshot 2023-10-31 at 22 32 14

We can select a range of bytes to edit.

Screenshot 2023-10-31 at 22 34 15

We can select a single byte to edit in-line.

Screenshot 2023-10-31 at 22 34 41

One minor inconsistency is that we we can close the multi-byte selection using the X in the 3rd column, but we lack that for the single byte selection.  To "escape out of" single byte mode edit, ESC works, but doesn't "escape out of" multi-byte edit mode.  It would be good to have ESC and the X work for both selection modes.

stricklandrbls:
> Things get more complicated when changes are made, then undone. The highlights are now incorrect when changes have been undone. It would be good to have the highlights work with undo/redo.

The issue with the undo / redo seems to be within the fact that there's no longer a history of changes being kept with messages between the extension, server, and UI. I believe this was changed with the search & replace got restructured to be incremental search & replace.

scholarsmate:
@stricklandrbls, that's right.  If a 1M file of 'a's gets changed to 'xX' that's a lot to keep track of and will not scale well.  I'd recommend doing the same here.  Just highlight the last replacement and the current search.  There will be no highlighting for replace all or replace rest since that's all done on the server side and is opaque on the client side.   

If I insert or delete bytes the highlights float on top and do not float with the content.  If I delete or insert bytes in the middle of a selection, it doesn't shrink or grow.  If I revert all changes, the highlights remain.

Maybe the "fix" is to clear the selections when any edit is made or any change is undone/redone.  I think it's still possible to show the last replacement and the current search, but any other change to the content clears the selections.

stricklandrbls:
>  ... If I delete or insert bytes in the middle of a selection, it doesn't shrink or grow. If I revert all changes, the highlights remain.

Can you elaborate on this one? I was able to make a selection and have the highlights grow & shrink according to the length of the edited segment.
shrink-grow-highlights

scholarsmate:
@stricklandrbls, I was deleting and inserting into the highlights from a search and replace.  That's pretty awesome how it works with the selection for editing.

stricklandrbls:
> I need those 3 bytes replaced with the new bytes (growing the file by 7 bytes), not just overwrite 10 bytes with 10 new bytes.

Is that not fundamentally how the `applyChanges` event works? Even on the extension side the message data members are: 
- `originalDataSegment` - The selected data from the unedited _viewport_ perspective.
- `editedSegment` - The segment of data to replace the original.
- `offset` - Where in the viewport to apply the overwrite.

The highlights are based off the byte length of the  `selectionDataStore`
replace-grow
Here I append `abc` to `PDF` and the replacement results is `PDFabc-1...` growing the file by 3 bytes.

scholarsmate:
@stricklandrbls, these selections should not grow unless we're doing overwrite only.  If we are doing overwrite only, I like the second GIF that shows what _additional_ bytes will be overwritten if the changes are applied.

`applyChanges` by default _effectively_ deletes the original segment, and inserts the new one.  If overwrite only is used, it _effectively_ overwrites the bytes starting at the given offset for the length of the edited segment.  I say _effectively_ because Ωedit™ has an algorithm to change one segment into another using the smallest change possible.

You _could_ use Ωedit™ [viewports](https://github.com/ctc-oss/omega-edit/wiki#viewports) to power the selections and keep them updated when changes are made.  They are designed to track changes in data segments from a content perspective (use `isFloating = true`) or from a position perspective (use `isFloating = false`).  In this case, we want the the viewport to move with changes i the content, so use `isFloating = true`.  Ωedit™ can handle _many_ of these for you and is one of the reasons for having floating viewports.

stricklandrbls:
@scholarsmate 
Is this a highlighting scenario that accurately depicts replacements of bytes `<` or `>` the original selection byte length?

shrink-grow-highlights

stricklandrbls:
The different highlighting for extending / shortening the byte selection via the editor panel, has been removed. Now, once a multiple byte selection has been made, only those bytes will be highlighted in the viewports. This is regardless of the size of the edited data segment.

scholarsmate:
Rebased on the most recent main.

scholarsmate (approved):
+1

src/dataEditor/dataEditorClient.ts
```diff
@@ -459,11 +459,17 @@ export class DataEditorClient implements vscode.Disposable {
       case MessageCommand.undoChange:
         await undo(this.omegaSessionId)
         await this.sendChangesInfo()
+        this.panel.webview.postMessage({
```
arosien, line 462: Question: I don't quite know enough to understand these additions, and to a newbie it seems like one is clearing changes right after the changes are sent, which is slightly confusing. Can you tell me more?

arosien (approved):
👍 

although i have a quick question posted

src/dataEditor/dataEditorClient.ts
stricklandrbls: Id be happy to elaborate but I'm not so sure these lines are related to clearing byte highlights.

PR #884, opened 2023-10-30
Augment xpath/dpath validation

### LINKED ISSUE #850: Augment XPath Validation to be More Complete
When viewing the PCAP schema using the extension, several items are highlighted as errors that we should look to resolve.

- `$dfdl:byteOrder`
- `$pcap:valueLen`
- `fn:concat` with 2 arguments

It looks like `$dfdl:byteOrder` is an external variable - we need to figure out how to identify external variables
It looks like `$pcap:valueLen` is created with a `dfdl:newVariableInstance` tag - we need to parse that to get the variables
We are already able to get variables and insert them into the XPath tokenizer - it's just a matter of figuring out how to augment our current detection with these types of variables.

@mbeckerle are there other cases/examples of XPath values in a schema (e.g outputValueCalc/inputValueCalc) that we are missing?

michael-hoke:
Closes #850 

Adds fn:concat function with 2 arguments

Adds hard-coded/prebuilt variables:
- dfdl:encoding
- dfdl:byteOrder
- dfdl:binaryFloatRep
- dfdl:outputNewLine

Check the prefix of variables, and don't flag them if the prefix is valid. Note that this will also prevent the above variables from being flagged. Do we need to do anything else to differentiate?

### DIFF
--- a/src/language/semantics/dfdlExt.ts
+++ b/src/language/semantics/dfdlExt.ts
@@ -3,19 +3,10 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { EOL } from 'os'
 import * as vscode from 'vscode'
-import {
-  XPathLexer,
-  ExitCondition,
-  LexPosition,
-  Token,
-  BaseToken,
-} from './xpLexer'
-import {
-  XslLexer,
-  GlobalInstructionData,
-  GlobalInstructionType,
-} from './xslLexer'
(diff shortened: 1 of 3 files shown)


michael-hoke:
@mbeckerle Would you mind checking these changes with a couple of schemas? If you don't want to build the branch, I can get you a .vsix - just let me know.

NolanMatt (approved):
Looks completed

shanedell (approved):
+1 Squash your commits and force push and this can merged in once CI passes

PR #893, opened 2023-11-01
add ability to save segments of the file (#800)

### LINKED ISSUE #800: Ability to write selection to a file operation
Use case example: Giant datafile and want to extract a bunch of test cases from it, like a write selection to file operation

scholarsmate:
In addition to downloading a segment profile as CSV, you can now also save that profiled segment itself to a new file.  There may be possible other integration points as well, but since the Data Profiler already works on data segments, the integration here was straightforward.

Screenshot 2023-11-01 at 13 24 26

closes #800


stricklandrbls:
Should the user be allowed to save a segment to the same file that is open in the data editor? If so, should the viewport data automatically reload?

I had a 2KB file and saved a 1KB segment of that file by overwriting the original 2KB file that was currently open in the data editor. I was still able to traverse and edit segments of the original 2KB file but if I just wanted to discard those changes and closed the editor, then only the 1KB segment would remain on disk.

scholarsmate:
@stricklandrbls, ah, nice test!  We have protections like that in place for the regular Save As, but we should also handle that here.

scholarsmate:
I implemented the second option.

PR #897, opened 2023-11-10
Fix check for using default values

### LINKED ISSUE #852: Launch properties that default to true override specific properties set to false
For example, if you set `stopOnEntry` to `false`, the backend will receive a value of `true`. This is because 

https://github.com/apache/daffodil-vscode/blob/1833bf04fd930005e1550ed5bfc657aabb8ddf59/src/utils.ts#L126C1-L131C4

checks the "boolean-ness" of a property, which evaluates to false, so the default value is used instead.

shanedell:
Fix check for using default values

- The check used for seeing if default values or not was working improperly as it was checking the boolean-ness of a variable.
  - So, if the value provided was false it would still use a default value.
  - To fix the issue we need to check if accessing the varaible returns undefined meaning it wasn't provided so use the default then.

Closes #852

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -125,9 +125,10 @@ export function getConfig(jsonArgs: object): vscode.DebugConfiguration {
 
   Object.entries(defaultValues).map(
     ([key, defaultValue]) =>
-      (launchConfigArgs[key] = launchConfigArgs[key]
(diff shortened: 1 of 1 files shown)


scholarsmate (approved):
+1

stricklandrbls (approved):
+1

Issue #878, opened 2023-10-03
Find & Eliminate TypeScript Errors That Allow Compiling


stricklandrbls:
There seems to be a lot of instances where strict types and values are not being implemented and the TS compiler is reporting them as errors in the console. The extension will still compile successfully and build the extension package / run a debug session. If there are errors that are being reported that are actual TS errors then **no compilation should succeed.**

This also makes it difficult to pinpoint areas of code when there is an actual error that's breaking the extension.

<details><summary><h3>Screenshot of TS errors</h3></summary>

![image](https://github.com/apache/daffodil-vscode/assets/30351915/e1900c19-03f9-430f-aa85-b9b9d66ed103)

</details>

**Note:** Errors only seem to be getting reported in VSCode _after_ the TS intellisense has had a change to index through the file. Meaning, errors only get reported after opening a file to edit. I'm assuming this is an artifact from the build system moving completely towards `webpack` and some files were not opened and generated errors.

hdalsania:
This issue may not be applicable anymore as webpack is replaced. Need to retest this for relevancy.  @lrbarber  Could you please test this?

stricklandrbls:
This seems to have been resolved or at least in the PR #1716 I am not seeing any _false positive_ errors.

hdalsania:
This issue is not relevant and we were not able to replicate this. It appears to be related to webpack that we are not using it now. Closing this issue.

PR #917, opened 2023-12-07
Implemented Data Editor Svelte Test Structure

### LINKED ISSUE #915: Utilize Mocha Testsuite For Data Editor TS Objects
The orthogonality of the data editor is becoming more difficult to maintain without the ability to test Svelte components and their stores. This can be significantly improved with the TDD development style.

Mocha is already being utilized within this project and can be utilized to test portions of the data editor that are not Svelte components. The setup is relatively simple:
- Create a `src/svelte/tests` directory & respective `*.test.ts` files.
- Add a `yarn` script in the `package.json:scripts` section to invoke `mocha` to test the `src/svelte/tests` directory. [^1]

There are other testing frameworks that exist specifically to test Svelte components but were much more difficult to implement. For the case of the data editor, the majority of the functionality lies within the objects / classes defined outside of Svelte components. This would make testing w/ Mocha much easier to implement while maintaining the orthogonality of the data used within the data editor.

[^1]: The `mocha` CLI command would need to implement the `-r ts-node/register` option

stricklandrbls:
- Added yarn script to run data editor tests using Mocha.
- Added example function and Svelte store tests.

This is a simple directory structure and test suite setup that utilizes the `mocha` test suite that's currently a utilized dependency. The `test:svelte` script should be ran manually during development and is not currently implemented to be ran during packaging of the extension.

With the `mocha -w` option, it does not currently support _traditional_ sourcing of the directory due a discrepancy with with node. [^1] Typing `r s <Enter>`, while the `test:svelte` script is watching, will manually source the directory structure and re-run the tests. [^2]

[^1]: https://mochajs.org/#nodejs-native-esm-support
[^2]: https://mochajs.org/#-watch-w

Closes #915


package.json
```diff
@@ -41,6 +41,7 @@
     "package": "run-func build/yarn-scripts.ts package && yarn --cwd dist/package install && yarn --cwd dist/package vsce package --out ../../",
     "pretest": "yarn compile && yarn webpack",
     "test": "sbt test && node ./out/tests/runTest.js",
```
shanedell, line 43: I would say that either before or after `node ./out/tests/runTest.js` you want to run the svelte tests. So either,
```json
"test": "sbt test && yarn test:svelte && node ./out/tests/runTest.js",
```
or
```json
"test": "sbt test && node ./out/tests/runTest.js && yarn test:svelte",
```
I personally prefer the first option but either way is fine with me also. Once that is updated I think this is good to go.

stricklandrbls: I initially did not have the tests being ran during the projects tests because there were a lot of compilation errors within the svelte tests. This is because of #878 

Although, I found out that there are only compilation errors when `mocha` is trying to compile the files with the errors. The example tests that I have do not have any so I went ahead and implemented your changes to the test scripts.

stricklandrbls:
@shanedell I added a new script & modified 2 existing scripts:
- Added `watch:svelte-tests` - For having only the svelte tests watched during development.
- Changed `test:svelte` to not have the `-w` option for watching.
- Changed `test` to your first suggestion.

stricklandrbls:
> @shanedell I added a new script & modified 2 existing scripts:
> 
>     * Added `watch:svelte-tests` - For having only the svelte tests watched during development.
> 
>     * Changed `test:svelte` to not have the `-w` option for watching.
> 
>     * Changed `test` to your first suggestion.

Looks like the `mocha` side is still failing on the CI side due to the same compilation errors I mentioned before. I'm not entirely sure why though because running `mocha -r ts-node/register ./src/svelte/tests/**/*.test.ts` or `yarn test:svelte` does not throw those errors on my machine.

nlewis05:
My local build is passing
image

shanedell (approved):
+1

PR #927, opened 2024-01-03
Ensure infoset file extension match output format.

### LINKED ISSUE #845: Create output infoset extension matches the configured format.
Even if you choose output type JSON, the output infoset file is named infoset.xml. Should be infoset.json.

Source of this ticket is approval VOTE [email](https://lists.apache.org/thread/0gcddqd3mhys5179t8zgkpp51l9oc6nv) from @mbeckerle for v1.3.1-rc1 on 08/29/2023

shanedell:
Ensure infoset file extension match output format.

- Done by checking if the filepath does not end with the proper file extension. If it doesn't then:
  - Display warning that the filepath does not end with the output format's file extension and that it will be automatically updated with that file extension.
  - If the path ends with a different extension it is replaced with the extension of the output format.
  - If the path does not end with a different extension then the extension for the output format is added to the end of the file path.

Closes #845

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -21,6 +21,7 @@ import * as os from 'os'
 import * as child_process from 'child_process'
 import path from 'path'
 import { VSCodeLaunchConfigArgs } from './classes/vscode-launch'
+import { InfosetOutput } from './daffodilDebugger'
 
 let currentConfig: vscode.DebugConfiguration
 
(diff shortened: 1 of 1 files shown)


NolanMatt (approved):
Love me some good regex matching, looks good.

src/utils.ts
```diff
@@ -79,6 +80,30 @@ export async function onDebugStartDisplay(viewsToCheck: string[]) {
   })
 }
 
+function checkInfosetFileExtension(
+  infosetOutput: InfosetOutput,
+  infosetFormat: string
+) {
+  // If the infoset file path doesn't end with the infoset output format, update to end with infoset output format.
+  if (!infosetOutput.path.endsWith(`.${infosetFormat}`)) {
```
stricklandrbls, line 88: It looks like there's a couple of other instances where `RegExp` objects are being tested / exec'ed. Exporting this to a function might be helpful for readability and future us.

```ts
let fileExtensionSearchResult = regexTestAndFilter( . . . )
```

src/utils.ts
```diff
@@ -79,6 +80,30 @@ export async function onDebugStartDisplay(viewsToCheck: string[]) {
...
+    let fileExtensionSearchResult = new RegExp('(\\.).*$')
+      .exec(infosetOutput!.path)
+      ?.filter((fileExt) => fileExt !== '.')[0]
+
+    /**
+     * If search result is not undefined replace the file extension with the correct extension.
+     * Else append proper file extension to the end of the output path.
+     */
+    infosetOutput.path =
```
stricklandrbls, line 97: _Same as the above_ [comment.](https://github.com/apache/daffodil-vscode/pull/927/files#r1443342302)

src/utils.ts
shanedell: So this is the only place I believe is filtering off a regex. Other places are using `.exec` and some use `.test`. These perform different as `exec` give you a result and `test` just tells you true or false. I mean we could make wrappers around these but not sure how useful that is as you would just be calling `WrapperFunction` each time instead of `exec` or `test`. We could also make a call back function but that would also kind of just achieve the same thing.

arosien, line 88: Concern: I don't think we should silently change the output file path without notifying the user. Perhaps we can log something or show a popup with a warning?

arosien, line 88: Concern: what happens if `infosetOutput.type` is not `file`, and `path` doesn't exist? Can there be a test for this case?

shanedell: @stricklandrbls I can't really add this change as what I am doing with this regexp is very different from others amd I think adding a function like this would basically involve passing the RegExp and then running the `.test` or `.exec` function which I don't think is completely necessary.

shanedell:
@arosien I added a check to make sure we are only checking the infoset's file extension when the output type is file. I also added a warning message that will let the user know they don't have the proper file extension in the output filepath and that it will updated to automatically have the correct file extension.

arosien (approved):
👍

stricklandrbls (approved):
+1

PR #929, opened 2024-01-05
Rename setting program to schema

### LINKED ISSUE #928: Rename setting from "program" to "schema"
In discussion with @mbeckerle on issue #653 it would be best to rename the setting `program` to `schema`. This ticket is specifically for renaming this setting. This will make it more clear as to what the user needs to point this setting to as `program` is a bit vague. Also, eventually in order to fix the main issue of #653 we will add a new setting called `root`, so having a `root` and `program` seems more confusing than `root` and `schema`, at least to me.

shanedell:
Rename setting program to schema

Closes #928


stricklandrbls (approved):
+1

src/launchWizard/launchWizard.js
```diff
@@ -64,7 +64,7 @@ function getConfigValues() {
     'openInfosetDiffView'
   ).checked
   const openInfosetView = document.getElementById('openInfosetView').checked
-  const program = document.getElementById('program').value
+  const schema = document.getElementById('schema').value
```
mbeckerle, line 67: Hmmm. Can you help me understand why we have both a launchWizard.js and a launchWizard.ts file? I thought ts was the VSCode typed flavor of javascript, and that its use subsumed javascript inside VSCode. 

My understanding is obviously superficial. Can you help me understand why we need both in this situation?

mbeckerle (approved):
+1

This is obviously correct/ok. 

I have to say, the fact that this change required 14 files to be edited was a bit shocking. 

Somehow I feel an opportunity to generate a bunch of artifacts from a common spec has been missed. Maybe not, but that was my visceral reaction.

shanedell:
> +1
> 
> 
> 
> This is obviously correct/ok. 
> 
> 
> 
> I have to say, the fact that this change required 14 files to be edited was a bit shocking. 
> 
> 
> 
> Somehow I feel an opportunity to generate a bunch of artifacts from a common spec has been missed. Maybe not, but that was my visceral reaction. 
> 
> 
> 
> 

Yes, so the reason for the TS and JS files is because the JS file directly interacts with the DOM/UI of the WebView. For the scripting of WebViews it only directly support JS and not TS, so we use JS to interact with UI and then forward that information to the TS. We also use JS to update the UI based on information sent back to the JS from the TS. So, overall JS is used mostly for DOM and UI interaction and the TS does everything else but they communicate back and forth with each other using the VSCode API. The JS version of that API in this case has a lot less capabilities than the VSCode in the TS and that is why we don't use only JS for this as well.

Does that help or make it more confusing?

src/launchWizard/launchWizard.js
shanedell: @mbeckerle Yes, so the reason for the TS and JS files is because the JS file directly interacts with the DOM/UI of the WebView. For the scripting of WebViews it only directly support JS and not TS, so we use JS to interact with UI and then forward that information to the TS. We also use JS to update the UI based on information sent back to the JS from the TS. So, overall JS is used mostly for DOM and UI interaction and the TS does everything else but they communicate back and forth with each other using the VSCode API. The JS version of that API in this case has a lot less capabilities than the VSCode API in the TS and that is why we don't use only JS for this as well.

Does that help or make it more confusing?

shanedell: I also I believe the reason it touched so much is because we based this on [vscode-mock-debug](https://github.com/microsoft/vscode-mock-debug). We mostly followed what they did which I believe starts with a `program` setting. So I believe that was where most of the references came from.

arosien:
@shanedell @mbeckerle FYI there is a mechanism to deprecate configuration settings, which emits a warning, although the deprecated property needs to be supported. See https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema under "deprecationMessage".

It would look like this in our usage:
Screen Shot 2024-01-11 at 2 56 33 PM

PR #933, opened 2024-01-11
Ignore requests for stale state.

### LINKED ISSUE #844: Don't abort when late-arriving event requests old state
If I have the ethernetIP jar on the daffodil debug classpath, only, then when I try to step into the ethernetIP schema I get a failure:
"debugger daffodil.error.scopenotfound", and the parse ends, no infoset is created. I have to put the ethernetIP/src/main/resources directory there as well.
I think it should not fail, and it should complain about not finding the source (though it IS in the jar and eventually we need it to find it there), but it should allow you to *step out of *that back to the PCAP schema file.

Source of this ticket is approval VOTE [email](https://lists.apache.org/thread/0gcddqd3mhys5179t8zgkpp51l9oc6nv) from @mbeckerle for v1.3.1-rc1 on 08/29/2023

arosien:
During 'continue' or 'next' events the frontend makes a sequence of requests to get the current state: asking for threads, the stack frames of each thread, the scopes listing the variables, and the values of the variables. If a rapid sequence of 'continue' or 'next' requests come in during this sequence, the debugger state is advanced and late-arriving requests from previous stepping will fail.

Rather than abort, we ignore the stale requests and log them. The later requests will still request the current state and the frontend will have the correct data.

Fixes #844.

### DIFF
--- a/debugger/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
@@ -368,9 +368,12 @@ class DAPodil(
           _ <- data.stack
             .findFrame(DAPodil.Frame.Id(args.frameId))
             .fold(
-              session.abort(
-                ErrorEvents.ScopeNotFoundError,
-                s"couldn't find scopes for frame ${args.frameId}: ${data.stack.frames.map(f => f.id -> f.stackFrame.name)}"
+              session.sendResponse(
+                request.respondFailure(
+                  Some(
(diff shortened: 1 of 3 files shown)


shanedell (approved):
+1

hdalsania:
@arosien so in order to test this PR, i should step over rapidly while debugging any schema or were you working with any particular schema?

arosien:
> @arosien so in order to test this PR, i should step over rapidly while debugging any schema or were you working with any particular schema?

@hdalsania yes, rapidly press "continue" and/or "step over". you may see intermittent warning messages in the logs, along with the same warning in the "Variables" pane, but the UI will resolve to a consistent state.

hdalsania (approved):
Everything looks good. +1

stricklandrbls (approved):
+1

PR #935, opened 2024-01-11
Added settings for root element name and namespace

### LINKED ISSUE #653: root element for debug must be specified by name and optional namespace with primary schema file from a jar
At Owl we have DFDL schemas made of numerous schema components, where the starting root element of the schema isn't even part of the schema, it's part of a component schema that is in a jar on the classpath. 

This happens when you have the very common envelope-payload idiom. There is a separate schema for the envelope (or header), and a separate schema for the payload.

A third _**assembly schema**_ that combines the envelope and payload together contains only the small glue part that connects the two together. It also contains tests of the assembly schema.  For example it defines an undefined part of the envelope schema to contain an instance of the payload schema.

The root element for parsing/unparsing is the envelope's root element, and that schema is in a jar file. That root element isn't part of the assembly schema at all.

I did not see any feature in launch.json for specifying the name/namespace of the root element. There is just the "program" element, which must be to a file in the current schema. 

The Daffodil library API allows you to specify a root element, and optionally a namespace URI for it. If the root element name is unambiguous the namespace URI is optional. If neither is provided then the primary schema file's first element declaration is assumed to be the root element. 

Right now the VSCode extension is assuming this heuristic is the only way a root element can be specified. 

The root needs to be able to be specified just as element name + optional namespace, where the DFDL schema for that root element name and namespace must be found on the daffodilDebugClasspath. 

The primary DFDL Schema file that contains that element must be able to be a file in a jar on the daffodilDebugClasspath.

shanedell:
Added settings for root element name and namespace

- Created setting "rootName" for specifing the root element's name.
- Created setting "rootNamespace" for specifing the root element's namespace.
- Tied both the settings "rootName" and "rootNamespace" into the extension and the debugger.
- Changed schema setting to be a json object that you can specify "path", "rootName" and "rootNamespace" inside of.

Closes https://github.com/apache/daffodil-vscode/issues/653

### Prepare schemas

To be able to test this PR you will need to test it with the envelope-payload. This requires cloning and building a couple different DFDL Schemas. Below are the commands needed to run locally to get the schemas prepared before we start testing the debugging of the schema.

```bash
for schema in $(echo "ethernetIP PCAP tcpMessage mil-std-2045 envelope-payload"); do
  git clone https://github.com/DFDLSchemas/$schema.git dfdl-$schema
  cd dfdl-$schema
  sbt compile
  sbt publishLocal
  cd ../
done
```

### Debugging envelope-payload

Now that you have built and prepared the needed schemas we can test the debugger. You will run the debugger locally like you normally do, make when the extension launches it is opening the parent directory that you cloned all the schemas to. Once you have the directory with all the schemas opened you will need to make `.vscode/launch.json`, if it doesn't exist already, with this inside of it:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "dfdl",
            "request": "launch",
            "name": "Envelope Payload",
            "schema": {
                "path": "${workspaceFolder}/dfdl-envelope-payload/src/main/resources/io/github/dfdlschemas/envelopepayload/xsd/envelopePayload.dfdl.xsd",
                "rootName": "message"
            },
            "stopOnEntry": true,
            "data": "${workspaceFolder}/dfdl-envelope-payload/src/test/resources/io/github/dfdlschemas/envelopepayload/test_01.dat",
            "infosetFormat": "xml",
            "infosetOutput": {
                "type": "file",
            },
            "debugServer": 4711,
            "openDataEditor": false,
            "openInfosetView": false,
            "openInfosetDiffView": false,
            "daffodilDebugClasspath": "${workspaceFolder}/dfdl-mil-std-2045/src/main/resources:${workspaceFolder}/dfdl-mil-std-2045/target/classes:${workspaceFolder}/dfdl-tcpMessage/src/main/resources:${workspaceFolder}/dfdl-tcpMessage/target/classes:${workspaceFolder}/dfdl-PCAP/src/main/resources:${workspaceFolder}/dfdl-PCAP/target/classes:${workspaceFolder}/dfdl-envelope-payload/src/main/resources:${workspaceFolder}/dfdl-envelope-payload/target/classes:${workspaceFolder}/dfdl-ethernetIP/src/main/resources:${workspaceFolder}/dfdl-ethernetIP/target/classes",
            "variables": {},
            "tunables": {},
            "dataEditor": {
                "port": 9000,
                "logLevel": "info"
            }
        }
    ]
}
```

If you already have a `.vscode/launch.json` at the same level as the schemas you cloned you can just copy the specific configuration. To be specific you will copy from and including opened curly brace (`{`) right after the opened square brace (`[`) to and including the closing curly brace (`}`) before the closing square brace (`]`). Now you want to go to the debug section and select the `Envelope Payload` launch configuration and hit the play button. The debugger should run without at problems and stop you right at in the `dfdl-tcpMessage/src/main/resources/io/github/dfdlschemas/tcpMessage/xsd/tcpMessage.dfdl.xsd` file. This shown in the screenshot below, after clicking continue the debugger should finish running and write the infoset.

Screenshot 2024-01-11 at 4 51 21 PM


debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
```diff
@@ -472,6 +492,28 @@ object Parse {
...
+    // Parse the root name field from the launch config
+    // Defaults to None
+    //
+    // arguments: Launch config
+    def parseRootName(arguments: JsonObject) =
+      Option(arguments.getAsJsonPrimitive("rootName"))
+        .map(_.getAsString())
+        .getOrElse("")
+        .asRight[String]
+        .toEitherNel
```
arosien, line 504: This doesn't actually default to `None`, it defaults to `""`, which I think we should avoid. Instead we can return an `Either[String, Option[String]]` and pass that along, so downstream code doesn't have to check for the empty string. Something like:
```suggestion
    // Defaults to None
    //
    // arguments: Launch config
    def parseRootName(arguments: JsonObject): Either[String, Option[String]] =
      Right(
        Option(arguments.getAsJsonPrimitive("rootName"))
          .map(_.getAsString())
      ).toEitherNel
```

shanedell: @arosien So there a way to make it just `Option[String]` instead of `Either[String, Option[String]]`? I had tried to do that before but it caused errors with the `parMapN` functions

package.json
```diff
@@ -340,6 +340,16 @@
                 "description": "Absolute path to the input data file.",
                 "default": "${command:AskForDataName}"
               },
+              "rootName": {
```
arosien, line 343: Concern: I kind of wish we could refactor the `schema` configuration as an object so that these properties are not all at the "root" level. But that would require breaking existing configuration formats.

Something like:
```json
"schema": {
  "path": ...,
  "rootName": ...,
  "rootNamespace": ...
}
```

shanedell: That makes sense to me. That might be something I could add into this PR, @mbeckerle how would you fell about this?

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
shanedell: When I add this and change all `rootName` to `Either[String, Option[String]]` I get a bunch of errors like below:

```bash
[error]  found   : (java.nio.file.Path, java.nio.file.Path, Boolean, String, org.apache.daffodil.debugger.dap.Parse.Debugee.LaunchArgs.InfosetOutput, Either[String,Option[String]], Either[String,Option[String]], Map[String,String], Map[String,String]) => org.apache.daffodil.debugger.dap.Parse.Debugee.LaunchArgs.Manual
[error]  required: (java.nio.file.Path, java.nio.file.Path, Boolean, String, Product with Serializable with org.apache.daffodil.debugger.dap.Parse.Debugee.LaunchArgs.InfosetOutput, Option[String], Option[String], scala.collection.immutable.Map[String,String], scala.collection.immutable.Map[String,String]) => ?
[error]         ).parMapN(LaunchArgs.Manual.apply)
```

Which are similar errors I got when trying to just do `Option[String]` and was the reason I went with `String` then just to do that matching before passing to `compileFile`

arosien: The signature of `LaunchArgs.Manual.apply` needs to be updated from taking `String` to taking `Option[String]`.

arosien: (same changes for `parseNamespace` since it is also optional)

arosien: If you can't get it I'll push something up.

shanedell: @arosien So I tried that but I am getting these errors:

```bash
[error] /Users/sdell/apache/daffodil-vscode/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala:295:18: could not find implicit value for parameter p: cats.NonEmptyParallel[[+A]scala.util.Either[java.io.Serializable,A]]
[error]         ).parMapN(LaunchArgs.Manual.apply)
[error]                  ^
[error] /Users/sdell/apache/daffodil-vscode/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala:326:20: could not find implicit value for parameter p: cats.NonEmptyParallel[[+A]scala.util.Either[java.io.Serializable,A]]
[error]           ).parMapN(LaunchArgs.Manual.apply)
[error]                    ^
[error] /Users/sdell/apache/daffodil-vscode/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala:343:24: could not find implicit value for parameter p: cats.NonEmptyParallel[[+A]scala.util.Either[java.io.Serializable,A]]
[error]               ).parMapN(LaunchArgs.TDMLConfig.Generate.apply)
[error]                        ^
[error] /Users/sdell/apache/daffodil-vscode/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala:358:24: could not find implicit value for parameter p: cats.NonEmptyParallel[[+A]scala.util.Either[java.io.Serializable,A]]
[error]               ).parMapN(LaunchArgs.TDMLConfig.Append.apply)
[error]                        ^
[error] /Users/sdell/apache/daffodil-vscode/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala:371:24: could not find implicit value for parameter p: cats.NonEmptyParallel[[+A]scala.util.Either[java.io.Serializable,A]]
[error]               ).parMapN(LaunchArgs.TDMLConfig.Execute.apply)
[error]                        ^
[error] /Users/sdell/apache/daffodil-vscode/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala:503:9: polymorphic expression cannot be instantiated to expected type;
[error]  found   : [A, AA >: A]cats.data.EitherNel[AA,Option[String]]
[error]     (which expands to)  [A, AA >: A]scala.util.Either[cats.data.NonEmptyList[AA],Option[String]]
[error]  required: Either[String,Option[String]]
[error]       ).toEitherNel
[error]         ^
[error] /Users/sdell/apache/daffodil-vscode/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala:513:9: polymorphic expression cannot be instantiated to expected type;
[error]  found   : [A, AA >: A]cats.data.EitherNel[AA,Option[String]]
[error]     (which expands to)  [A, AA >: A]scala.util.Either[cats.data.NonEmptyList[AA],Option[String]]
[error]  required: Either[String,Option[String]]
[error]       ).toEitherNel
```

package.json
mbeckerle: > That makes sense to me. That might be something I could add into this PR, @mbeckerle how would you fell about this?

Well these 3 things are, worst case, needed to specify the right root element. 

- path is a URI resolvable via the classpath
- rootName is optional, and defalts to the first element declaration in the schema file identified by the path, if it is not specified (or is null), that is if such global element declaration exists. It may not. The root element can actually be defined anywhere in the schema. 
- rootNamespace is only meaningful if rootName is provided, and it is optional unless it is needed because the schema, transitively, has multiple global element declarations matching rootName. Usually it is not needed, because just the rootName is unambiguous, but it is needed in case people are lazy and call their root element "root" or "main" or something very generic like that.
 
I think a DAP service that, given the path, provides the possibilities for name (w/ associated namespace if ambiguous) is needed to provide a decent UI picker/chooser for this. 

But note also, if a DFDL schema gets compiled to binary form, the decision about which root element is being used is over, and is welded into that compiled schema. We definitely do need to work from compiled schemas because schema compilation times are *long* for these large DFDL schemas (5 minutes for the VMF schema which has 800 files and 180K lines of DFDL). 

Short term, we can continue to depend on people just having to know the rootName + namespace when they are needed.

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
arosien: Try merging in this PR: https://github.com/shanedell/daffodil-vscode/pull/120

package.json
shanedell: @mbeckerle So for now the current `rootName` and `rootNamespace` manual settings are okay? But in the future, I will make a new issue for this, we should add a feature where given the schema path the extension can provide user different options to select from for the root element name and namspace?

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
shanedell: So @arosien with your changes what do I send over as `rootNamespace` from the extension to the debugger so that it is `None`. Currently if I send `""` it will be `""` which causes an error as this defaults the namespace to `{No_Namespace}`, trying to send `null` causes an error

shanedell: So this actually only happens if you specify in the launch.json `"rootNamespace": ""` otherwise it works as expected. Maybe this how it should function also in general right based of on how Daffodil is handling it in the backend?

scholarsmate:
Rebase to hopefully fix CI.

hdalsania (approved):
+1

After extensively working with Shane, it is tested out well in Windows and Rocky 9 Linux Platform as expected.

stricklandrbls:
I believe I followed the PR setup instructions correctly but receiving an vscode error info box stating that all the files listed in the `"daffodilDebugclasspath"` do not exist.

> ```
> File or directory: c:\path\to\pr935/dfdl-mil-std-2045/src/main/resources:c:\path\to\pr935/dfdl-mil-std-2045/target/classes:c:\path\to\pr935/dfdl-tcpMessage/src/main/resources:c:\path\to\pr935/dfdl-tcpMessage/target/classes:c:\path\to\pr935/dfdl-PCAP/src/main/resources:c:\path\to\pr935/dfdl-PCAP/target/classes:c:\path\to\pr935/dfdl-envelope-payload/src/main/resources:c:\path\to\pr935/dfdl-envelope-payload/target/classes:c:\path\to\pr935/dfdl-ethernetIP/src/main/resources:c:\path\to\pr935/dfdl-ethernetIP/target/classes doesn't exist
```

I manually verified that all those directory locations did in fact exist. I also didn't received any errors from the previous shell script for cloning & running the `sbt` commands.

#### PR Directory Structure
```
pr935/
│
├ .vscode/
  ├ launch.json
│
├ dfdl-envelope-payload/
  ├ ...
│
├ dfdl-ethernetIP/
  ├ ...
│
├ dfdl-mil-std-2045/
  ├ ...
│
├ dfdl-PCAP/
  ├ ...
│
├ dfdl-tcpMessage/
  ├ ...
```

shanedell:
@stricklandrbls  So for windows you want to use `;` not `:` in the classpath. You may also need to change `/` to `\\`

stricklandrbls:
> @stricklandrbls So for windows you want to use `;` not `:` in the classpath. You may also need to change `/` to `\\`

Okay I'll have to edit my launch config and test that out. Though I feel like that should be parsed accordingly.

I know we're doing a lot of _if Windows then_ already, for example `exportVar()`, could this not be added?

shanedell:
> > @stricklandrbls So for windows you want to use `;` not `:` in the classpath. You may also need to change `/` to `\\`
> 
> 
> 
> Okay I'll have to edit my launch config and test that out. Though I feel like that should be parsed accordingly.
> 
> 
> 
> I know we're doing a lot of _if Windows then_ already, for example `exportVar()`, could this not be added?

So currently we can't do anything for the ; and : difference because doing a replace would cause issues because of C:. So for now the user needs to know to use ; if on Windows and : otherwise. But another issue has been to make the daffodil debug classpath setting a list of strings and then we will join them together using either : or ; based on the OS which I believe would be the solution to what you mentioned.

stricklandrbls:
> ... make the daffodil debug classpath setting a list of strings and then we will join them together using either : or ; based on the OS which I believe would be the solution to what you mentioned.

Yeah I like that. I just feel like from the user's point of view that getting an error because of something like that would be confusing. Or it would be something that's easily overlooked in documentation for setting up a `launch.json`.

I'm lucky enough to have instant messaging w/ you during the day so I can figure out what's wrong ha.

shanedell:
@stricklandrbls Was that able to fix your issue?

stricklandrbls:
> @stricklandrbls Was that able to fix your issue?

Oh I misunderstood, I thought that the issue you mentioned was something that would be implemented. I'll test this out now.

shanedell:
> > @stricklandrbls Was that able to fix your issue?
> 
> 
> 
> Oh I misunderstood, I thought that the issue you mentioned was something that would be implemented. I'll test this out now.

Well the changing of the daffodil debug classpath to a list would be yes. But for right now if you can the separator from : to ; it should fix your initial problem.

stricklandrbls (approved):
+1

With the `:` changed to `;` for Windows it works as described.

nlewis05 (approved):
image

Issue #10, opened 2021-10-08
Open TDML files


jw3:
Custom view that reads a TDML file and shows the available schemas and data.

Original issue https://github.com/jw3/example-daffodil-vscode/issues/44

Issue #596, opened 2023-04-27
Add Testing Integration Support to TDML


michael-hoke:
Investigate integrating TDML Test Cases into VSCode's native Testing Integration tab.

https://code.visualstudio.com/api/extension-guides/testing

michael-hoke:
- Get Test Cases to show up in the Testing Integration Window on the left hand side of VSCode.
- Figure out what duplicates look like in that window. Currently, test case name and description are used to detect duplicates. Add duplicate checking for test cases in a TDML file. This should, at a minimum, flag problems in the diagnostic window when duplicates exist. Duplicates are when the Test Case name and description match. XPath?
    - If it's possible to implement that as a part of the Append command that feels good for users, do it there as well.
- Remove the Execute command from DFDL's command palette
- Ensure that a test case can be executed from the Testing Integration Window

- Is there a way to easily edit the name/description of a test case? Possibly from the Testing Integration Window... Ideally, we want a way to allow the user to easily modify the name/description of an auto-generated test case

hdalsania:
@NolanMatt  Could you please provide an update on this? There is no associated PR for this.

Issue #595, opened 2023-04-27
Create GUI editor for TDML files


michael-hoke:
We should be able to open a TDML file and have the language identified as TDML. We should then be able to move TDML items from the command palette that are locked to the DFDL language into the new TDML language.

michael-hoke:
See here for some naming suggestions to keep in mind when rebuilding TDML support. Also keep in mind that we are starting off implementing the bare minimum of the TDML spec, so we should try to avoid potentially conflicting names of potential future features.

https://github.com/apache/daffodil-vscode/pull/625#discussion_r1187619693

michael-hoke:
- Add 'none' back in as an option for tdmlAction. Keep 'generate' as the default option.
    - Is this right? Is there a scenario where turning off TDML generation makes sense? If there isn't, then leave 'none' out.
- Make tdmlConfig a mandatory part of the launch config
- Change the default location of the generated TDML file to be in /tmp (or OS alternative)
- Add a new command to copy the TDML file from /tmp to a user-defined location
- Remove the Generate TDML command from the command palette
- Create a new 'TDML' language that inherits from XML (should be set up similarly to dfdl language).
- Move Append TDML command to the new TDML language
- Add a command to create zip file in TDML language

michael-hoke:
If time permits, explore the possibility of using the Selenium tests for the new GUI

michael-hoke:
Current issue that we are running into is that when we try to use HTML to render a GUI, we are unable to import anything from the WebView we are using to produce the HTML. Our best guess at the moment is that whatever is responsible for packaging up dependencies to be used dynamically during runtime is not doing it correctly. Second best guess is that we are importing using the wrong path, but the files we are pointing to exist on disk in the location we expect them to exist (might be related to index.d.ts or .jms missing when expected).

We used a guide (https://timheuer.com/blog/resx-editor-for-visual-studio-code/) to base our initial version of a WebView on, but that wasn't working due to the import issue. We tried using that extension (https://marketplace.visualstudio.com/items?itemName=TimHeuer.resx-editor) and it works fine. However, when we try to use the same methods in the DFDL extension, it does not.

NolanMatt:
to add to what Mike said, if we remove all imports from our webview.js file, then it will function as expected, but obviously doesn't do anything because its just dispalying raw hardcoded html at that point since we can't do any xml manipulation without importing xml libraries.

[...]

PR #936, opened 2024-01-15
Tdml custom editor

NolanMatt:
TDML Custom Editor, Draft PR for now


project/Rat.scala
```diff
@@ -30,8 +30,10 @@ object Rat {
       * well.
       */
     file("language/dfdl.json"),
+    file("language/tdml.json"),
     file("language/syntax.json"),
     file("language/syntaxes/dfdl.tmLanguage.json"),
+    file("language/syntaxes/tdml.tmLanguage.json"),
     file("package.json"),
     file("svelte/package.json"),
     file(".prettierrc"),
```
shanedell, line 39: @NolanMatt Add `src/tdmlEditor/media/codicon.ttf` to this file, this is the binary file causing your Rat Check to fail

NolanMatt: @stevedlawrence How do I go about labeling it?

stevedlawrence: In addition to updating the appropriate LICENSE/NOTICE fie, we add a mention in in the license section of the README.md. Or alternatively, we try to use ALv2 or other Category A licensed files instead and avoid the issue entirely.

shanedell:
@NolanMatt Don't use the `yarn.lock` from `main`. You have added other packages in this PR so you need to run an `yarn install` and then commit the updates `yarn.lock` file.

rthomas320 (approved):
Works as expected.

project/Rat.scala
shanedell: Looks like the `codeicon.tff` was removed so I am going to mark this resolved.

shanedell:
This is not working for when I open TDML files from different TDML schemas. All I get is the `TDML Test Cases` header. If I take one our `src/tests/data/` TDML files it seems to work. So I think there is an issue with parsing, I believe this was proved when I tried to Append TDML in an schemas TDML as I got an error in the extensions debug console.

So, I think this needs updated to have more robust parsing so it works for actual schemas. There should also probably be a option to turn this editor off because if this parsing fails the TDML can't be viewed at all.

shanedell:
Is this supposed to close #10 , #595 , #596 ?

scholarsmate (approved):
+1

PR #943, opened 2024-01-18
Properly await killed process during test shutdown.

arosien:
Previous version did this asynchronously, which then attempted and failed to delete files in use by the (still alive) process.


shanedell (approved):
+1

arosien:
Flaky CI tests pass when re-run. Ick. But no deleting file cleanup problems.

scholarsmate (approved):
+1, LGTM

PR #946, opened 2024-01-22
Make CI more similar to (failing) nightly builds.

arosien:
Nightly uses `vscode: [ 'stable', 'insiders' ]`, whereas CI doesn't specify the vscode version, which defaults to the `package.json` vscode engine version, currently set to `^1.67.2`, which gets resolved to `1.67.2`.

However, let's make the builds as similar as possible to provoke the same failures in CI as nightly.

This provokes the same failures as nightly, and the fix is to not attempt to delete the extracted backend debugger directory. This was added to make local development safer, but re-running the tests locally will always delete the extracted backend debugger before re-extracting it, which happens during the webpack step (using webpack/ext-dev.webpack.config.js).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -133,11 +133,14 @@ jobs:
         java_version: [ 8, 11, 17 ]
         os: [ macos-11, ubuntu-20.04, windows-2019 ]
         node: [ '16', '18' ]
+        vscode: [ 'stable' ]
       fail-fast: false  # don't immediately fail all other jobs if a single job fails
     runs-on: ${{ matrix.os }}
     defaults:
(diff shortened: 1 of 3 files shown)


arosien:
Yay I got the same failures as nightly, now I can try to find a proper fix.

NolanMatt (approved):

shanedell (approved):
+1 👍

shanedell:
@arosien For future reference you need to have 2 green approvals before merging.

arosien:
> @arosien For future reference you need to have 2 green approvals before merging.

ah, **green** is the important word here. got it! (there were failed tests due to network timeouts...)

PR #952, opened 2024-01-26
Change daffodilDebugClasspath to array of string instead of string

### LINKED ISSUE #934: Make daffodilDebugClasspath more extensible
Something that I noticed with testing changes for #653 is that the `daffodilDebugClasspath` setting can become incredibly long. I think making this setting a JSON object that will that be looped over and join the values together separated by a colon (`:`).

So for example:

#### Snippet 1

```json
{
    ...
   "daffodilDebugClasspath": "${workspaceFolder}/dfdl-mil-std-2045/src/main/resources:${workspaceFolder}/dfdl-mil-std-2045/target/classes:${workspaceFolder}/dfdl-tcpMessage/src/main/resources:${workspaceFolder}/dfdl-tcpMessage/target/classes:${workspaceFolder}/dfdl-pcap/src/main/resources:${workspaceFolder}/dfdl-pcap/target/classes:${workspaceFolder}/dfdl-envelope-payload/src/main/resources:${workspaceFolder}/dfdl-envelope-payload/target/classes:${workspaceFolder}/dfdl-ethernet/src/main/resources:${workspaceFolder}/dfdl-ethernet/target/classes",
    ...
}
```

Could instead be:

#### Snippet 2
```json
{
    ...
    "daffodilDebugClasspath": {
        "dfdl-mil-std-2045": "${workspaceFolder}/dfdl-mil-std-2045/src/main/resources:${workspaceFolder}/dfdl-mil-std-2045/target/classes",
        "dfdl-tcpMessage": "${workspaceFolder}/dfdl-tcpMessage/src/main/resources:${workspaceFolder}/dfdl-tcpMessage/target/classes",
        "dfdl-pcap": "${workspaceFolder}/dfdl-pcap/src/main/resources:${workspaceFolder}/dfdl-pcap/target/classes",
        "dfdl-envelope-payload": "${workspaceFolder}/dfdl-envelope-payload/src/main/resources:${workspaceFolder}/dfdl-envelope-payload/target/classes",
        "dfdl-ethernet": "${workspaceFolder}/dfdl-ethernet/src/main/resources:${workspaceFolder}/dfdl-ethernet/target/classes",
    }
    ...
}
```

Another thing that we could also do is test if the value passed is a directory, if it is the extension won't add the general directory but would instead add `${DIR_PATH}/{src/main/resources,target/classes}`. Then if say its a file like a JAR it just leaves as is and will join in together with the other items. This would than make the above Snippet 2 to be even simpler by only needing this instead:

#### Snippet 3

```json
{
    ...
    "daffodilDebugClasspath": {
        "dfdl-mil-std-2045": "${workspaceFolder}/dfdl-mil-std-2045",
        "dfdl-tcpMessage": "${workspaceFolder}/dfdl-tcpMessage",
        "dfdl-pcap": "${workspaceFolder}/dfdl-pcap",
        "dfdl-envelope-payload": "${workspaceFolder}/dfdl-envelope-payload",
        "dfdl-ethernet": "${workspaceFolder}/dfdl-ethernet",
    }
    ...
}
```

But in the end with exported variable would still contain all paths needed like Snippet 1 manually defines.

What does everyone think about this? 

@mbeckerle @stevedlawrence @scholarsmate @stricklandrbls @NolanMatt @arosien @lrbarber @nlewis05 @hdalsania @rthomas320

shanedell:
Change daffodilDebugClasspath to array of string instead of string

- This helps prevent issues where users might not be aware to use : for mac/linx and ; for windows.
- Update launch wizard to work properly with an array of strings instead of string.
- Update some other code to not try to split the daffodilDebugClasspath.
- When the classpath is passed to the debugger we join the array of string by either : or ; based on the OS.
- Allow rootName and rootNamespace to be null and allow null to be passed between TS and Scala.
  - Both also default to null.
  - Update properties of schema to require the path and define path, rootName and rootNamespace as properties.

Closes #934

### DIFF
--- a/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
@@ -399,8 +399,10 @@ object Parse {
     // arguments: Launch config
     def parseRootName(arguments: JsonObject) =
       Right(
-        Option(arguments.getAsJsonObject("schema").getAsJsonPrimitive("rootName"))
-          .map(_.getAsString())
+        arguments.getAsJsonObject("schema").get("rootName") match {
+          case null => None
+          case rn   => if (rn.isJsonNull()) None else Some(rn.getAsString())
+        }
       ).toEitherNel
 
     // Parse the root namespae field from the launch config
@@ -409,8 +411,10 @@ object Parse {
     // arguments: Launch config
     def parseRootNamespace(arguments: JsonObject) =
       Right(
-        Option(arguments.getAsJsonObject("schema").getAsJsonPrimitive("rootNamespace"))
-          .map(_.getAsString())
+        arguments.getAsJsonObject("schema").get("rootNamespace") match {
+          case null => None
+          case rn   => if (rn.isJsonNull()) None else Some(rn.getAsString())
+        }
       ).toEitherNel
 
     // Parse the data field from the launch config
--- a/package.json
+++ b/package.json
@@ -333,10 +333,28 @@
               "schema": {
                 "type": "object",
                 "description": "Contains the absolute path to the DFDL schema file, root element name and namespace.",
+                "required": ["path"],
+                "properties": {
+                  "path": {
+                    "type": "string",
+                    "description": "Absolute path to the DFDL schema file",
+                    "default": "${command:AskForSchemaName}"
+                  },
+                  "rootName": {
+                    "type": ["string", "null"],
+                    "description": "Name of the root element",
+                    "default": null
+                  },
+                  "rootNamespace": {
+                    "type": ["string", "null"],
+                    "description": "Name of the root element namespace",
+                    "default": null
+                  }
+                },
                 "default": {
                   "path": "${command:AskForSchemaName}",
-                  "rootName": "",
-                  "rootNamespace": ""
+                  "rootName": null,
+                  "rootNamespace": null
                 }
               },
               "data": {
@@ -407,9 +425,9 @@
                 "default": false
               },
               "daffodilDebugClasspath": {
-                "type": "string",
+                "type": "array",
                 "description": "Additional classpaths to be exported to the debugger",
-                "default": ""
+                "default": []
               },
               "variables": {
                 "type": "object",
@@ -456,8 +474,8 @@
             "name": "Ask for file name",
             "schema": {
               "path": "${command:AskForSchemaName}",
-              "rootName": "",
-              "rootNamespace": ""
+              "rootName": null,
+              "rootNamespace": null
             },
             "data": "${command:AskForDataName}",
             "stopOnEntry": true,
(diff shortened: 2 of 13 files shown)


shanedell:
@stricklandrbls @hdalsania @rthomas320 This should fix your issues you saw with my previous PR https://github.com/apache/daffodil-vscode/pull/935, where for windows you needed to make to use delimiter `;` instead of `:`. Please pull this PR and update your daffodilDebugClasspath to be an array, each folder should be its own entry such as:

```json
"daffodilDebugClasspath": [
    "${workspaceFolder}/dfdl-mil-std-2045/src/main/resources",
    "${workspaceFolder}/dfdl-mil-std-2045/target/classes",
    "${workspaceFolder}/dfdl-tcpMessage/src/main/resources",
    "${workspaceFolder}/dfdl-tcpMessage/target/classes",
    "${workspaceFolder}/dfdl-PCAP/src/main/resources",
    "${workspaceFolder}/dfdl-PCAP/target/classes",
    "${workspaceFolder}/dfdl-envelope-payload/src/main/resources",
    "${workspaceFolder}/dfdl-envelope-payload/target/classes",
    "${workspaceFolder}/dfdl-ethernetIP/src/main/resources",
    "${workspaceFolder}/dfdl-ethernetIP/target/classes"
],
```

stricklandrbls:
The changes to the `daffodilDebugClasspath` work as expected with the `launch.json` from #935 .

Though currently, there is a debug compilation error when launching a debug session w/ a config created from the Launch Wizard. If the `rootName` and `rootNamespace` are not populated in the wizard then they are populated in the `launch.json` with default values of `""`. This causes the debug session to crash upon startup.

Even `undefined` values will cause the same crash. I believe that if they are unpopulated in the Launch Wizard then they should not populate whatsoever in the `launch.json` _or_ the default values need to be corrected.

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
```diff
@@ -399,8 +399,10 @@ object Parse {
     // arguments: Launch config
     def parseRootName(arguments: JsonObject) =
       Right(
-        Option(arguments.getAsJsonObject("schema").getAsJsonPrimitive("rootName"))
-          .map(_.getAsString())
+        arguments.getAsJsonObject("schema").get("rootName") match {
```
shanedell, line 402: This change was needed for both `parseRootName` and `parseRootNamespace`. Without it, passing `null` from the TS to the Scala would cause an error:

```bash
class com.google.gson.JsonNull cannot be cast to class com.google.gson.JsonPrimitive
```

shanedell:
@stricklandrbls The issue you ran into should not be fixed. except instead of using `undefined` we need to use `null`. This was because the `package.json` doesn't allow for the use of `undefined` but it does allow for `null`

rthomas320 (approved):
The change is working as expected for the launch config for PR 935 and for new launch configs

stricklandrbls (approved):
+1 
The recent changes worked.

hdalsania:
This does not work for me if I have set schema and data file path set to "${command:AskForSchemaName}" and  "${command:AskForDataName}" regardless of what schema file and data file I choose, I get following error messages and debugger never starts.


image

PR #964, opened 2024-02-06
Fix error thrown when using ${command:AskForSchemaName}

### LINKED ISSUE #962: Error with using ${command:AskForSchemaName} in launch.json
This error was initially reported by @hdalsania and confirmed by @rthomas320 and @lrbarber. I believe this issue deals with an update that changed `schema` from a `string` to a `object`. What happens is an error is thrown in the debugger that a JsonPrimitive cannot be converted to JsonObject.

shanedell:
Fix error thrown when using ${command:AskForSchemaName}

- The issue dealt with setting `schema` to the result of the command instead of using `schema.path`.

Closes #962

### DIFF
--- a/src/daffodilDebugger/debugger.ts
+++ b/src/daffodilDebugger/debugger.ts
@@ -152,7 +152,7 @@ export async function getDebugger(
 
     // Get schema file before debugger starts to avoid timeout
     if (config.schema.path.includes('${command:AskForSchemaName}')) {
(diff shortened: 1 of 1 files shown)


rthomas320 (approved):
This change fixes the problem where a jsonPrimitive cannot be cast to JsonObject.  I was able to successfully open a schema file and a datafile.

stricklandrbls (approved):
+1

PR #966, opened 2024-02-13
Bump omega-edit to v0.9.78

shanedell:
Bump omega-edit to v0.9.78

- Bump macos-11 to macos-12 in CI.
  - macos-11 is deprecated and omega-edit uses macos-12 to build the lib files.

This new version of omega-edit builds the macOS and Ubuntu arm64 native library through CI. I tested this on my mac M1 both native on the host and then in a Ubuntu 20.04 and Rocky Linux 9 VM. All 3 worked without issue.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -131,7 +131,7 @@ jobs:
       matrix:
         java_distribution: [ temurin ]
         java_version: [ 8, 11, 17 ]
-        os: [ macos-11, ubuntu-20.04, windows-2019 ]
+        os: [ macos-12, ubuntu-20.04, windows-2019 ]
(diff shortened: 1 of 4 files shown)


shanedell:
@scholarsmate I know you mentioned earlier being worried about the java 8 support. But as mentioned we only needed to bump to Java 11 for the `macos-14` runner to be able to test `omega-edit` Scala code. However, the Scala Server that is used by the omega-edit packages, which in turn is used here in daffodil-vscode, is still built against Java 8. So it still works, as shown in this PR as all the OSes passed all tests even when using Java 8.

stricklandrbls (approved):
+1 Works as expected

scholarsmate (approved):
+1

PR #991, opened 2024-03-15
More helpful error messages

### LINKED ISSUE #931: Launch error messages are not helpful
For example, if a required sub-schema isn't loaded, like from the classpath, you get a popup like:
Screen Shot 2024-01-10 at 9 32 29 AM

That is, the message: `debugger daffodil.error.request`.

I understand we don't want to directly return stack traces, but we could produce better messages. In this particular case we have the actual error message from Daffodil that is useful:
```
org.apache.daffodil.debugger.dap.Compiler$CompilationFailed: Schema Definition Error: Error loading schema due to org.xml.sax.SAXParseException; DaffodilXMLLoader: Unable to resolve schemaLocation='com/owlcyberdefense/dfdl/xsd/basicByteBinary.dfdl.xsd'.
Schema context: file:/Users/arosien/nteligen/PCAP/src/main/resources/com/tresys/pcap/xsd/pcap.dfdl.xsd Location in file:/Users/arosien/nteligen/PCAP/src/main/resources/com/tresys/pcap/xsd/pcap.dfdl.xsd
```

arosien:
Rather than the user having to scroll through the mess of the backend debugger log in the Terminal, we can send the exception message along with the event type back to the extension, which then logs it in a distinct "Daffodil" Output tab.

The user then clicks the "show logs" to switch to that tab.

The current errors are:
* `ErrorEvent.UnhandledRequest`: if the backend receives a type of request it doesn't understand
* `ErrorEvent.LaunchArgsParseError`: if the backend can't parse the launch args when starting a DAP session
* `ErrorEvent.RequestError`: the backend had a problem executing a request; this includes runtime exceptions from Daffodil like schema compilation
* `ErrorEvent.SourceError`: used during an experiment for having backend-provided source files, not used currently
* `ErrorEvent.UnexpectedError`: used when a watch expression can't be evaluated; this error seems to be redundant and could perhaps be merged with `RequestError`

The PR doesn't display the particular the error type, only the underlying message. The user can look at the regular logs for the details.

Fixes #931.

### DIFF
--- a/debugger/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.debugger.dap/DAPodil.scala
@@ -49,9 +49,9 @@ trait DAPSession[Req, Res, Ev] {
 
   def sendResponse(response: Res): IO[Unit]
   def sendEvent(event: Ev): IO[Unit]
-  def abort(event: DebugEvent): IO[Unit]
-  def abort(event: DebugEvent, logMessage: String): IO[Unit]
-  def abort(event: DebugEvent, logMessage: String, t: Throwable): IO[Unit]
+  def abort(event: Ev): IO[Unit]
+  def abort(event: Ev, logMessage: String): IO[Unit]
+  def abort(event: Ev, logMessage: String, t: Throwable): IO[Unit]
 }
 
 object DAPSession {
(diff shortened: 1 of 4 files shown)


lrbarber:
What types of error messages are affected by this change?

hdalsania:
@arosien How do we test this PR?

hdalsania (approved):
+1 It is working as expected. I tried to debug a schema with syntax error and I get Show logs and that took me to Output tab with distinct "Daffodil" tab from the dropdown.  More detailed logs is available on regular terminal window.

shanedell (approved):
+1 👍

shanedell:
@arosien For future notice, normally 2 green checkmarks are desired before merging in a PR.

arosien:
> @arosien For future notice, normally 2 green check marks are desired before merging in a PR.

there were 2: @hdalsania and @shanedell

stevedlawrence:
I don't think @hdalsania is a Daffodil committer or PMC. Although we definitely appreciate reviews from non-committers/PMC, and doing so is a good way to show commitment to the project and earn the merit to become one, their votes do not count towards the required 2 +1's ASF requires merge. The green checkmark is a good indication of this since non-committers only get a grey checkmark when they approve a PR.

PR #993, opened 2024-03-18
Omega Edit status messages must be in tolerance

### LINKED ISSUE #989: CPU Load Average reports -1.0.0 in WIndows
In Windows, hover over the green dot and see that the CPUI Load Average is -1.00, which is incorrect.  It may not be getting reported properly by the server or may not be available from the JVM in Windows.

If the CPU Load Average is not easily found for Windows, instead of showing -1.00, we should not report it at all.

scholarsmate:
To test this, run the Data Editor from a Windows machine, hover over the green dot down where the Omega Edit status information is, and check that the CPU Load Average is now not being reported because it is not in tolerance.

Fixes CPU Load Average reports -1.0.0 in Windows.

closes #989

### DIFF
--- a/src/svelte/src/components/ServerMetrics/ServerMetrics.svelte
+++ b/src/svelte/src/components/ServerMetrics/ServerMetrics.svelte
@@ -107,32 +107,49 @@ limitations under the License.
         on:mouseenter={() => showHeartbeatInfo(true)}
         on:mouseleave={() => showHeartbeatInfo(false)}
       >
-        {#if heartbeat.latency < 20}
+        {#if 0 < heartbeat.latency && heartbeat.latency < 20}
           <circle cx="50%" cy="50%" r="4pt" fill="green" />
-        {:else if heartbeat.latency < 40}
+        {:else if 0 < heartbeat.latency && heartbeat.latency < 40}
           <circle cx="50%" cy="50%" r="4pt" fill="yellow" />
-        {:else if heartbeat.latency > 60}
+        {:else if 0 < heartbeat.latency && heartbeat.latency > 60}
           <circle cx="50%" cy="50%" r="4pt" fill="red" />
         {:else}
           <circle cx="50%" cy="50%" r="4pt" fill="grey" />
         {/if}
       </svg>
       <div class="heartbeat-info">
-        <b>Latency:</b>
-        {heartbeat.latency}ms,
-        <b>CPU Load Avg:</b>
-        {(heartbeat.serverCpuLoadAverage
-          ? heartbeat.serverCpuLoadAverage
-          : 0
-        ).toFixed(2)},
-        <b>Memory Usage:</b>
-        {heartbeat.serverUsedMemory},
-        <b>Session Count:</b>
-        {heartbeat.sessionCount},
-        <b>Uptime:</b>
-        {prettyPrintUptime(heartbeat.serverUptime)}
-        <b>JVM Version:</b>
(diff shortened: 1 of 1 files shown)


stricklandrbls:
>  check that the CPU Load Average is now no being reported because it is not in tolerance.

Is that supposed to say "_not_ being reported"? Running this PR as described but I don't see the `CPULoadAverage` from this section:
https://github.com/apache/daffodil-vscode/blob/4b09ed25b80d6c9f7cd1a811c0aa68871e1b6407/src/svelte/src/components/ServerMetrics/ServerMetrics.svelte#L133-L136

**Screenshot Reference**
image

rthomas320 (commented):
This change seems to have broken the "none" "action" option for tdml configuration.
image
If I change the tdml "action" to "generate" the selected documents and the dataEditor open. when I hover over the green dot, the CPU load average is not present.
image

scholarsmate:
@stricklandrbls, that is the expected behavior.  If a measurement is not in tolerance, it doesn't get reported at all.  Not sure why the Windows JVM can't properly report the average CPU load, but if that ever does get repaired, then it will be properly reported in the metrics at that time without any changes to this code.

@rthomas320, perhaps that's happening because this branch needs to be rebased to incorporate any recent updates and fixes.  If that's the case, I'll do a rebase and force push it.

scholarsmate:
@rthomas320, the branch is currently up-to-date with main.  Can you please try again?  Be sure to run `git clean -fdx` before the build to make sure it's clean.

stricklandrbls:
@scholarsmate 
I retested the PR with the main rebase and it still works as expected.

--- 
> ... If a measurement is not in tolerance, it doesn't get reported at all ...

I'm wondering if we should have a placeholder for this in the output instead of just omitting it entirely.

src/svelte/src/components/ServerMetrics/ServerMetrics.svelte
```diff
@@ -107,32 +107,49 @@ limitations under the License.
...
+        {/if}
+        {#if heartbeat.serverUptime > 0}
+          <b>Uptime:</b>
+          {prettyPrintUptime(heartbeat.serverUptime)}
+        {/if}
+        {#if heartbeat.sessionCount > 0}
+          <b>Session Count:</b>
+          {heartbeat.sessionCount},
+        {/if}
+        {#if heartbeat.serverCpuLoadAverage > 0}
```
hdalsania, line 133: Because of this change, windows does not display CPU Load Avg. at all because it is not >0? 

Windows:
image


Linux:
image

scholarsmate: Correct, you and @stricklandrbls said the same.  I wanted to avoid placeholders simply due to (lack of) real estate.

stricklandrbls: > I wanted to avoid placeholders simply due to (lack of) real estate.

The using `humanReadableByteLength(byteLength: number): string ` from `./src/svelte/src/utilities/display.ts` on the "Memory Usage" value might help with the real estate. I'm be fine approving this PR with omitting the `CPULoadAverage` value for Windows and possibly creating another ticket for restructuring the way server information is output.

stricklandrbls (approved):
+1

Though we should look into restructuring the server info output method.

rthomas320 (approved):
Works with TDML action set to none and CPU load average is not showing in the hover over status green dot.

PR #994, opened 2024-03-22
Fix multiple bugs with infoset files

### LINKED ISSUE #972: Infoset popup to open the infoset fails to open the infoset.
When using the daffodil debugger to build a complete infoset, a popup will display at the completion of the infoset asking if you want to open the infoset.  If you select the option to open the infoset, it doesn't open. (This may be a windows only issue that has to do with backslash vs forwardslash.
image

shanedell:
Fix multiple bugs with infoset files.

- Resolve issue where the infoset popup to open infoset didn't actually open the infoset on windows.
- Resolve issue where the infoset.save command was trying to save to file://<filename> instead of file://${workspaceFolder}/<filename>.
  - The above would also cause the file not to be able to be opened or deleted.

Closes #972

### DIFF
--- a/src/infoset.ts
+++ b/src/infoset.ts
@@ -25,6 +25,7 @@ import {
   ensureFile,
   tmpFile,
 } from './utils'
+import * as path from 'path'
 
 // Function to display an infomation message that the infoset file has been created
 // If the user wishes to open the file then they may click the 'Open' button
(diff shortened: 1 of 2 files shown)


shanedell:
Not sure why, but the `yarn.lock` seemed like it also needed updated.

scholarsmate (approved):
+1

rthomas320 (approved):
Infoset opens in an editor from the infoset popup query in windows 10 and Rocky 9.

PR #1014, opened 2024-04-26
Refactor to fix event delivery guarantees.

### LINKED ISSUE #995: DataEditor character highlight doesn't advance when the continue (F5) option is selected.
The character highlight in the dataEditor view moves appropriately when stepover (F10) is selected, but it doesn't move to the appropriate character when continue (F5) is selected.  If stepover is selected after continue, the character highlight will jump to the appropriate character.  But if continue is selected again the character highlight won't move again until stepover is selected.  This behavior occurs in both the Windows 10 environment and the Rocky 9 environment.

arosien:
Events were emitted, or sometimes not emitted at all, before state changes were committed.

Some refactorings to both directly fix this and also improve the code:
- New `Parse.Deliver` class encapsulates translation of `Parse.Event`s into DAP state and events, delivered via the `Debugee` interface. It also ensures state is delivered before events so that consumers can react to events and read the correct state values. `Parse.Deliver` also fixes the bug of not delivering certain events, like the data position upon the "continue" action, by delivering events at every kind of stopping action, which includes "step", "continue", and a hit breakpoint.
- Remove `Debugee.state` and have the `Debugee` implementation emit the DAP events directly via `Debugee.events`.
- Instead of using `None`-terminated `Queue`, which enqueues with `Some` and terminates with `None`, use the `Channel` data type which only uses the actual message type.
- To lift an `IO` into a `Resource`, replace `Resource.eval(io)` with `io.toResource`.

Fixes #995.

### DIFF
--- a/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
@@ -34,7 +34,6 @@ import java.nio.file._
 import org.apache.commons.io.FileUtils
 import org.apache.daffodil.debugger.dap.{BuildInfo => DAPBuildInfo}
 import org.apache.daffodil.runtime1.debugger.Debugger
-import org.apache.daffodil.runtime1.api.InfosetElement
 import org.apache.daffodil.runtime1.infoset.{DIDocument, DIElement, InfosetWalker}
 import org.apache.daffodil.runtime1.processors._
 import org.apache.daffodil.runtime1.processors.dfa.DFADelimiter
@@ -50,6 +49,7 @@ import org.typelevel.log4cats.slf4j.Slf4jLogger
 import scala.collection.JavaConverters._
 import scala.util.Try
 
+/** A Daffodil parse of a schema against data. */
 trait Parse {
 
   /** Run the parse, returning the bytes of the final infoset. */
@@ -70,6 +70,7 @@ object Parse {
           .mkString("\n")
       )
 
+  /** Create a parse using the given schema, data, etc. */
   def apply(
       schema: Path,
       data: InputStream,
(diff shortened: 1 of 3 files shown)


debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
```diff
@@ -1117,11 +1152,13 @@ object Parse {
...
             Delimiter(if (i < pstate.mpstate.delimitersLocalIndexStack.top) "remote" else "local", delimiter)
           },
-          pstate.diagnostics
+          pstate.diagnostics,
+          infoset
         )
     }
     case class EndElement(state: StateForDebugger) extends Event
     case object Fini extends Event
+    case class Control(state: DAPodil.Debugee.State) extends Event
```
arosien, line 1161: Rather than having two queues, one for data (state) changes, and another for events (control), we have a single event type so we can emit & commit state changes before event notifications are delivered.

hdalsania (approved):
+ 1 (Grey)

I tested this on Windows and RHEL9 platform. Data editor updates along with breakpoints in debugger.

rthomas320 (commented):
The change has fixed the problem with the continue button not advancing the highlight marker in the data editor. Everything else appears to work as before the change.  I will give my +1 when the CI passes Windows node 18 tests.

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
```diff
@@ -1364,33 +1399,40 @@ object Parse {
...
             .ifM(
-              events.offer(Some(new Event.StartElement(pstate, isStepping = true))) *> setInfoset,
-              events.offer(Some(new Event.StartElement(pstate, isStepping = false)))
+              events.offer(Some(new Event.StartElement(pstate, infoset))),
+              events.offer(Some(new Event.StartElement(pstate, None))) // we only emit infoset when stepping
             ) *>
           logger.debug("post-control await") *>
           logger.debug("pre-checkBreakpoints") *>
-          // TODO: there's a race between the readers of `events` and `state`, so somebody could react to a hit breakpoint before the event data was committed
+          // TODO: infoset should have been sent prior to a breakpoint hit
```
scholarsmate, line 1435: Is this still a TODO?

shanedell:
@arosien Pull down the latest main and rebase this PR off of it. This will fix your CI issues, once those are passed and I double check everything is working for me I will +1 this PR.

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
arosien: done

rthomas320 (approved):
Works as expected. +1

shanedell (approved):
+1 👍

stricklandrbls (approved):
+1

PR #1016, opened 2024-04-30
fix save segment dialog (#1012)

### LINKED ISSUE #1012: Incorrect Data Editor Profiler's Save Segment As Notification
When saving a segment from the Profiler, VSCode sends a notification that the segment has been saved but the filepath is incorrect. 

For example, if the target filepath is `/home/user/some_dir/targetfile.txt` and a segment is saved from the profiler w/ the filepath `/tmp/saved-seg1.txt` then the VSCode info message states: "Saved: /home/user/some_dir/targetfile.txt".

scholarsmate:
Fixes #1012.

### DIFF
--- a/src/dataEditor/dataEditorClient.ts
+++ b/src/dataEditor/dataEditorClient.ts
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-import fs from 'fs'
 import {
   ALL_EVENTS,
   beginSessionTransaction,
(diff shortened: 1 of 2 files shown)


src/dataEditor/dataEditorClient.ts
```diff
@@ -60,34 +59,35 @@ import {
...
 import path from 'path'
+import * as vscode from 'vscode'
 import XDGAppPaths from 'xdg-app-paths'
-import assert from 'assert'
-import { SvelteWebviewInitializer } from './svelteWebviewInitializer'
+import { extractDaffodilEvent } from '../daffodilDebugger/daffodil'
+import {
+  EditByteModes,
+  VIEWPORT_CAPACITY_MAX,
+} from '../svelte/src/stores/configuration'
```
stricklandrbls, line 73: I'm confused as to why these are marked as additions. Especially the removal of `import fs from 'fs'` on line 18 and then it's added again on line 63. Same goes for `net`, `os`, and `vscode`...

I feel as if there's been a rebasing issue here.

scholarsmate: Not a rebasing issue, just VSCode sorting the import statements on a save.

stricklandrbls (approved):
+1

shanedell (approved):
+1

rthomas320 (approved):
+1 Works for me on Windows 10.

PR #1022, opened 2024-05-02
Fix broken TDML append, copy and execute functionality

### LINKED ISSUE #1013: Fix TDML Execute and Append functionality
TDML Execute and Append functionality is broken since inception of the TDML GUI editor (PR# 936). 

Fix the default settings for test case name, description and path as it is showing as "Undefined" in Launch.json file.

hdalsania:
closes #1013 

Fixed TDML functionality. Now copyTDML, Append and Execute TDML works as expected.  

Following functionality still need re-visit and further enhancement issues needs to be created. 

- in copy, it is currently being hardcoded to getTmpTDMLFile location. If a user is specifying a custom value in the launch config, we won't be able to pick that up right now (append is probably working the same way) (#1027 )
- in copy, there's no error message when it doesn't work (#1027)
- in append, if you are trying to add a duplicate, there is no error message letting the user know that's why nothing happened (#1028 )
- in execute, if the user takes longer than the timeout to enter name/description, the server will disconnect. #596 should fix this issue. (#1029 )
- in generate, TDML name, description and infoset path is defaults to string "undefined" (#1024 )
- Execute functionality should only require TDML test case name only to execute, currently it requires name and description. #596 should fix this issue. (#1030 )
- append is listed as a valid action in the launch wizard - Append is not a server action and should not be available as an option in launch config. (#1031 )
- TDML Copy function creates the wrong path in some Windows 10 environments (#1034)
- Fix TDML Copy, Execute and Append functionality in MacOS Platform (#1035)

Steps to test TDML copy, execute, append

Configure launch.json to generate a TDML file.
image

Run the debug extension, choose a dfdl schema and data file.  After the schema opens make sure the language mode is DFDL. 
image

Press the continue button to produce the infoset.
image

When the infoset generates, a temporary TDML schema will generate.
image

Close any open window except the dfdl schema window.  Click “Copy TDML File” in the dropdown at the top tight of the editor window. 
image

Enter a name for the TDML file, click “Save TDML File in the lower right corner of the popup. 
image

Close the dfdl schema in the editor window.  Click on the explore tab in the top left corner to display the file view. Click on the save TDML file. 
image

After the TDML file opens click the menu in the top right corner and select “Execute TDML” 
image

Quickly select a test case a press enter.  Then quickly accept the description and press enter.
image

The previously selected dfdl schema will open and the previously selected data file will be used and empty infoset will also open. 
image

To append to the existing TDML file, open the test file and click the button in the upper right corner to open in a text editor. 
image

Change the test case name and save the file.
image

Select append from the TDML dropdown menu at the upper right. 
image

The original default test case will be appended to saved TDML file with the new test case.
image

### DIFF
--- a/debugger/src/main/scala/org.apache.daffodil.tdml/TDML.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.tdml/TDML.scala
@@ -123,6 +123,7 @@ object TDML {
       .relativize(new File(path.toString()).toURI())
       .getPath()
       .toString()
+      .trim()
   }
 
   // Generate a new TDML file.
@@ -253,7 +254,13 @@ object TDML {
   // Returns a tuple containing the following (Path to DFDL Schema, Path to Data File)
   // All paths returned could be either relative or absolute - it depends on what exists in the TDML file
   def execute(tdmlName: String, tdmlDescription: String, tdmlPath: String): Option[(Path, Path)] = {
-    val basePath = Paths.get(tdmlPath).toAbsolutePath().getParent()
+
+    // Correct basepath for Windows OS
+    var basePath = ""
+    if (System.getProperty("os.name").toLowerCase.startsWith("win") == true)
+      (
+        basePath = Paths.get(tdmlPath).toAbsolutePath().getParent().toString()
+      )
 
     val testCaseList = JAXBContext
       .newInstance(classOf[TestSuite])
@@ -278,6 +285,7 @@ object TDML {
                 .asInstanceOf[JAXBElement[DocumentPartType]]
                 .getValue()
                 .getValue()
+                .trim()
             )
             .normalize()
           (schemaPath, dataPath)
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -29,7 +29,7 @@ import * as dfdlExt from '../language/semantics/dfdlExt'
 import {
   appendTestCase,
   getTmpTDMLFilePath,
-  readTDMLFileContents,
+  copyTestCase,
 } from '../tdmlEditor/utilities/tdmlXmlUtils'
 import xmlFormat from 'xml-formatter'
 
@@ -85,6 +85,46 @@ async function getFile(fileRequested, label, title) {
   return file
 }
 
+/** Method to show dialog to save TDML file
+ * Details:
+ *   Required so that the vscode api commands:
+ *     - extension.dfdl-debug.getValidatedTDMLCopyPath
+ *   can be sent a file instead of always opening up a prompt.
+ */
+async function showTDMLSaveDialog(fileRequested, label, title) {
+  let file = ''
+
+  file = await vscode.window
+    .showSaveDialog({
+      saveLabel: label,
+      title: title,
+      filters: {
+        TDML: ['tdml'],
+      },
+    })
+    .then((fileUri) => {
+      if (fileUri) {
+        let path = fileUri.fsPath
+
+        if (
+          process.platform === 'win32' &&
+          path.length > 2 &&
+          path.charCodeAt(0) > 97 &&
+          path.charCodeAt(0) <= 122 &&
+          path.charAt(1) === ':'
+        ) {
+          path = path.charAt(0).toUpperCase() + path.slice(1)
+        }
+
+        return path
+      }
+
+      return ''
+    })
+
+  return file
+}
+
 // Function for setting up the commands for Run and Debug file
 function createDebugRunFileConfigs(
   resource: vscode.Uri,
@@ -115,6 +155,12 @@ function createDebugRunFileConfigs(
 
       if (tdmlAction) {
         tdmlConfig.action = tdmlAction
+
+        if (tdmlAction === 'execute') {
+          tdmlConfig.name = '${command:AskForTDMLName}'
+          tdmlConfig.description = '${command:AskForTDMLDescription}'
+          tdmlConfig.path = targetResource.fsPath
+        }
       }
 
       vscode.debug.startDebugging(
@@ -124,11 +170,12 @@ function createDebugRunFileConfigs(
           request: 'launch',
           type: 'dfdl',
           schema: {
-            path: targetResource.fsPath,
+            path: tdmlConfig.action === 'execute' ? '' : targetResource.fsPath,
             rootName: null,
             rootNamespace: null,
           },
-          data: '${command:AskForDataName}',
+          data:
+            tdmlConfig.action === 'execute' ? '' : '${command:AskForDataName}',
           debugServer: false,
           infosetFormat: 'xml',
           infosetOutput: {
@@ -189,10 +236,7 @@ export function activateDaffodilDebug(
           targetResource = vscode.window.activeTextEditor.document.uri
         }
         if (targetResource) {
-          appendTestCase(
-            await readTDMLFileContents(getTmpTDMLFilePath()),
-            await readTDMLFileContents(targetResource.fsPath)
-          )
+          appendTestCase(getTmpTDMLFilePath(), targetResource.fsPath)
             .then((appendedBuffer) => {
               fs.writeFileSync(targetResource.fsPath, xmlFormat(appendedBuffer))
             })
@@ -209,19 +253,43 @@ export function activateDaffodilDebug(
         createDebugRunFileConfigs(resource, 'run', 'execute')
       }
     ),
-    vscode.commands.registerCommand('extension.dfdl-debug.copyTDML', (_) => {
-      // Ask for destination path
-      // Copy file in /tmp to destination path
-      // TDMLConfig.path should not be used here because that only matters when sending to the server
-      // We could make it so that if someone wants to specify the path and set the action to 'copy', but
-      //   that doesn't make a whole lot of sense.
-      let targetResource = vscode.commands.executeCommand(
-        'extension.dfdl-debug.getTDMLPath'
-      )
+    vscode.commands.registerCommand(
+      'extension.dfdl-debug.copyTDML',
+      async (_) => {
+        // Ask for destination path
+        // Copy file in /tmp to destination path
+        // TDMLConfig.path should not be used here because that only matters when sending to the server
+        // We could make it so that if someone wants to specify the path and set the action to 'copy', but
+        //   that doesn't make a whole lot of sense.
+        let targetResource = await vscode.commands.executeCommand(
+          'extension.dfdl-debug.getValidatedTDMLCopyPath'
+        )
 
-      // Is there a better way of error checking this?
-      fs.copyFileSync(getTmpTDMLFilePath(), targetResource as unknown as string)
-    }),
+        // Is there a better way of error checking this?
+        if (targetResource) {
+          copyTestCase(
+            getTmpTDMLFilePath(),
+            targetResource as unknown as string
+          )
+            .then((copiedBuffer) => {
+              fs.writeFileSync(
+                targetResource as unknown as string,
+                xmlFormat(copiedBuffer)
+              )
+            })
+            .catch((reason) => {
+              // Not sure if we need to do something different/more here
+              console.log(reason)
+            })
+        }
+
+        // fs.copyFile(
+        //   getTmpTDMLFilePath(),
+        //   targetResource as unknown as string,
+        //   (_) => {}
+        // )
+      }
+    ),
     vscode.commands.registerCommand(
       'extension.dfdl-debug.toggleFormatting',
       (_) => {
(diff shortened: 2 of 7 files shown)


debugger/src/main/scala/org.apache.daffodil.tdml/TDML.scala
```diff
@@ -123,6 +123,8 @@ object TDML {
       .relativize(new File(path.toString()).toURI())
       .getPath()
       .toString()
+      .replaceFirst("^\\s++", "")
+      .replaceFirst("\\s++$", "")
```
scholarsmate, line 127: Just use `.trim()` here instead.  It will remove leading and trailing whitespace characters from strings.

debugger/src/main/scala/org.apache.daffodil.tdml/TDML.scala
```diff
@@ -278,6 +280,8 @@ object TDML {
                 .asInstanceOf[JAXBElement[DocumentPartType]]
                 .getValue()
                 .getValue()
+                .replaceFirst("^\\s++", "")
+                .replaceFirst("\\s++$", "")
```
scholarsmate, line 284: Use `.trim()`.

debugger/src/main/scala/org.apache.daffodil.tdml/TDML.scala
hdalsania: .trim() only removes a whitespace and it does not remove multiple whitespaces before and after. We also used .strip() function but it is only available post Java11 and make it noncompatible with Java8

debugger/src/main/scala/org.apache.daffodil.tdml/TDML.scala
hdalsania: .trim() only removes a whitespace and it does not remove multiple whitespaces before and after. We also used .strip() function but it is only available post Java11 and make it noncompatible with Java8

debugger/src/main/scala/org.apache.daffodil.tdml/TDML.scala
arosien:
```
scala> "'" + "  foo\t  ".trim + "'"
val res1: String = 'foo'
```
`trim` does the right thing

arosien: although... it is very odd to have to do this to a string that is derived from a `File` and `Path`... what's up?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -85,6 +85,52 @@ async function getFile(fileRequested, label, title) {
   return file
 }
 
+/** Method to select file path to save TDML file
+ * Details:
+ *   Required so that the vscode api commands:
+ *     - extension.dfdl-debug.getValidatedTDMLCopyPath
+ *   can be sent a file instead of always opening up a prompt.
+ */
+async function saveFile(fileRequested, label, title) {
```
arosien, line 94: Concern: this doesn't actually save a file, it returns a path where the data may be saved. Perhaps rename to something like `getTDMLSavePath`?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -85,6 +85,52 @@ async function getFile(fileRequested, label, title) {
...
+  let file = ''
+
+  if (fileRequested && fs.existsSync(fileRequested)) {
+    file = fileRequested
+  } else if (fileRequested && !fs.existsSync(fileRequested)) {
+    file = ''
+  } else {
+    file = await vscode.window
```
arosien, line 102: Actually, my suggestion would be to move the `if/else` logic to where this method is called, and have this method be named `showTDMLSaveDialog` and return the file path.

src/daffodilDebugger/debugger.ts
```diff
@@ -57,22 +58,42 @@ async function getTDMLConfig(
...
     if (config?.tdmlConfig?.description === undefined)
+      config.tdmlConfig.description = await vscode.commands.executeCommand(
+        'extension.dfdl-debug.getTDMLDescription'
+      )
+  }
+
+  if (config?.tdmlConfig?.action === 'generate') {
+    if (
+      config?.tdmlConfig?.name === undefined ||
+      config?.tdmlConfig?.name === 'undefined'
```
arosien, line 82: Concern: this looks really wonky. Why is it needed?

src/tdmlEditor/utilities/tdmlXmlUtils.ts
```diff
@@ -260,9 +260,36 @@ export async function getTestCaseDisplayData(
...
+  destinationFilePath: string
+) {
+  return copyTestCase(xmlFilePath, destinationFilePath, true)
+}
+
+/*
+ * Copy and optionally append test case in one XML file into another file's test suite
+ *
+ * xmlFilePath: String containing the XML file path of a TDML file with
+ *   exactly one test case
```
arosien, line 273: Concern: it doesn't really expect a TDML file with only one test case, does it? Then append would only work once and never again on the same file.

src/tests/suite/tdmlXmlUtils.test.ts
```diff
@@ -171,17 +167,16 @@ suite('TDML Utils Test Suite', () => {
...
-            // This is checked because we threw
-            strictEqual(reason, 'Duplicate Test Case Name Found')
-          })
-      }
+    return appendTestCase(
+      join(data_directory, 'test.tdml'),
+      join(data_directory, 'test.tdml')
     )
+      .then((_) => {
+        // This is not checked because we threw
```
arosien, line 175: Question: put in a test failure to check it really isn't invoked?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -209,19 +259,43 @@ export function activateDaffodilDebug(
...
+            targetResource as unknown as string
+          )
+            .then((copiedBuffer) => {
+              fs.writeFileSync(
+                targetResource as unknown as string,
+                xmlFormat(copiedBuffer)
+              )
+            })
+            .catch((reason) => {
+              // Not sure if we need to do something different/more here
```
arosien, line 287: Questin: What happens if `catch` is removed? Does the whole extension stop, or is the error sent to the console anyway?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -209,19 +259,43 @@ export function activateDaffodilDebug(
...
-      // Ask for destination path
-      // Copy file in /tmp to destination path
-      // TDMLConfig.path should not be used here because that only matters when sending to the server
-      // We could make it so that if someone wants to specify the path and set the action to 'copy', but
-      //   that doesn't make a whole lot of sense.
-      let targetResource = vscode.commands.executeCommand(
-        'extension.dfdl-debug.getTDMLPath'
-      )
+    vscode.commands.registerCommand(
+      'extension.dfdl-debug.copyTDML',
```
arosien, line 263: Question: Why is this command available to the user? It assumes some TDML file exists at some path... I just don't get it.

src/adapter/activateDaffodilDebug.ts
```diff
@@ -272,6 +346,64 @@ export function activateDaffodilDebug(
     )
   )
 
+  context.subscriptions.push(
```
arosien, line 349: Concern: I don't know why these functions need to be registered as user-available commands. Can't they just be functions used by the extension itself?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -209,19 +259,43 @@ export function activateDaffodilDebug(
...
       }
     ),
-    vscode.commands.registerCommand('extension.dfdl-debug.copyTDML', (_) => {
-      // Ask for destination path
-      // Copy file in /tmp to destination path
-      // TDMLConfig.path should not be used here because that only matters when sending to the server
-      // We could make it so that if someone wants to specify the path and set the action to 'copy', but
-      //   that doesn't make a whole lot of sense.
-      let targetResource = vscode.commands.executeCommand(
-        'extension.dfdl-debug.getTDMLPath'
```
arosien, line 219: Ok, I think I see one of the bugs: there was no `getTDMLPath` command registered. But I still don't know why this needs to be a user command, it could just be a function.

I think you followed the--to my eyes weird--pattern, where a command was declared but not defined. It should just be a function and not declared as a command.

src/adapter/activateDaffodilDebug.ts
hdalsania: We can use getSavePath function name to make it generic function to get a path to save data, in this case TDML data.

src/adapter/activateDaffodilDebug.ts
hdalsania: We can do that

src/daffodilDebugger/debugger.ts
hdalsania: This is definitely not needed, this is a harmless workaround to handle current launch config wizard implementation. We are planning to create another technical debt or bug issue #1024 is created to handle this.

src/tdmlEditor/utilities/tdmlXmlUtils.ts
hdalsania: This functionality need to be revisited in future release.

src/adapter/activateDaffodilDebug.ts
hdalsania: This functionality is implemented based on the design discussion listed [here](https://github.com/ctc-oss/daffodil-vscode/discussions/126) and described below,

- We should not need to specify a location for the TDML file. We should always put it in /tmp (or OS equivalent) under a static name
- We should provide 'Cleanup TDML File' and 'Copy TDML File' actions on the command palette.
- Copy TDML File should copy the existing file to the user-defined location under $workspace.
- There should be an Append/Copy option that takes the test case from the file under /tmp and appends it to the user-defined location under $workspace.

src/adapter/activateDaffodilDebug.ts
hdalsania: Changed the function name to getSavePath

src/adapter/activateDaffodilDebug.ts
arosien: Thanks for the link and summary, it makes sense now.

lrbarber (approved):
Tested successfully on Windows with node v16.14.0

rthomas320 (approved):
Works as described in Rocky 9, Ubuntu, and RHEL 9 environments.  It doesn't work in my Windows 10 environment.  In that environment the "temp" tdml file is created with 2 extra directory backup symbols (../).  IT seems to work in most Windows 10 enviroments though.  A difference with my PC setup is my user name is different than the name used in my home path.  More research needs to be done to determine exactly what is causing the issue.

stricklandrbls:
### M1 Macbook Testing
Testing this on my M1 macbook resulted in no generatedTDML file in any temporary directory. 

### Pop!OS (Ubuntu 22.04) Testing
The generated file is created successfully and I can run the Copy, Append, and Execute operations. Though when executing there is an uninformative info dialog message that pops up that only contains the string of the infoset file name:

image

scholarsmate (approved):
+1, tested on Windows.

scholarsmate:
Didn't want to merge this yet until @stricklandrbls approves.

stricklandrbls:
### M1 Macbook Pro Test Failure
@hdalsania @scholarsmate @shanedell - I just tested this again on my M1 Macbook Pro and I still cannot Copy / Append because the generated tmp file is not created. 

The _Copy TDML File_ button exists but will not save the `.tdml` file anywhere once the data debugger is finished and I attempt to save.

<h3>VSCode Debug Output</h3>

```
simulatorDirPath: undefined
deviceFilePath: undefined
simulatorDirPath: undefined
deviceFilePath: undefined
detected java homes: [{"path":"/Library/Java/JavaVirtualMachines/jdk-11.0.16.1.jdk/Contents/Home","version":"11.0.16","security":0,"isJDK":true,"is64Bit":true,"executables":{"java":"/Library/Java/JavaVirtualMachines/jdk-11.0.16.1.jdk/Contents/Home/bin/java","javac":"/Library/Java/JavaVirtualMachines/jdk-11.0.16.1.jdk/Contents/Home/bin/javac","javap":"/Library/Java/JavaVirtualMachines/jdk-11.0.16.1.jdk/Contents/Home/bin/javap"}},{"path":"/Library/Java/JavaVirtualMachines/microsoft-11.jdk/Contents/Home","version":"11.0.16","security":0,"isJDK":true,"is64Bit":true,"executables":{"java":"/Library/Java/JavaVirtualMachines/microsoft-11.jdk/Contents/Home/bin/java","javac":"/Library/Java/JavaVirtualMachines/microsoft-11.jdk/Contents/Home/bin/javac","javap":"/Library/Java/JavaVirtualMachines/microsoft-11.jdk/Contents/Home/bin/javap"}},{"path":"/Library/Java/JavaVirtualMachines/temurin-8.jdk/Contents/Home","version":"1.8.0","security":302,"isJDK":true,"is64Bit":true,"executables":{"java":"/Library/Java/JavaVirtualMachines/temurin-8.jdk/Contents/Home/bin/java","javac":"/Library/Java/JavaVirtualMachines/temurin-8.jdk/Contents/Home/bin/javac","javap":"/Library/Java/JavaVirtualMachines/temurin-8.jdk/Contents/Home/bin/javap"}}]
choosing java home at /Library/Java/JavaVirtualMachines/jdk-11.0.16.1.jdk/Contents/Home, version 11.0.16, is at least JDK 17: false
No test suite found in source XML buffer
```


The output line: `No test suite found in source XML buffer` gets outputted every time I attempt to save the generate `.tdml` file via _Copy TDML File_.

shanedell:
@hdalsania @stricklandrbls I also get a `No test suite found in source XML buffer` when I try to `Copy TDML File` or when I try to `Append TDML`.

stricklandrbls:
@hdalsania @scholarsmate @shanedell I did a little digging and running the extension in debug mode w/ breakpoints. It looks like this is coming from this scope of code:

https://github.com/apache/daffodil-vscode/blob/5bd426e1dbdf25b40193ba0a2eb17dbc9d861ba1/src/tdmlEditor/utilities/tdmlXmlUtils.ts#L277-L295

I think it has something to do with this function being `async` but not returning a `Promise` because this block of code throws with _all_ of the scoped variables being `undefined` or uninitialized.

hdalsania:
We created a bug issue #1035 to address MacOS platform and changed the scope of this PR to be limited to Windows/Linux platform.

PR #1060, opened 2024-06-19
Configure Local Extension Module Resolution

### LINKED ISSUE #1059: Reduce Project Inter-Module Resolution Names
Module path resolution has become too verbose now that the extension contains _inter-module_ imports. The `compilerOptions.paths` tsconfig configuration attribute can be utilized to ensure cleaner and more understandable module importing.

stricklandrbls:
- Configured the build/extension environment to resolve local modules to allow for inter-module importing w/o the need to resolve to a relative path.

This configuration also removes a lot of the path resolution discrepancies between the `tsc` and `webpack` path resolutions. These were typically output to the console upon packages/debugging.

It also allows for easier module resolution for future functionality that includes components that have a large "relative path distance" or standalone project, such as the `./src/svelte` directory.

Closes #1059


shanedell:
@stricklandrbls I have taken a look at this and have also not been able to figure out why the items are not being resolved. I am assuming it is something with how we run tests since we do a `node ./out/tests/runTest.js`. Maybe we need to figure out a way to use mocha or something similar to how the data editor tests are ran? Those are working fine with the aliases right? If so maybe the way to move forward is trying to run via a mocha command inside of using mocha's library like we currently. However, I don't know how well that plays with the need for the VSCode API and what not but might be good to look into.

shanedell:
@stricklandrbls Below is a commit message I added:

```text
@shanedell
Fix most tests:

- Below lines that were added to src/tests/suite/index.ts allowed the tests to resolve module aliases:
  - require('ts-node/register')
  - require('tsconfig-paths/register')
- Renamed src/launchWizard/launchWizard.js to src/launchWizard/script.js
  - This resolved an issue with the alias where it would sometimes import the launchWizard.js instead of launchWizard.ts.
- The tests still not passing seem to be failing because of pico
```

So I was able to get most tests passing except some for the data editor that seem due to pico. Wasn't sure if you would be able to figure that quicker than I can.

shanedell:
@stricklandrbls Below is a commit message I added:

```text
@shanedell
Fix most tests:

- Below lines that were added to src/tests/suite/index.ts allowed the tests to resolve module aliases:
  - require('ts-node/register')
  - require('tsconfig-paths/register')
- Renamed src/launchWizard/launchWizard.js to src/launchWizard/script.js
  - This resolved an issue with the alias where it would sometimes import the launchWizard.js instead of launchWizard.ts.
- The tests still not passing seem to be failing because of pico
```

So I was able to get most tests passing except some for the data editor that seem due to pico. Wasn't sure if you would be able to figure that quicker than I can. But according to CI I might have just been hitting the issue locally as some stages were passing all tests.

scholarsmate (approved):
+1, LGTM.

rthomas320 (approved):
+1 I didn't see any obvious problems in the code and the extension works as expected.

PR #1066, opened 2024-07-12
fixes a bug where quoted text is replaced by a space when a slash is typed

### LINKED ISSUE #1064: Unwanted replacement value inserted when typing a slash within quotes
If a schema file is opened in the editor without going through the launch.json configuration that opens a schema file and data file in the debugger a slash in some quoted fields causes the quoted value to be replaced with a space.
The code is treating "schemaLocation" in '<xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />' as an attribute.  The slash was triggering an attribute value replacement function that replaces the quoted value with a space and presents the user with the predefine list of values for that attribute. Since there was no predeifine value list the default alphabetic list of keywords in the file was provide by VS Code by default.
The same problem occurs with "source" in ' <xs:appinfo source="http://www.ogf.org/dfdl/">' and in the namespace defines in the schema tag.

rthomas320:
fixes a bug where quoted text is replaced by a space when a slash is typed

To test:
Either create and install the extension or run the extension in the vscode debugger.
In the vscode window running the extension, open the folder containing the dfdl schema you are going to test with.
Open the dfdl schema in that folder, but don't use debug to open the schema with the launch.json configuration.
Place the cursor on one of the quoted values of the schema namespace defines.
(if the schema you select contains <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd" />
or ' <xs:appinfo source="http://www.ogf.org/dfdl/"> you could also try placing the cursor in those quoted values)
Press the '/' slash key.
The quoted text should not be replaced.
Try the typing a space in the same quoted value. (A space is the only defined trigger for this function, I'm not sure why slash is triggering it).
The quoted text should not be replaced .
You can also try this test using the launch.json config that asks for a schema file and a data file, but I could not reproduce the failure when opening the file that way.

closes #1064

### DIFF
--- a/src/language/providers/attributeHover.ts
+++ b/src/language/providers/attributeHover.ts
@@ -17,6 +17,7 @@
 
 import * as vscode from 'vscode'
 import { attributeHoverValues } from './intellisense/attributeHoverItems'
+import { attributeCompletion } from './intellisense/attributeItems'
 
 export function getAttributeHoverProvider() {
   return vscode.languages.registerHoverProvider('dfdl', {
@@ -28,11 +29,17 @@ export function getAttributeHoverProvider() {
       const range = document.getWordRangeAtPosition(position)
       const word = document.getText(range)
 
+      let itemNames: string[] = []
+      attributeCompletion('', '', 'dfdl:', '', '').items.forEach((r) =>
+        itemNames.push(r.item)
+      )
       if (word.length > 0) {
-        return new vscode.Hover({
-          language: 'dfdl',
-          value: attributeHoverValues(word),
-        })
+        if (itemNames.includes(word)) {
+          return new vscode.Hover({
+            language: 'dfdl',
+            value: attributeHoverValues(word),
+          })
+        }
       }
     },
   })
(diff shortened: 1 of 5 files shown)


stricklandrbls:
### Current Test Environment 
**Operating System**: Debian 12.5 _Bookworm_
**Extension Runner**: Running the `Extension` debug launch config
**DFDL Schema**: PNG _( png.dfdl.xsd_ )

When typing a _space_ within the quoted text of line 26 ( `<xs:appinfo source="http:// www.ogf.org/dfdl/">` ), the quoted text **does** disappear. However, typing a _/_ character **does not** remove the quoted text.

**Screenshot**
dfdl-space-delete
In this screenshot I initially type a _/_ then a _space_ which causes the text deletion.

rthomas320:
@stricklandrbls , Add fixes for the issues that you mentioned above.

stricklandrbls:
@rthomas320 
I grabbed my M1 series MacBook and was testing this PR and found that these issues we were seeing **do not occur**.
Jul-16-2024 09-08-11

I'm going to test it a bit on my Debian machine to make sure it's working there too.

stricklandrbls (approved):
+1

scholarsmate (approved):

Issue #1065, opened 2024-07-12
VS Code triggers code completion when typing a slash '/', but the trigger is defined as a space '  ' only


rthomas320:
When opening a dfdl schema file in a vscode editor with the daffodil-vscode extension installed, but not using the launch.json to open the file, code completion is triggered by a '/' slash in a function that only has space as a trigger.

PR #1077, opened 2024-08-11
Fix code completion trigger for slash in dfdl schema files

Suvendu-UI:
The code completion trigger in dfdl schema files was incorrectly set to a space only, but it was also being triggered by a slash '/'. This pull request fixes the issue by updating the trigger to only happen for a space '  '.

### DIFF
--- a/src/language/providers/closeElementSlash.ts
+++ b/src/language/providers/closeElementSlash.ts
@@ -62,7 +62,7 @@ export function getCloseElementSlashProvider() {
           return undefined
         }
 
-        if (triggerText.endsWith('/')) {
+        if (triggerText.endsWith(' ')) {
           let range = new vscode.Range(backpos, position)
 
           await vscode.window.activeTextEditor?.edit((editBuilder) => {
@@ -84,7 +84,7 @@ export function getCloseElementSlashProvider() {
       },
     },
     '/'
-    // triggered whenever a '/' is typed
+    // triggered whenever a ' ' is typed
   )
 }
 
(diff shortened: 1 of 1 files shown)


rthomas320:
I created issue #1065, "VS Code triggers code completion when typing a slash '/', but the trigger is defined as a space ' ' only", which I think your PR is addressing, but I didn't fully understand the problem when I put the PR in.  We still want the '/' slash to trigger the self closing end tag.  The actual problem was described in an email from one of the Daffodil project managers.  You can get more information about joining the dev email list here: https://cwiki.apache.org/confluence/display/DAFFODIL/For+Contributors
The email described this problem:

> (blocking) There is a real problem with "/" being typed. I know it is trying to detect closing the element, but it has to be context sensitive since "/" is such a common character to appear inside strings.
> To reproduce, at the top level of the schema, try typing <xs:include schemaLocation="/ I.e., a string that begins with a "/" character. The GUI refuses to let you do this. It removes the quotation marks and anything you've typed inside them.
> So I tried typing <xs:include schemaLocation="" /> then going back and putting a "/" inside the quotes. Again the editor refuses, removes the quotes, remotes the "/" just typed.
> I tried using single quotes, no improvement.
> I tried typing some characters e.g., 'abc/ It still immediately reacts to the "/" character being typed, and deletes the string and quotation mark.
> I re-opened the file from scratch. There is a line like this:
> <xs:include schemaLocation="org/apache/daffodil/xsd/DFDLGeneralFormat.dfdl.xsd"
> />
> I position the cursor before the "org" and type "/", and it removes the entire path.
> Interestingly, if I hit ctrl-Z to undo, it puts the text back, including the new leading "/" character. But this is not a workaround.
Sorry for any confusion.

PR #1078, opened 2024-08-13
Slash space fix

rthomas320:
Fixes a problem where a space or slash typed in a quoted attribute value for attributes without pre-defined values removes the quoted text. Also fixes a problem where a slash typed almost anywhere in the schema will insert '/>' the self closing end tag.

Description from Mike Beckerle's email:

> (blocking) There is a real problem with "/" being typed. I know it is trying to detect closing the element, but it has to be context sensitive since "/" is such a common character to appear inside strings.
To reproduce, at the top level of the schema, try typing <xs:include schemaLocation="/ I.e., a string that begins with a "/" character. The GUI refuses to let you do this. It removes the quotation marks and anything you've typed inside them.
So I tried typing <xs:include schemaLocation="" /> then going back and putting a "/" inside the quotes. Again the editor refuses, removes the quotes, remotes the "/" just typed.
I tried using single quotes, no improvement.
I tried typing some characters e.g., 'abc/ It still immediately reacts to the "/" character being typed, and deletes the string and quotation mark.

### DIFF
--- a/src/language/providers/attributeCompletion.ts
+++ b/src/language/providers/attributeCompletion.ts
@@ -79,10 +79,10 @@ export function getAttributeCompletionProvider() {
           .text.substring(0, position.character)
         const charBeforeTrigger = triggerText.charAt(position.character - 1)
         const charAfterTrigger = triggerText.charAt(position.character)
-        let nearestOpenItem = nearestOpen(document, position)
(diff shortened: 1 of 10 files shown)


scholarsmate:
@rthomas320, please rebase this to pass CI.

PR #1107, opened 2024-09-19
Fixes intellisense problems discovered in RC1

### LINKED ISSUE #1065: VS Code triggers code completion when typing a slash '/', but the trigger is defined as a space '  ' only

rthomas320:
without default values when typng a space or slash in the quoted text
Fixes a problem where typing a slash almost anywhere inserts '/>' a self closing tag
Prevents hover popups for non-attribute items
Fixes intellisense suggesting incorrect attributes for elements that spread across multiple lines

closes #1065

### DIFF
--- a/src/language/providers/attributeCompletion.ts
+++ b/src/language/providers/attributeCompletion.ts
@@ -326,6 +326,26 @@ function checkNearestOpenItem(
         spacingChar,
         afterChar
       )
+    case 'include':
+      return getCompletionItems(
+        ['schemaLocation'],
+        '',
+        '',
+        nsPrefix,
+        '',
+        spacingChar,
+        afterChar
+      )
+    case 'import':
+      return getCompletionItems(
+        ['schemaLocation', 'namespace'],
+        '',
+        '',
+        nsPrefix,
+        '',
+        spacingChar,
+        afterChar
+      )
     case 'assert':
       return getCompletionItems(
         ['testKind', 'test', 'testPattern', 'message', 'failureType'],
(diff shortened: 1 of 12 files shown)


rthomas320:
Testing the intellisense changes:
These test are performed using the schema https://github.com/OpenDFDL/examples/blob/master/helloWorld/src/main/resources/helloWorld.dfdl.xsd
and the data file https://github.com/OpenDFDL/examples/blob/master/helloWorld/src/main/resources/helloWorld.dat
Type a slash (/) or space at the position indicated by the red arrow.  The action of typing the slash or space should not cause the text within the quotes to be truncated.
image
Type a slash at the position indicated by a red arrow. Typing the slash should not result in "/>" appearing at the cursor position.
image
Hover over the word element indicated by the red arrow no hoer hint should be provided.
image
Hover over the word maxOccurs indicated by the red arrow. Hover test with a definition should display.
image
Step through the schema in the debugger. Pressing the stepover button at line 24 should not result in a dsiplay of intellisense suggestions.
image

shanedell:
@rthomas320 I having a problem where if I try adding a `/` it is being automatically deleted. The only time it seems to not be deleted is when I am adding it some strings already present. But in some of those cases the whole string is being deleted. I will try to upload a recording of it

shanedell:
sampleWorkspace-1727193611074

rthomas320:
The problems Shane noted above should be fixed. The keyword "import" was missing from the element list and the keyword "namespace" was missing from the attribute list.  A problem with intellisense not supplying a self closing endtag when the line immediately following the trigger line had a self closing tag was resolved.

shanedell (approved):
+1 👍 Retested the latest update and was able to get expected results

PR #1118, opened 2024-10-16
Main

### LINKED ISSUE #1028: No Error message o user notification for Duplicate append test case to TDML file
TDML Append Functionality - if you are trying to add a duplicate, there is no error message letting the user know that's why nothing happened

lrbarber:
Output TDML error messages to user.  
Closes #1028

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -28,13 +28,23 @@ jobs:
       matrix:
         os: [ macos-12, ubuntu-20.04, windows-2019, macos-latest, ubuntu-latest, windows-latest ]
         node: [ '16', '18.20.1' ] # version 18.20.2 (current latest of 18) is broken on windows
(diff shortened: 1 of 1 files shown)


rthomas320 (commented):
The coding changes look fine.  The source compiles and I was able to run and create an infoset and a tdml file using copy tdml.  I'm not sure how to create the conditions that would trigger the actual code. I tried deleting the test file name from the tdml file and running execute tdml. But that didn't trigger the error "No test case found".  That maybe because of a problem with the incorrect path being generated in the tdml file.  I received this error "ERROR o.a.d.d.d.DAPSession - couldn't launch, request #2 launch" which may supercede the change you made.  The path in the tdml file is always incorrect for me and it may be because of the way my user directory was setup on my machine as rthomas instead of registhomas. Also the single commit and type script formatting failed in CI.

PR #1151, opened 2025-01-20
Remove excludes for windows and vscode versions.

### LINKED ISSUE #1101: Investigate why tests fail with newer versions of vscode
After VSCode version 1.91.1 it seems the test fail. This should also involve an update to running the tests to not strictly define using `1.67.2` from the `package.json`. Instead, if the environment variable is not set the default version should be `stable`. The CI file should also be fixed to removed pinned version `1.91.1` and revert back to `stable`.

shanedell:
Remove excludes for windows and vscode versions.

- Newer versions of VSCode seem to work okay now on regular CI so adding it back into nightly
- Add sbt setup step to CI files to ensure all runners will have sbt installed.

Closes #1101

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -155,9 +155,9 @@ jobs:
           distribution: ${{ matrix.java_distribution }}
           java-version: ${{ matrix.java_version }}
       
-      - name: Install sbt - macos-13 only
-        run: brew install sbt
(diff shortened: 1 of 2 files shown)


scholarsmate (approved):
+1

mbeckerle (approved):
+1
Sure. Good to get this back working.

PR #1153, opened 2025-01-23
Update launch wizard to use dropdowns for log levels

### LINKED ISSUE #1152: Create dropdown list for Config Wizard Log Level setting
Modify the Launch Configuration Wizard to have a drop-down list of the valid options for log level setting.
Currently the user must guess what the valid option settings might be.
Should be able to just search the codebase for the valid values, then going into the typescript code for the configuration wizard and copy the code that does drop down lists into the log level setting and install the correct values.

shanedell:
Update launch wizard to use dropdowns for log levels

Closes #1152

### DIFF
--- a/src/launchWizard/launchWizard.ts
+++ b/src/launchWizard/launchWizard.ts
@@ -40,6 +40,14 @@ export async function activate(ctx: vscode.ExtensionContext) {
   )
 }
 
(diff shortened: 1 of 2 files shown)


scholarsmate (approved):
+1

PR #1154, opened 2025-01-30
Output TDML errors via vscode pop-up

lrbarber:
This fixes "No Error message or user notification for Duplicate append test case to TDML file #1028" by calling vscode.window.showErrorMessage(...) which creates a small pop-up window in the lower right corner of the VSCode window.


scholarsmate (approved):
+1

Issue #1139, opened 2024-12-05
vsix file is not reproducible


stevedlawrence:
The .vsix file generated by `vsce package` contains timestamps that prevent it from being easily reproducible. It looks like everything inside the zip is reproducible, so the vsix file itself is likely the only issue. I believe all the daffodil and daffodil-sbt releases are reproducible, so this is the last remaining piece that when fixed will allow us to switch to tag/CI based [automated release building/signing](https://infra.apache.org/release-signing.html#automated-release-signing), which should make the release process significantly easier.

I've done some digging and it looks like the issue is vsce includes the mtimes of the files in the zip/vsix file.

I've tried adding a new script in build/yarn-scripts.ts that is run just before we package the vsix file to change all the timestamps to SOURCE_DATE_EPOCH if defined. This works for most files included in the vsix file, but there are a couple that vsce generates during the package process that I don't think we can modify like that. So this likely is not a complete solution.

Another option is to create a PR to vscode-vsce to update it to use SOURCE_DATE_EPOCH when creating the vsix. I've skimmed vscode-vsce and think it could probably done with just a few lines, but I'm not all that familiar with node.js or how to test those kinds of changes. Also, even if we did make the changes, it requires getting a PR merged and updating the dependency to it. The first is probably doable, but daffodil-vscode depends on a pretty old version of vscode-vsce--is there anything blocking us from updating to a newer version if one were released with this kind of fix?

An alternative approach might be to just add a post processing script that unzips the contexts of the vsix file to a temp directory, modifies all the timestamps to SOURCE_DATE_EPOCH, and then rezips it. That's a bit of a pain, but doesn't have any dependency concerns.

stevedlawrence:
For reference, this patch to vscode-vsce seems to fix it so it can create reproducible zips by sorting the files and using SOURCE_DATE_EPOCH if defined:

```diff
diff --git a/src/package.ts b/src/package.ts
index d27b71c..9416b45 100644
--- a/src/package.ts
+++ b/src/package.ts
@@ -1800,13 +1800,21 @@ function writeVsix(files: IFile[], packagePath: string): Promise<void> {
 			() =>
 				new Promise((c, e) => {
 					const zip = new yazl.ZipFile();
-					files.forEach(f =>
+
+					const sde = process.env.SOURCE_DATE_EPOCH;
+					const epoch = sde ? parseInt(sde) : undefined;
+					const mtime = epoch ? new Date(epoch * 1000) : undefined;
+
[...]

PR #1173, opened 2025-02-10
Enable reproducible JAXB generated TDML code

### LINKED ISSUE #1142: JAXB generated TDML is not reproducible
Steps to reproduce:

```base
sbt clean resources xjc
cp -R debugger/target/scala-2.12/src_managed_cxf/ /tmp
sbt clean resources xjc
diff -rq debugger/target/scala-2.12/src_managed_cxf/ /tmp/src_managed_cxf
```

Repeat lines 3 and 4 until it shows a diff.

So far I've been able to pretty easily reproduce this on Java 8 (`OpenJDK Runtime Environment (Temurin)(build 1.8.0_432-b06)`) and less likely but still reproducible on Java 11 (`OpenJDK Runtime Environment Temurin-11.0.25+9 (build 11.0.25+9)`) and so far not at all on Java 21. Not sure if this is a Java version issue, or some versions just make it more/less likely for some reason.

An example of a diff usually start with something like this:

```diff
diff -Naur debugger/target/scala-2.12/src_managed_cxf/org/apache/daffodil/tdml/DFDLBaseType.java /tmp/src_managed_cxf/org/apache/daffodil/tdml/DFDLBaseType.java
--- debugger/target/scala-2.12/src_managed_cxf/org/apache/daffodil/tdml/DFDLBaseType.java	2024-12-06 10:09:30.318167823 -0500
+++ /tmp/src_managed_cxf/org/apache/daffodil/tdml/DFDLBaseType.java	2024-12-06 10:09:16.896018288 -0500
@@ -24,10 +24,10 @@
  *     &lt;extension base="{http://www.ogf.org/dfdl/dfdl-1.0/}DFDLType"&gt;
  *       &lt;sequence&gt;
  *       &lt;/sequence&gt;
- *       &lt;attGroup ref="{http://www.ogf.org/dfdl/dfdl-1.0/}MarkupAG"/&gt;
- *       &lt;attGroup ref="{http://www.ogf.org/dfdl/dfdl-1.0/}CommonAG"/&gt;
  *       &lt;attGroup ref="{http://www.ogf.org/dfdl/dfdl-1.0/}AlignmentAG"/&gt;
  *       &lt;attGroup ref="{http://www.ogf.org/dfdl/dfdl-1.0/}BaseAG"/&gt;
+ *       &lt;attGroup ref="{http://www.ogf.org/dfdl/dfdl-1.0/}MarkupAG"/&gt;
+ *       &lt;attGroup ref="{http://www.ogf.org/dfdl/dfdl-1.0/}CommonAG"/&gt;
  *       &lt;anyAttribute processContents='lax' namespace='urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:int urn:ogf:dfdl:2013:imp:daffodil.apache.org:2018:ext http://www.w3.org/XML/1998/namespace http://www.ibm.com/xmlns/dfdl/testData'/&gt;
  *     &lt;/extension&gt;
  *   &lt;/complexContent&gt;
@@ -52,6 +52,36 @@
...
```
It's always related to attribute groups.

It looks like however xjc outputs attribute groups, it outputs their generated comments, members, and methods in a random order, instead of the order they are defined in the schema. This doesn't technically affect anything since the the order of attribute groups and their members/functions don't matter, but it does mean the compiled .class files have members/methods in a different order, which leads to non-reproducibility.

stevedlawrence:
The resource generator in build.sbt currently extracts all schemas from the daffodil-lib jar and uses xjc to generate JAXB code that allows reading and writing TDML files. Note that this includes the large and complex XML schema for DFDL because that schema is referenced and used by the TDML schema to support embedded DFDL schemas and configurations.

However, because the XML Schema for DFDL is so large and complex, it requires special xjc bindings, and also generates java that is not deterministic, making reproducible builds difficult.

Fortunately, VS Code doesn't really care about embedded schemas in TDML files--it really only uses JAXB for modifying TDML test cases. So it does not need the many JAXB generated classes that cause the above issues related to embedded schemas.

To avoid generating these classes, after the TDML schema is extracted, an XSLT is run to replace the embedded DFDL schema and configuration complex types with "any" elements. This prevents xjc from generating special types for embedded schemas and configurations, instead generating code that just treats them as generic XML Objects. This is sufficient for VS Code uses.

Additionally, this allows us to only extract tdml.xsd from daffodil-lib since that is the only schema we need now. We can also remove the bindings.xjb file needed only for the no longer used XML Schema for DFDL.

Also changes how we get the TDML namespace for the DocumentPart element, something with this changed how namespaces work. It looks like JAXB no longers maintain namespace declarations.

This also moves the resource generator into the xjc/sources task. The two tasks actually ran in parallel and so could lead to a situation where the resource generator hadn't completed when xjc/sources was accessed and led to inconsistent results. The XSD files are really resources anyways, so this is probably the more accurate way to handle this. Additinoally, caching is added so we aren't constant extrating generating JAXB files--instead we only generate if daffodil-lib changes, which should be rare.

Closes #1142


stevedlawrence:
It's hard to know for sure if this fixes the non-deterministic JAXB code issue, but I wasn't able to reproduce it after many attempts with these changes. It was pretty easy to reproduce with the old code. Also, tdml.xsd doesn't have any elements with multiple attributeGroups, which I think is what caused the problem, so I think this change avoids that edge case.

mbeckerle (approved):
+1
Clever technique. 
I will keep in mind your build.sbt approach here. This idea that you can use an XSLT preprocessor in a build.sbt could be handy in many places.

scholarsmate (approved):
+1, I see no harm in merging this pull request.  In order to achieve full reproducibility, we'll want to explore upgrading VCSE, which will mean changing the minimum supported VSCode.

stevedlawrence:
I'll leave it up to the core daffodil vscode team to determine what minimum version of VS Code to support--I don't know enough about VS Code to know what common versions still need support.

If you decide not to merge #1166, I think we still have workarounds to resolve #1139, which is to post process the .vsix file to fix the timestamps (for example using debian's [strip-nondeterminism](https://salsa.debian.org/reproducible-builds/strip-nondeterminism) tool).

shanedell:
I can look at the VSCE PR #1166 and make the necessary updates to make sure all the CI will work properly. But I think upgrading our minimum version of NodeJS should be fine we will just need to make sure to update the documentation

Issue #1127, opened 2024-11-12
Respect global settings in settings.json, then the local ones in launch.json


scholarsmate:
Port 9000 is the default port for the Ωedit server that powers the Data Editor.  The problem is that I have another service that is running on port 9000, so I need to change the port that the Ωedit server runs on.  I want to use port 9001 for that.  I think the most intuitive place to persist that variable is in a global settings.json that is configurable at the extension level.

<img width="910" alt="Screenshot 2024-11-12 at 12 12 35 PM" src="https://github.com/user-attachments/assets/271c1144-631e-49ac-8e05-cb906b18f2df">
<img width="1335" alt="Screenshot 2024-11-12 at 12 13 29 PM" src="https://github.com/user-attachments/assets/a0011a26-28f5-4339-98a9-f33b426dd075">

Clicking on `Edit in settings.json` allows me to change the port, but that setting seems to be *ignored*.  It would be good to have that work as it will allow users to configure a different port and use the Data Editor without requiring a .vscode/launch.json in their repository.

The workaround for now is to set the port in launch.json like so:

```json
           "dataEditor": {
                "port": 9001,
                "logging": {
                    "file": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
                    "level": "info"
                }
            },
```

Having it respect the global settings, then allow them to be overridden in a local launch.json would be better.

scholarsmate:
Will be assigned to @egbicker.

JeremyYao:
### How to Replicate Issue 

1. Open Visual Studio Code up in your SampleWorkspace folder
2. Delete or rename any existing launch.json files in .vscode (ex: launch.old.json) so you can keep them
3. Navigate to your VSCode users folder (ex: C:\Users\jeremy.yao\AppData\Roaming\Code\User\settings.json)
4. Insert the following key-value into your settings.json 
```
  "dataEditor": {
  

    "port": 9991,
    "logging": {
      "file": "${workspaceFolder}/dataEditor-${omegaEditPort}.log",
      "level": "info"
    }
  }
```
5. In a separate powershell window, type in `python -m http.server 9000`
6. Open up the command pallete with `Ctrl` + `Shift` + `P`, type and select `Daffodil Debug: Data Editor`
7. You should see something like the following (Powershell w/ Python on the right and VScode on the right)

![Image](https://github.com/user-attachments/assets/b70da895-bc82-4cd9-8b02-49bedfe039de)

[...]

PR #1183, opened 2025-02-25
Recognize settings.json dataEditor port

JeremyYao:
https://github.com/apache/daffodil-vscode/issues/1127


src/dataEditor/config/Extract.ts
```diff
@@ -41,12 +41,24 @@ function extractDataEditorConfigFromQuery(
...
+  const workspaceConfig = workspace.getConfiguration()
+  const defaultConfig: Config = {
+    ...Config.Default, // Default configs
+    port: workspaceConfig.dataEditor?.port || Config.Default.port, // Get dataEdtior port from settings.json if it exists
+
+    // To PR Reviewers, do we need the following below or no? Otherwise, the bottom 2 attributes get ignored from settings.json VSCode users folder.
+    // logLevel:
+    //   workspaceConfig.dataEditor?.logging?.level || Config.Default.logLevel,
+    // logFile:
+    //   workspaceConfig.dataEditor?.logging?.file || Config.Default.logFile,
```
scholarsmate, line 53: Uncomment these as we want those to be settable in settings.json and respected.

JeremyYao: Uncommented

JeremyYao: Seems to cause the builds to fail. The biggest issue is that it's not able to resolve paths and substitute in VSCode-specific variables into the string. 

image

JeremyYao: Issue only applies to the logFile. We can keep the logLevel as is.

JeremyYao: Addressed this

rthomas320 (approved):
The changes are working as described.
image

shanedell (approved):
+1

PR #1184, opened 2025-02-26
Add GitHub action to create release candidates

stevedlawrence:
Adds a new "Release Candidate" workflow that uses the new GitHub action and is triggered when a release candidate tag is created. This simplifies the release candidate process to just pushing a single signed tag. Workflow dispatch is allowed for testing, but it will not publish anything.

DAFFODIL-2971

Note: marking this as draft while the proposal is under discussion and needs ASF Secruity Team approval


VERSION
```diff
@@ -0,0 +1 @@
+1.4.1-SNAPSHOT
```
stevedlawrence, line 1: Unfortunately, I don't think there is a way for package.json to read this file. So the for daffodil-vscode, version bumps require modifying two files, package.json and VERSION

stevedlawrence: The action was intended to be run immediately after the checkout action and before yarn/npm/sbt/etc commands are run that could generate that file. I *guess* it might work to run something to generate the VERSION file in between checkout and asf-release-candidate actions, but it feels safer to run from a clean slate.

I'm open to other suggestions for the action to figure out the version, but I would like to avoid parsing build configs. That worked in the past because we only have three projects, but I really want this action to be general and simple enough that it could be used by other ASF projects so that maybe someday the maintenance could be shared.

stevedlawrence:
ASF Infra has add all the necessary secrets. I've squashed the fixup commits, removed this from draft status, and marked it ready for review.

Please let me know if duplicating version information in package.json and the VERSION file is okay. I'm open to other suggestions.

shanedell (approved):
+1 👍 I am okay with having to update package.json and VERSION to bump version.

We can possibly add a yarn script that makes this easy. So, say a command like `update-version` that takes in a version as an argument, it will check that the version is in proper format and then write the new version to both files. So for example this how the command could be ran: `yarn update-version v1.4.1`

stevedlawrence:
@shanedell would you prefer the script and/or CI be added as part of this PR or done in separately?

shanedell:
@stevedlawrence Either would be perfectly fine with me

shanedell:
@stevedlawrence Will this be something we can merge soon? We are planning to release 1.4.1-rc1 and put it out for a vote soon, I wanted to wait to do it so we can use this new release workflow. I would just merge it but wasn't sure if you wanted that script/CI to bump the version in both files to this PR or not.

stevedlawrence:
I don't think I have time right now to right the script. I'll merge this now and the script can be added later, and just manually bump edit the two files in the meantime and be careful to ensure they stay in sync.

PR #1189, opened 2025-02-28
Use ubuntu-22.04 for CI workflows.

### LINKED ISSUE #1176: Update workflows to use newer version of ubuntu
Currently the workflows use `ubuntu-20.04` which is being retired.

- For the normal CI workflow we need to upgrade to `ubuntu-22.04`. 
- For the nightly workflow we need to simply delete `ubuntu-22.04` from the matrix.

shanedell:
Use ubuntu-22.04 for CI workflows.

Closes #1176

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,7 +25,7 @@ jobs:
   single-commit:
     name: Single Commit Pull Request
     if: github.event_name == 'pull_request'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Check Single Commit
         uses: actions/github-script@v7.0.1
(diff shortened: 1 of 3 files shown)


stevedlawrence (approved):
+1, note that the upcoming GitHub Action automate release workflow uses ubuntu-22.04, so this is consistent with that

Related, another thing worth considering is the daffodil-infrastructure build-release container uses ubuntu-22.04 so that it uses  a similar build environment to GitHub actions CI/releases, helping with build reproducibility. So if we ever want to bump to ubuntu-24.04, we'll want to do it for all projects and the build-release container at the same time, so that everything uses the same environment.

stricklandrbls (approved):
+1

PR #1204, opened 2025-03-19
Add left over data message to debugger and extension

### LINKED ISSUE #974: No warning given when data left over after parse.
When running daffodil from the command line, if a schema parse completes, but does not parse the entire data file, a warning message about there being left over data is issued.
When this occurs inside Visual Studio, there is no such warning message.
There may be other similar messages that are not being presented by the debugger.

shanedell:
Add left over data message to debugger and extension

- After a parse completes, the debugger will check if there is data left over.
- When data is left over the debugger will log an error message to the console.
- The debugger also sends data, bitPos1b, bytePos1b, leftOverBits and message to the extension.
- Currently the extension parses the sent over data and shows an error message with the message the debugger sent.

Closes #974

### DIFF
--- a/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
@@ -74,12 +74,14 @@ object Parse {
   def apply(
       schema: Path,
       data: InputStream,
+      dataFilePath: Path,
       debugger: Debugger,
       infosetFormat: String,
       rootName: Option[String],
       rootNamespace: Option[String],
       variables: Map[String, String],
-      tunables: Map[String, String]
+      tunables: Map[String, String],
+      dapEvents: Channel[IO, Events.DebugEvent]
   ): IO[Parse] =
     for {
       dp <- Compiler()
@@ -104,16 +106,31 @@ object Parse {
               case "json" => new JsonInfosetOutputter(os, true)
             }
 
+            val dataSize = data.available()
+            val dataInputStream = new InputSourceDataInputStream(data)
+            val parseResult = dp.parse(dataInputStream, infosetOutputter)
+            val loc = parseResult.location()
+
             val parse =
               IO.interruptibleMany(
-                dp.parse(
-                  new InputSourceDataInputStream(data),
-                  infosetOutputter
-                )
+                parseResult
                 // WARNING: parse doesn't close the OutputStream, so closed below
               ).ensureOr(res => new Parse.Exception(res.getDiagnostics.toList))(res => !res.isError)
-                .guarantee(IO(os.close) *> done.set(true))
                 .void
+                .flatMap { _ =>
+                  val leftOverBits = (dataSize - (loc.bytePos1b - 1)) * 8
+
+                  if (leftOverBits > 0) {
+                    val message = DataLeftOverUtils.getMessage(dataFilePath, loc.bitPos1b, loc.bytePos1b, leftOverBits)
+
+                    Logger[IO].error(message) *> dapEvents.send(
+                      DataLeftOverEvent(loc.bitPos1b, loc.bytePos1b, leftOverBits, message)
+                    ) *> IO.unit
+                  } else {
+                    IO.unit
+                  }
+                }
+                .guarantee(IO(os.close) *> done.set(true))
 
             stopper &> parse
           }
@@ -754,12 +771,14 @@ object Parse {
             Parse(
               args.schemaPath,
               in,
+              args.dataPath,
               debugger,
               args.infosetFormat,
               args.rootName,
               args.rootNamespace,
               args.variables,
-              args.tunables
+              args.tunables,
+              dapEvents
             )
           )
           .toResource
@@ -1261,6 +1280,8 @@ object Parse {
       )
   }
   case class DataEvent(bytePos1b: Long) extends Events.DebugEvent("daffodil.data")
+  case class DataLeftOverEvent(bitPos1b: Long, bytePos1b: Long, leftOverBits: Long, message: String)
+      extends Events.DebugEvent("daffodil.dataLeftOver")
   case class InfosetEvent(content: String, mimeType: String) extends Events.DebugEvent("daffodil.infoset")
   object InfosetEvent {
     def apply(format: String, node: DIElement): InfosetEvent =
--- a/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Utils.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Utils.scala
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.debugger.dap
+
+import java.io.RandomAccessFile
+import java.nio.ByteBuffer
+import java.nio.file.Path
+import org.apache.daffodil.io.DataDumper
+
+/** Data Left Over Utils */
+object DataLeftOverUtils {
+
+  /** Get message based on dataFilePath, bitPos1bm bytePos1b and leftOverBits */
+  def getMessage(dataFilePath: Path, bitPos1b: Long, bytePos1b: Long, leftOverBits: Long): String = {
+    val leftOverBitsText =
+      s"Left over data. Consumed ${bitPos1b - 1} bit(s), with at least ${leftOverBits} bit(s) remaining.\n"
+
+    val Dump = new DataDumper
+
+    val numBytes = math.min(8, (leftOverBits / 8)).toInt
+
+    // read to byte array starting where the parse ended
+    val f = new RandomAccessFile(dataFilePath.toFile, "r")
+    val bytes = new Array[Byte](numBytes)
+    f.seek(bytePos1b - 1)
+    f.readFully(bytes)
+    f.close
+
+    // convert those bytes to hex/binary
+    val dumpString =
+      if (bytes.length > 0)
+        Dump
+          .dump(
+            Dump.TextOnly(Some("utf-8")),
+            0,
+            bytes.length * 8,
+            ByteBuffer.wrap(bytes),
+            includeHeadingLine = false
+          )
+          .mkString("\n")
+      else ""
+
+    val dataHex =
+      if (bytes.length > 0)
+        s"Left over data (Hex) starting at byte ${bytePos1b} is: (0x${bytes.map { a =>
+            f"$a%02x"
+          }.mkString}...)\n"
+      else ""
+
+    val dataText =
+      if (bytes.length > 0)
+        s"Left over data (UTF-8) starting at byte ${bytePos1b} is: (${dumpString}...)"
+      else ""
+
+    leftOverBitsText + dataHex + dataText
+  }
+}
--- a/src/daffodilDebugger/daffodil.ts
+++ b/src/daffodilDebugger/daffodil.ts
@@ -25,6 +25,14 @@ export interface DaffodilData {
   bytePos1b: number
 }
 
+export const dataLeftOverEvent = 'daffodil.dataLeftOver'
+export interface DaffodilDataLeftOver {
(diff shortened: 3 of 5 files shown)


debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
```diff
@@ -104,16 +106,31 @@ object Parse {
...
+                parseResult
                 // WARNING: parse doesn't close the OutputStream, so closed below
               ).ensureOr(res => new Parse.Exception(res.getDiagnostics.toList))(res => !res.isError)
                 .guarantee(IO(os.close) *> done.set(true))
                 .void
+                .flatMap { _ =>
+                  val bitsRead = loc.bitPos1b - 1
+                  val leftOverBits = (dataContents.length - (loc.bytePos1b - 1)) * 8
+
+                  if (bitsRead > 0 && leftOverBits > 0) {
```
stevedlawrence, line 124: Why the `bitsRead > 0` condition? It's legal for a DFDL schema to successfully parse but not consume any data, so that bitsRead will be zero. If were was data to consume it this case, it feels like we would still want the left over data message.

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Parse.scala
```diff
@@ -754,12 +771,14 @@ object Parse {
             Parse(
               args.schemaPath,
               in,
+              FileUtils.readFileToByteArray(args.dataPath.toFile),
```
stevedlawrence, line 774: This will prevent you from being able to support files larger than 2GB. If this is important, you might want to just pass in the `File` and then the functions can call things like `Files.size()`, or `Files.read()` with offset to read just the parts of the file you want to output in diagnostics. That would avoid the 2GB limitation.

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Utils.scala
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.daffodil.debugger.dap
+
+import java.nio.charset.StandardCharsets
+
+/** Data Left Over Utils */
+object DataLeftOverUtils {
```
stevedlawrence, line 23: This seems a bit text-centric, which things like skipping whitespace. If there's left over data, spaces can be very important to understanding what is left over and usually want to to be output. And if data is binary, then users just want to see raw bytes. You might want to look at what Daffodil does to output left over data for diagnostics for inspiration.

https://github.com/apache/daffodil/blob/main/daffodil-cli/src/main/scala/org/apache/daffodil/cli/Main.scala#L1339

Daffodil has a special "Dump" class we use in the CLI, it's not technically part of the public API, can you are still free to use it, it hasn't changed in a very long time to it should be pretty stable.

stevedlawrence (approved):
+1, just some minor comments

src/daffodilDebugger/daffodil.ts
```diff
@@ -83,13 +96,15 @@ export class DaffodilDebugEvent<
...
+
   constructor(
     readonly event: E,
     readonly body: B
   ) {
     this.session = debug.activeDebugSession!
   }
-  asEditorMessage(): { command: string; data: any } {
+
+  asObject(): { command: string; data: any } {
```
stricklandrbls, line 107: Looks like the `{command: string; data: any}` could have a type definition since this signature is being called in multiple places.

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Utils.scala
stevedlawrence: I could have sworn Java had a built-in way to read part of a file, but I can't find what I was thikning of. Though, looks like you could do something like
```scala
// calculate number of bytes to dump beyond the end of the parse, maybe
// something like min(10, bytesRemaining) 
val numBytes = ...

// read to byte array starting where the parse ended
val f = new RandomAccessFile(...)
val bytes = new Array[Byte](dumpSize)
f.seek(bytePos1b - 1)
f.readFully(bytes)

// convert those bytes to hex/binary
val hexLines = DataDumper.dump(HexKind, loc.bitPos1b - 1, numBytes * 8, ByteBuffer.wrap(bytes), includeHeadingLine = false)
val textLines = DataDumper.dump(TextKind, loc.bitPos1b - 1, numBytes * 8, ByteBuffer.wrap(bytes), includeHeadingLine = false)
```

That's not tested, but I was thinking something along those lines.

JeremyYao:
Performed tests as outlined in the ticket. 

image

for first test case.

For parse error test case, seems to show the Parse Error in the debug console, but pop-up on the bottom right doesn't show 

image

Test case for bad schema definition error shows this

image

shanedell:
@JeremyYao The error doesn't show up in the Debug Console, it will be Terminal output. I am wondering if its not showing in there and that's why the popup is not coming up. Currently I am facing that same issue with some of the changes I have made

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Utils.scala
stevedlawrence: Yep, looks good to me

shanedell: @stevedlawrence do you notice anything wrong with the way I am logging that is causing it not to happen?

stevedlawrence: Nothing jumps about at me, but I'm not too familiar with the details of how the vscode extensions and dap stuff works.

stevedlawrence: That is the expected output of the dump function--it converts non-visible characters to visible glyphs. For example, space is converted to  `␣`, NUL characters are conerted to `␀`. This makes it easier for users to visualize what the non-printable characters are.

Maybe the logging system can't handle those more complex characters or they need to be escaped somehow?

shanedell: That is what I am assuming since the message is correct and when using `println` it can display the message perfectly fine

hdalsania:
> @JeremyYao The error doesn't show up in the Debug Console, it will be Terminal output. I am wondering if its not showing in there and that's why the popup is not coming up. Currently I am facing that same issue with some of the changes I have made

Errors are showing up in Terminal Output but popup is only showing up for data left over error. Other two test cases from Jeremy's post do not show the popup window but error shows up in the Terminal Output window.

shanedell:
@hdalsania @JeremyYao So this PR does only the left over data message as a popup. It does not have a popup for the other two errors that @JeremyYao mentioned, those would want to be separate issues but could probably be 1 PR that fixes both.

hdalsania:
> @hdalsania @JeremyYao So this PR does only the left over data message as a popup. It does not have a popup for the other two errors that @JeremyYao mentioned, those would want to be separate issues but could probably be 1 PR that fixes both.

Make sense. We will create a new issue for this and track that separately.

hdalsania (approved):
+1

debugger/src/main/scala/org.apache.daffodil.debugger.dap/Utils.scala
shanedell: @stevedlawrence So I think I have something good locally, but it doesn't not work fully if I include the DataDumper stuff specifically. Do you know what could be causing my issue?

shanedell:
@stevedlawrence Can you re-review this? I pushed up my changes that use `DataDumper`, however the message is not currently showing in the log or the pop up window. This seems to be an issue caused by `DataDumper` because if I remove the `DataDumper` related parts and only add the dataHex then the message and popup works fine. Is there an error I could be encountering with `DataDumper` that you are aware of that might not be easy to spot? I have been working on this for a couple days and can't seem to figure out what is going on.

shanedell:
> > Can you re-review this?
> 
> Sure thing, I'll take a look.

@stevedlawrence I just tested and if I remove these lines

```scala
val Dump = new DataDumper
val dumpString =
  if (bytes.length > 0)
    Dump
      .dump(
        Dump.TextOnly(Some("utf-8")),
        0,
        bytes.length * 8,
        ByteBuffer.wrap(bytes),
        includeHeadingLine = false
      )
      .mkString("\n")
  else ""
```

then do

```scala
val dumpString = new String(bytes, StandardCharsets.UTF_8)
```

I get this output:

```bash
2025-03-25 13:21:10,199 [io-compute-7] ERROR o.a.d.d.d.Parse - Left over data. Consumed 840 bit(s), with at least 405840 bit(s) remaining.
Left over data (Hex) starting at byte 106 is: (0x736d616c6c20746f...)
Left over data (UTF-8) starting at byte 106 is: (small to...)
```

So, I imagine my issue is something related with that `Dump.dump`

stevedlawrence:
All Dump.dump does is convert non-printable characters into special unicode characters that are visible. It seems like something can't handle those special unicode characters.

For example, if you skip Dump.dump and you do this to create the dump string:

```scala
val bytes = Array[Byte](0xE2.toByte, 0x90.toByte, 0x80.toByte)
val dumpString = new String(bytes, StandardCharsets.UTF_8)
```

That creates a string with just the special `␀` character.

I tried tracking it down, but it's not clear what's breaking with these unicode symbols. It could be a lot of places. Maybe the DAP sender doesn't support it, maybe the DAP receiver doesn't, maybe it can't extract it from the DAP protocol, maybe VSCode can't display these unicode symbols. Or maybe there's en encoding mismatch between sender and receiver.

I don't have VS Code build setup so it's hard to test where things are failing, but my guess is something can't handle these more complex unicode characters.

Note that special unicode characters are nothing special, and can appear in things like infosets, so this isn't necessarily specific to left over data and debug messages.

shanedell:
@stevedlawrence So its failing I believe right at:

```scala
Logger[IO].error(message) *> dapEvents.send(
  DataLeftOverEvent(loc.bitPos1b, loc.bytePos1b, leftOverBits, message)
) *> IO.unit
```

It seems that neither are working, meaning its both not being logged or forwarded to the extension. I was able to verify this when setting a breakpoint in the extension code where it would catch the event being sent over, this breakpoint would never get hit. I tried to clean out special characters after getting the message but that didn't seem to help anything. At this point I am unsure of what I should do to try to resolve the issue other than not using `DataDumper` but not sure if that's something we want

stevedlawrence:
The core issue has nothing to do with the DataDumper except that is is more likely to output weird unicode characters. But left over data can contain any bytes and so could just as easily contain bytes the break the logger or dapEvent thing. 

Can you confirm if it's the Logger or dapEvents.send that is breaking? If you remove the logger and do this does it still fail?

```scala
dapEvents.send(DataLeftOverEvent(loc.bitPos1b, loc.bytePos1b, leftOverBits, message)
```

Or if you just do:
```scala
Logger[IO].error(message)
```
Does it correctly log the message?

rthomas320:
I pulled in the code for PR #1204 Left over data message.  I am not seeing the left over data message on Windows. 10.  Also the Omega Edit server window is no longer closing after debugging finishes.

stevedlawrence:
I've managed to get the extension installed and I can reproduce the issue. I can confirm that it does look like it's somehow related to `Dump.dump`--if I manually create strings that output ` ␀` it works fine, so Dump.dump is does seem to be doing something weird. I'll continue investigating and let you know what I find.

shanedell:
> I pulled in the code for PR #1204 Left over data message. I am not seeing the left over data message on Windows 10. Also the Omega Edit server window is no longer closing after debugging finishes.

@rthomas320 Currently @stevedlawrence and I are debugging the issue with the left over message not showing when using DataDumper so that should hopefully get resolved soon. However, the issue with `omega-edit` is not related to this PR, that might be another issue that needs to created, it might have broken with @scholarsmate's latest merged PR possibly?

shanedell:
@stevedlawrence That makes a lot of sense, so we basically need to make it so that basically the program allows for the `Dump.dump(...)` to fully complete before trying to close out? Now, from what you are saying its going to be something with `cats-effects` and not the DataDumper itself right?

shanedell:
@stevedlawrence found the fix. I had the reorder methods the parse methods, so it needed to be this:

```scala
val parse =
  IO.interruptibleMany(
    parseResult
    // WARNING: parse doesn't close the OutputStream, so closed below
  ).ensureOr(res => new Parse.Exception(res.getDiagnostics.toList))(res => !res.isError)
    .void
    .flatMap { _ =>
      val leftOverBits = (dataSize - (loc.bytePos1b - 1)) * 8
      if (leftOverBits > 0) {
        val message = DataLeftOverUtils.getMessage(dataFilePath, loc.bitPos1b, loc.bytePos1b, leftOverBits)

        logger.error(message) *> dapEvents.send(
          DataLeftOverEvent(loc.bitPos1b, loc.bytePos1b, leftOverBits, message)
        ) *> IO.unit
      } else {
        IO.unit
      }
    }
    .guarantee(IO(os.close) *> done.set(true))
```
So I needed to flip where `.flatMap` and `.guarantee` were and everything seems to work!

shanedell:
@stevedlawrence I pushed that update up, so everything should be good now if you wanted to re-review. @rthomas320 You should now be able to pull down the updated code and test it, you should get the logging message and popup for left over data.

scholarsmate (approved):
+1

PR #1205, opened 2025-03-24
Added .xml extension after .prev to be recognized as an xml file

### LINKED ISSUE #1011: Infoset Diff View not consistently showing syntax highlighting on previous
Start the infoset diff view with nothing currently in the infoset view, it will not highlight the previous side properly. The file extension is using a infoset.xml.prev extension which may be causing this issue or it could be the file needs something in it prior to opening to trigger the syntax highlighting properly.

hdalsania:
Now, when  openInfosetDiffView option is turned on within Launch.json file, it will highlight prev file with proper xml tags highlighting at startup.

Steps to test:

- Select openInfosetDiffView options to be true in Launch.json file

- Run Debugger and step through the schema few times to generate content within .prev file and verify that xml tags are properly highlighted as shown below,

image

closes #1011


shanedell (approved):
+1

scholarsmate (approved):
+1

PR #1227, opened 2025-04-09
Address incorrect relative paths when Copying or Appending TDML

### LINKED ISSUE #1034: TDML Copy function creates the wrong path in some Windows 10 environments
The generate TDML process creates a valid TDML file in my Windows 10 environment, in the users appdata/local/temp directory.
image
But the copy function adds additional back directory symbols (..\) to the location of the described files.
image
This causes the TDML execute function to fail to locate the described files.
image
One thing to note:
My user path is setup as C:\Users\rthomas\, but my user ID is regis.thomas.
The Windows 10 users that the TDML copy process works for have their user path setup like C:\Users\regis.thomas\, matching their user ID.  This maybe the cause of the problem.

JeremyYao:
Closes #1034

### DIFF
--- a/src/tdmlEditor/utilities/tdmlXmlUtils.ts
+++ b/src/tdmlEditor/utilities/tdmlXmlUtils.ts
@@ -336,9 +336,22 @@ export async function copyTestCase(
       node.name?.endsWith(testCaseElement)
     )
 
+    /**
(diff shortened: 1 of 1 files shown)


scholarsmate (approved):
+1

shanedell (approved):
+1

PR #1231, opened 2025-04-17
Test if release candidate SVN secrets are set correctly

stevedlawrence:
For only the daffodil-vscode repo, testing the release candidate action fails with the error:

  Input required and not supplied: svn_username

The same configuration works with other repots, and other secrets work correctly in this repo. To help debug this, temporarily add a new step to the release candidate action that determines if the svn secrets are set or not.

This commit will be reverted once we determine if secret is missing or empty.


stevedlawrence:
I will revert this shortly after merging and running a quick test. I think this doesn't necessitate 2 +1's. Merging now.

PR #1249, opened 2025-05-08
Address Incorrect Number of Fields for none TDML Action Upon First Op…

### LINKED ISSUE #1248: Configure Launch.json: Incorrect Number of Fields for none TDML Action Upon First Opening
### Description

Upon first opening up the Launch Config Wizard, if none is set as the TDML Action, then it shows the TDML Name and Description field. 

Image

If you switch the TDML action to something else other than none and switch back, it shows no fields

Image

Same thing happens if you switch the launch config to another one that has none as the TDML action 

Image

### How to Replicate

1. Create a new launch.json 

Image

2. Set the TDML action to none and save

Image

3. Configure the launch.json w/ the Daffodil option

Image

4. Observe 

Image

5. Note: 
If you switch the TDML action to something else other than none and switch back, it shows no fields

Image

Same thing happens if you switch the launch config to another one that has none as the TDML action 

Image

JeremyYao:
…ening

Closes #1248

### DIFF
--- a/src/launchWizard/launchWizard.ts
+++ b/src/launchWizard/launchWizard.ts
@@ -510,7 +510,7 @@ class LaunchWizard {
 
     // tdml items need 0 height and width when hidden so there is no large empty space
     let tdmlNameDesVisOrHiddenStyle =
-      tdmlAction !== null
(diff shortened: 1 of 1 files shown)


rthomas320 (approved):
Works as described. Tested on windows 10.

PR #1256, opened 2025-05-09
Remove Byte value of 256 from CSV of profiled data

### LINKED ISSUE #1255: Data Editor: Profiled Data as CSV Includes 256 as a Byte Value
> The CSV should only cover the range from 0 - 255 inclusive (the 256 result should not exist).  For the byte frequency chart, you're not seeing some of the higher value bytes due to insufficient horizontal real-estate since a vertical scrollbar is eating up that space. We need to give the Data Profiler more horizontal space to facilitate a possible vertical scrollbar. 

 _Originally posted by @scholarsmate in [#1253](https://github.com/apache/daffodil-vscode/issues/1253#issuecomment-2867332363)_

When profiling data, the CSV outputs 256 as a byte value which isn't a valid value. 

### How to replicate

Good sample files:

- [staircase_bytes.txt](https://github.com/user-attachments/files/20126269/staircase_bytes.txt)
- [staircase_bytes_reverse.txt](https://github.com/user-attachments/files/20126268/staircase_bytes_reverse.txt)
- [staircase_bytes_v.txt](https://github.com/user-attachments/files/20126266/staircase_bytes_v.txt)
- [uniform_bytes.txt](https://github.com/user-attachments/files/20126267/uniform_bytes.txt)

1. Download as many of the above sample files
2. Open the data editor 

Image

3. Open a sample file 
4. Start profiling the file 

Image

5. Click on profile as CSV and save the csv file 

Image

6. Observe how the last item of the CSV is 256,0 

Image

JeremyYao:
Closes #1255

### DIFF
--- a/src/svelte/src/components/DataMetrics/DataMetrics.svelte
+++ b/src/svelte/src/components/DataMetrics/DataMetrics.svelte
@@ -27,6 +27,7 @@ limitations under the License.
   import Tooltip from 'src/components/layouts/Tooltip.svelte'
 
   const PROFILE_DOS_EOL = 256
+  const MAX_BYTE_VALUE = 255
 
   // title for the byte profile graph
   export let title: string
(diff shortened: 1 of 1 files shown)


scholarsmate (approved):
+1. The profiler is using the 256 value for the 2-byte DOS EOL, but does not need to be included in the CSV.

shanedell (approved):
+1

PR #1260, opened 2025-05-15
Remove people who will no longer be collaborators or are now PMC members from .asf.yaml

shanedell:
Remove people who will no longer be collaborators or are now PMC members from .asf.yaml


stricklandrbls (approved):
+1

PR #1262, opened 2025-05-19
Update version to 1.4.1, in preparation for rc1

shanedell:


JeremyYao (approved):

rthomas320 (approved):

hdalsania (approved):
+1

stevedlawrence (approved):
+1

FYI, when doing the 3.11.0 Daffodil release we found that Nexus authentication tokens are not correct and need updates. We've opened a ticket with infra. Note that I don't think this affects vscode since it doesn't publish to nexus, but I think you're still waiting on the vsce reproducible build fix. But all projects are essentially blocked from using the new release process for now. The old process should be used instead, here:

https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=81789390

Note that we found one issue with the container that has been fixed, so make sure to rebuild the container with the latest daffodil source.

Also, because the old workflow pushes a tag, that will cause the new workflow to run, which will likely fail or possibly delete files created by the release candidate container. So for now, I've manually disabled the release-candidate workflow for all projects so it shouldn't interfere. We'll have to manually reenable it once we are ready for the next releases--hopefully all issues will be resolved by then.

PR #1295, opened 2025-06-09
Add CI job for checking node licenses

shanedell:


stevedlawrence (approved):
+1, definitely simplifies verifying the complex node dependencies

Note that it is still important to do other kinds of license checks that this doesn't cover.

For example, dev dependencies still must still meet the [ASF license policy](https://www.apache.org/legal/resolved.html), they just don't need to be listed in LICENSE/NOTICE files. There are also sbt libraryDependencies that need to be considered. And when dependencies are updated we do still need to confirm that those licenses haven't changed to something that is no-longer ASF compatible or to something that has different requirements (e.g. LICENSE vs NOTICE inclusion). Those kinds of checks probably still need to be done manually.

stevedlawrence:
Sounds reasonable to me

JeremyYao:
@shanedell, another tool could be https://www.npmjs.com/package/licensee. I haven't done any investigation on this currently. From the looks of it, having a customizable .licensee.json file would be nice to have.

JeremyYao:
I wonder too if it's possible to automate the creation of build/package/LICENSE and build/package/NONLICENSE. Would it be feasible to scan through the node_modules folder and somehow use the output from the license checker tools.

shanedell:
@stevedlawrence I did make that other CI job. Currently its a little hard coded on which LICENSES it allows through. I plan to look over at the ASF allow 3rd libraries and trying to figure out there code of node

rthomas320 (approved):
+1

PR #1313, opened 2025-06-30
Fix issue with TDML config object spread and have checks for empty st…

### LINKED ISSUE #1024: In Launch config wizard,  TDML name, description and infoset path is defaults to string "undefined"
In Launch config wizard,  TDML name, description and infoset path is defaults to string "undefined".  We either should have a default values or ability to not set them at all, so default values can be used. With default "undefined" default value, it is treated as string "undefined" when this value is interpreted in code.

JeremyYao:
…ring instead.

Closes #1024


JeremyYao:
image

Tdml config fields get populated with non `"undefined"` values when creating a new config

JeremyYao:
Question: Should we keep the behavior for 'undefined'. Ie: having checks to substitute it in or should we remove those checks?

CoverRyan (approved):

PR #1323, opened 2025-07-08
Update debugger to use Daffodil 3.11 and Scala 2.13

### LINKED ISSUE #1320: Update to Apache Daffodil 3.11
We are currently using version 3.10.0, Version 3.11.0 is the latest daffodil version as of now, we should update our package and ensure functionally of the extension is still stable. This update has breaking changes with newer scala version support. This will require to update the scala code within the extension codebase to be compatible with Scala 2.13

shanedell:
Update debugger to use Daffodil 3.11 and Scala 2.13

Closes #1320

### DIFF
--- a/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Compiler.scala
+++ b/debugger/src/main/scala/org.apache.daffodil.debugger.dap/Compiler.scala
@@ -22,7 +22,7 @@ import cats.syntax.all._
 import java.nio.file.Path
 import org.apache.daffodil.sapi._
 
-trait Compiler {
+trait DAPCompiler {
   def compile(
       schema: Path,
       rootName: Option[String],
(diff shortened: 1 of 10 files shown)


debugger/src/main/scala/org.apache.daffodil.debugger.dap/Compiler.scala
```diff
@@ -22,7 +22,7 @@ import cats.syntax.all._
 import java.nio.file.Path
 import org.apache.daffodil.sapi._
 
-trait Compiler {
+trait DAPCompiler {
```
shanedell, line 25: With Scala 2.13 using the name `Compiler` cashed a crash with `fs2.Compiler` which is imported in the `Parse.scala` file using `fs2._` so it needed renamed. If there is a different name we would rather use I can change it

scholarsmate: I think DAPCompiler is fine.

scholarsmate (approved):
+1

JeremyYao (approved):

PR #1334, opened 2025-07-23
Corrects a logic error in determining a multi-line tag

### LINKED ISSUE #1332: Intellisense hangs trying to determine the attributes to suggest on a line with multiple XML tags
Intellisense hangs trying to determine the attributes to suggest on a line with multiple XML tags.  If a single line contains multiple XML tags and intellisense is triggered at a tag other that the first tag, the intellisense process will hang.

Ex:

Image

### LINKED ISSUE #1333: Intellisense fails to suggest items after closing tag of a multi-line XML tag
Intellisense fails to suggest items after closing tag of a multi-line XML tag.

Ex:

Image

rthomas320:
corrects a problem with the logic to determine if an XML tag is spread across multiple lines
corrects a problem where a cursor after an end tag is not recognized correctly in some instances

closes #1332
closes #1333

### DIFF
--- a/src/language/providers/utils.ts
+++ b/src/language/providers/utils.ts
@@ -273,11 +273,13 @@ export function checkTagOpen(
     }
   }
 
+  // If there is one or less xml tags on the line and the beginning of the line
+  // contains the start tag and there is not an ending tag character after the
+  // start tag assume it is a multi-line tag
   if (
-    !(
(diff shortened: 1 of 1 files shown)


CoverRyan (approved):

shanedell (approved):
+1 I also like @stricklandrbls suggestion if this is something that can be incorporated into a different PR

Issue #1030, opened 2024-05-14
TDML Execute functionality - should only require TDML test case name only to execute


hdalsania:
TDML Execute functionality - should only require TDML test case name only to execute, currently it requires name and description.

lrbarber:
Could this just provide a pulldown list of the test cases in the TDML file?

JeremyYao:
Per conversations today, @michael-hoke you may want to look at https://github.com/apache/daffodil-vscode/pull/1336 to see what I've done so far as well as note areas where there may be merge conflicts.

PR #1336, opened 2025-07-29
Prompt for TDML test case name and description before starting the de…

### LINKED ISSUE #1029: TDML Execute - shorter server timeout to enter name/description for TDML test case
TDML Execute Functionality - if the user takes longer than the timeout to enter name/description, the server will disconnect. 

Check issue #596 , this issue may fix this issue if implemented.

JeremyYao:
…bugger.

Closes #1029

For information regarding how I tested this. Please see https://github.com/apache/daffodil-vscode/pull/1336#issuecomment-3135703245. Adapt the instructions to your testing needs if needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -579,12 +579,12 @@
                   },
                   "name": {
                     "type": "string",
-                    "default": "${command:AskForTDMLName}",
+                    "default": "",
                     "description": "TDML Case Name"
                   },
                   "description": {
                     "type": "string",
-                    "default": "${command:AskForTDMLDescription}",
+                    "default": "",
                     "description": "TDML Case Description"
                   },
                   "path": {
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -137,10 +137,10 @@ async function showTDMLSaveDialog(fileRequested, label, title) {
 }
 
 // Function for setting up the commands for Run and Debug file
-function createDebugRunFileConfigs(
+async function createDebugRunFileConfigs(
   resource: vscode.Uri,
   runOrDebug: String,
-  tdmlAction: String | undefined,
+  tdmlAction: string | undefined,
   runLast = false
 ) {
   let targetResource = resource
@@ -168,35 +168,52 @@ function createDebugRunFileConfigs(
         tdmlConfig.action = tdmlAction
 
         if (tdmlAction === 'execute') {
-          tdmlConfig.name = '${command:AskForTDMLName}'
-          tdmlConfig.description = '${command:AskForTDMLDescription}'
-          tdmlConfig.path = targetResource.fsPath
+          tdmlConfig.path = targetResource.fsPath.toString()
+
+          tdmlConfig.name =
+            tdmlConfig.name ||
+            (await vscode.commands.executeCommand(
+              'extension.dfdl-debug.getTDMLName',
+              tdmlConfig.path
+            ))
+
+          tdmlConfig.description =
+            tdmlConfig.description ||
+            (await vscode.commands.executeCommand(
+              'extension.dfdl-debug.getTDMLDescription',
+              tdmlConfig.path
+            ))
+
+          if (!tdmlConfig.name || !tdmlConfig.description) {
+            vscode.window.showInformationMessage(
+              'ESC pressed on a TDML Test Case selection picker! Stopping debugging.'
+            )
+            return
+          }
         }
       }
 
-      vscode.debug.startDebugging(
-        undefined,
-        getConfig({
-          name: 'Run File',
-          request: 'launch',
-          type: 'dfdl',
-          schema: {
-            path: tdmlConfig.action === 'execute' ? '' : targetResource.fsPath,
-            rootName: null,
-            rootNamespace: null,
-          },
-          data:
-            tdmlConfig.action === 'execute' ? '' : '${command:AskForDataName}',
-          debugServer: false,
-          infosetFormat: 'xml',
-          infosetOutput: {
-            type: 'file',
-            path: '${workspaceFolder}/' + infosetFile,
-          },
-          tdmlConfig: tdmlConfig,
-        }),
-        { noDebug: noDebug }
-      )
+      const config = getConfig({
+        name: 'Run File',
+        request: 'launch',
+        type: 'dfdl',
+        schema: {
+          path: tdmlConfig.action === 'execute' ? '' : targetResource.fsPath,
+          rootName: null,
+          rootNamespace: null,
+        },
+        data:
+          tdmlConfig.action === 'execute' ? '' : '${command:AskForDataName}',
+        debugServer: false,
+        infosetFormat: 'xml',
+        infosetOutput: {
+          type: 'file',
+          path: '${workspaceFolder}/' + infosetFile,
+        },
+        tdmlConfig: tdmlConfig,
+      })
+
+      vscode.debug.startDebugging(undefined, config, { noDebug: noDebug })
     }
 
     vscode.debug.onDidTerminateDebugSession(async () => {
@@ -399,24 +416,22 @@ export function activateDaffodilDebug(
   context.subscriptions.push(
     vscode.commands.registerCommand(
       'extension.dfdl-debug.getTDMLName',
-      async (config) => {
+      async (tdmlConfigPath) => {
         // Config item gets passed
 
         // get test case name options for dropdown
-        const test_case_names: string[] = getTDMLTestCaseItems(
-          config?.tdmlConfig?.path
-        )
+        const test_case_names: string[] = getTDMLTestCaseItems(tdmlConfigPath)
           .filter((obj) => typeof obj.name == 'string')
           .map((obj) => obj.name) as string[]
 
-        // dropdown
-        return await vscode.window
-          .showQuickPick(test_case_names, {
-            placeHolder: 'Test Case Name',
-          })
-          .then((value) => {
-            return value // return selected dropdown value
-          })
+        if (test_case_names.length == 0) {
+          return
+        }
+
+        // Await showQuickPick directly and return the result
+        return await vscode.window.showQuickPick(test_case_names, {
+          placeHolder: 'Test Case Name',
+        })
       }
     )
   )
@@ -424,26 +439,29 @@ export function activateDaffodilDebug(
   context.subscriptions.push(
     vscode.commands.registerCommand(
       'extension.dfdl-debug.getTDMLDescription',
-      async (config) => {
+      async (tdmlConfigPath) => {
         // Config item gets passed
 
         // get unique test case description options for dropdown as it may be possible for overlapping description
         const test_case_unique_descriptions: string[] = [
           ...new Set(
-            getTDMLTestCaseItems(config?.tdmlConfig?.path)
+            getTDMLTestCaseItems(tdmlConfigPath)
               .filter((obj) => typeof obj.description == 'string')
               .map((obj) => obj.description)
           ),
         ] as string[]
 
-        // dropdown
-        return await vscode.window
-          .showQuickPick(test_case_unique_descriptions, {
+        if (test_case_unique_descriptions.length == 0) {
+          return
+        }
+
+        // Await showQuickPick directly and return the result
+        return await vscode.window.showQuickPick(
+          test_case_unique_descriptions,
+          {
             placeHolder: 'Test Case Description',
-          })
-          .then((value) => {
-            return value // return selected dropdown value
-          })
+          }
+        )
       }
     )
   )
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -441,11 +441,26 @@ export function getTDMLTestCaseItems(
   const xml_obj = parser.parse(fileData)
 
   // Read through TDML test cases and populate each TDML test case item if XML file is valid enough
-  if (Array.isArray(xml_obj['testSuite']?.['parserTestCase'])) {
+  if (
+    // Otherwise, it's an array
+    Array.isArray(xml_obj['testSuite']?.['parserTestCase'])
+  ) {
     return xml_obj['testSuite']['parserTestCase'].map((obj) => ({
       name: obj['@_name'],
       description: obj['@_description'],
     }))
+  } else if (
+    // One item results in an object type
+    xml_obj['testSuite']?.['parserTestCase'] &&
+    typeof xml_obj['testSuite']?.['parserTestCase'] == 'object'
+  ) {
+    const obj = xml_obj['testSuite']?.['parserTestCase']
+    return [
+      {
+        name: obj['@_name'],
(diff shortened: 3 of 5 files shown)


src/adapter/activateDaffodilDebug.ts
```diff
@@ -168,35 +168,50 @@ function createDebugRunFileConfigs(
...
+        config.tdmlConfig.name = await vscode.commands.executeCommand(
+          'extension.dfdl-debug.getTDMLName',
+          config
+        )
+
+        config.tdmlConfig.description = await vscode.commands.executeCommand(
+          'extension.dfdl-debug.getTDMLDescription',
+          config
+        )
+      }
```
michael-hoke, line 213: I don't like this block here. We're already doing some of these modifications up above - I'd prefer to see us modify tdmlConfig before startDebugging is called, then pass it into the config object without further modifications... or get rid of the above block, but I don't see a good reason to do that.

Also, those filename checks are explicitly done already. Unless somebody is going to manually override VSCode's file type selection, we're already automatically detecting TDML files and setting the file type to TDML. Then, we're limiting the 'Execute TDML' command to only be present when the file type is TDML.

src/adapter/activateDaffodilDebug.ts
```diff
@@ -168,35 +168,50 @@ function createDebugRunFileConfigs(
...
+      // If we're calling execute TDML from a .tdml file
+      // set the name and the description of the TDML test case we want
+      // before running the debugger
+      if (
+        typeof config.tdmlConfig?.path === 'string' &&
+        path.extname(config.tdmlConfig?.path).toLowerCase() == '.tdml' &&
+        typeof config.tdmlConfig?.action === 'string' &&
+        config.tdmlConfig?.action.toLowerCase() == 'execute'
+      ) {
+        config.tdmlConfig.name = await vscode.commands.executeCommand(
```
michael-hoke, line 204: I don't see how this is any different than what we were doing before. We're `await`ing the command, and the debugger starts while the user is meant to be typing/selecting. When you run the 'Execute TDML' command, can you let the prompt sit for 30 seconds without the backend crashing? Does the debugger start before you've made your selection?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -168,35 +168,50 @@ function createDebugRunFileConfigs(
...
+          'extension.dfdl-debug.getTDMLName',
+          config
```
michael-hoke, line 206: I think there's a really strong argument to changing this from config to the path

src/adapter/activateDaffodilDebug.ts
```diff
@@ -137,7 +137,7 @@ async function showTDMLSaveDialog(fileRequested, label, title) {
 }
 
 // Function for setting up the commands for Run and Debug file
-function createDebugRunFileConfigs(
+async function createDebugRunFileConfigs(
```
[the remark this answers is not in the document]
michael-hoke: I think I have to change my mind on this one. I don't think there's a way to avoid making this async. However, this means we need to look at the callers. From what I see, the caller (activateDaffodilDebug) is non-async, but its caller (activate) is. This means we should be able to flip all of the non-async functions to async.

src/adapter/activateDaffodilDebug.ts
michael-hoke: I partially take this back. I think using the awaits here probably will be okay, but you'll have to use the then to set the name. I'm doing something very similar on my setup, and I'm still getting the timeout. I'll test explicitly with this code tomorrow to see if there are any differences.

JeremyYao: Sounds like a plan. I posted my testing methodology as a comment (https://github.com/apache/daffodil-vscode/pull/1336#issuecomment-3135703245) if you would like to refer to it.

src/adapter/activateDaffodilDebug.ts
JeremyYao: Thanks for the notes. 


### First Paragraph

Can you expand upon what you mean by `I'd prefer to see us modify tdmlConfig before startDebugging is called`? 

My understanding of the block of code and rationales for my changes are as follows: `tdmlConfig` appears to be an object used for organizing the object we're passing into `getConfig()` (see attached code snippet below).

 The if statements are a safety check and may be warranted in the event of an edge case where `config.tdmlConfig.path` point to a non-tdml file (example: a file w/o the .tdml extension). 

```TypeScript
      var tdmlConfig = <TDMLConfig>{}

      if (tdmlAction) {
        tdmlConfig.action = tdmlAction

        if (tdmlAction === 'execute') {
          tdmlConfig.path = targetResource.fsPath
        }
      }

      const config = getConfig({
        name: 'Run File',
        request: 'launch',
        type: 'dfdl',
        schema: {
          path: tdmlConfig.action === 'execute' ? '' : targetResource.fsPath,
          rootName: null,
          rootNamespace: null,
        },
        data:
          tdmlConfig.action === 'execute' ? '' : '${command:AskForDataName}',
        debugServer: false,
        infosetFormat: 'xml',
        infosetOutput: {
          type: 'file',
          path: '${workspaceFolder}/' + infosetFile,
        },
        tdmlConfig: tdmlConfig,
      })

      // If we're calling execute TDML from a .tdml file
      // set the name and the description of the TDML test case we want
      // before running the debugger
      if (
        typeof config.tdmlConfig?.path === 'string' &&
        path.extname(config.tdmlConfig?.path).toLowerCase() == '.tdml' &&
        typeof config.tdmlConfig?.action === 'string' &&
        config.tdmlConfig?.action.toLowerCase() == 'execute'
      ) {
        config.tdmlConfig.name = await vscode.commands.executeCommand(
          'extension.dfdl-debug.getTDMLName',
          config
        )

        config.tdmlConfig.description = await vscode.commands.executeCommand(
          'extension.dfdl-debug.getTDMLDescription',
          config
        )
      }

      vscode.debug.startDebugging(undefined, config, { noDebug: noDebug })
```

### Second paragraph

Could you point to where the filename checks are being handled currently? I can cross-verify and make changes accordingly. As you mentioned, `Unless somebody is going to manually override VSCode's file type selection, we're already automatically detecting TDML files and setting the file type to TDML`, the changes could be good to include in there for catching edge cases when someone overrides file type detection. 

We should definitely include checks as you mentioned at the end. Accounting for the scenario the user attempts to execute `Execute TDML` for non-TDML files.

src/adapter/activateDaffodilDebug.ts
michael-hoke: We're passing config into the command, but we're only using config.tdmlConfig.path - why not just pass that in?

JeremyYao:
From conversations with @michael-hoke, he will be incorporate changes in this PR for his PR for https://github.com/apache/daffodil-vscode/issues/1030 as changes in this PR and his PR may conflict with each other.

package.json
```diff
@@ -829,7 +829,6 @@
         "variables": {
           "AskForSchemaName": "extension.dfdl-debug.getSchemaName",
           "AskForDataName": "extension.dfdl-debug.getDataName",
-          "AskForTDMLName": "extension.dfdl-debug.getTDMLName",
```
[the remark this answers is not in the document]
michael-hoke: For this to be used again, we would need to call it like this: `${command:AskForTDMLName}` and I don't see a way to pass the argument in like that.

src/adapter/activateDaffodilDebug.ts
michael-hoke: There should also be an "error" check if the quick pick doesn't have any values to display. It should tell the user that no TDML test cases were found and it should not do the debugging.

package.json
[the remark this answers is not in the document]
shanedell: Yes it is the default, without this variable the `${command:AskForTDMLName}` will not function properly. So we either to need make it not the default or add it back in. The user can still set the value to be `${command:AskForTDMLName}` which once its seen on start up it would be parsed but maybe we don't want that anymore?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -410,13 +425,12 @@ export function activateDaffodilDebug(
...
-          })
-          .then((value) => {
-            return value // return selected dropdown value
-          })
+        // Await showQuickPick directly and return the result
+        const selection = await vscode.window.showQuickPick(test_case_names, {
+          placeHolder: 'Test Case Name',
+        })
+
+        return selection
```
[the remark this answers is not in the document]
shanedell: @michael-hoke I'm not sure what you mean by not being okay with the update to the if statement. The only change happening is making sure the type of the attributes of the TDMLConfig class are `string` and not `String`. This makes it so you don't have to do `typeof tdmlConfig.path === 'string'` or `typeof tdmlConfig.action === 'string'` as both `path` and `action` should only be either a `string` or `undefined`. So, we should not have to strictly check that type for either `path` and `action`, currently you have to do that since its `String | undefined`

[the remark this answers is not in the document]
shanedell: @michael-hoke So are you saying to do something more along the line of the code below?

```ts
      var tdmlConfig = <TDMLConfig>{}

      if (tdmlAction) {
        tdmlConfig.action = tdmlAction

        if (tdmlAction === 'execute') {
          tdmlConfig.path = targetResource.fsPath

          tdmlConfig.name = await vscode.commands.executeCommand(
            'extension.dfdl-debug.getTDMLName',
            tdmlConfig
          )

          tdmlConfig.description = await vscode.commands.executeCommand(
            'extension.dfdl-debug.getTDMLDescription',
            tdmlConfig
          )
        }
      }

      const config = getConfig({
        name: 'Run File',
        request: 'launch',
        type: 'dfdl',
        schema: {
          path: tdmlConfig.action === 'execute' ? '' : targetResource.fsPath,
          rootName: null,
          rootNamespace: null,
        },
        data:
          tdmlConfig.action === 'execute' ? '' : '${command:AskForDataName}',
        debugServer: false,
        infosetFormat: 'xml',
        infosetOutput: {
          type: 'file',
          path: '${workspaceFolder}/' + infosetFile,
        },
        tdmlConfig: tdmlConfig,
      })

      vscode.debug.startDebugging(undefined, config, { noDebug: noDebug })
```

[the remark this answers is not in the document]
michael-hoke: basically, yes, in addition to the string -> String updates

src/adapter/activateDaffodilDebug.ts
JeremyYao: That's a good point. I'll go ahead and modifying it to take in config.tdmlConfig.path

JeremyYao: Adressed

package.json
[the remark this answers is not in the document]
shanedell: @JeremyYao @michael-hoke Mean't keep this removed but update the `configurationAttributes`, `launch` config default for the `TDML Case Name` to be `""` instead of `${command:AskForTDMLName}`

src/adapter/activateDaffodilDebug.ts
JeremyYao: Added error checking via the vscode window showing an error message and returning, skipping starting debugging

src/adapter/activateDaffodilDebug.ts
```diff
@@ -168,35 +168,62 @@ function createDebugRunFileConfigs(
...
+                'Error: no TDML test cases were found! Stopping debugging.'
+              )
+              return
+            }
+          } else {
+            vscode.window.showErrorMessage(
+              "Error: TDML file specified doesn't have correct TDML file extension ().tdml)!"
+            )
+            return
+          }
```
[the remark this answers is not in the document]
JeremyYao: Removed .tdml file extension check. Removed ? from `tdmlConfig`s

src/adapter/activateDaffodilDebug.ts
```diff
@@ -399,51 +426,44 @@ export function activateDaffodilDebug(
...
       'extension.dfdl-debug.getTDMLName',
-      async (config) => {
+      async (tdmlConfigPath) => {
         // Config item gets passed
 
         // get test case name options for dropdown
-        const test_case_names: string[] = getTDMLTestCaseItems(
-          config?.tdmlConfig?.path
-        )
+        const test_case_names: string[] = getTDMLTestCaseItems(tdmlConfigPath)
```
michael-hoke, line 423: If this returns an empty list, we shouldn't even be calling the getTDMLName (or Description) functions. We should be returning empty immediately, and the caller should see that and abort.

src/utils.ts
```diff
@@ -441,11 +441,26 @@ export function getTDMLTestCaseItems(
...
+    typeof xml_obj['testSuite']?.['parserTestCase'] == 'object'
+  ) {
+    const obj = xml_obj['testSuite']?.['parserTestCase']
+    return [
+      {
+        name: obj['@_name'],
+        description: obj['@_description'],
+      },
+    ]
   } else {
```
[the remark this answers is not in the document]
JeremyYao: From my perspective, chaining ?s and :s makes the code look harder to read and understand.

src/adapter/activateDaffodilDebug.ts
```diff
@@ -168,35 +168,52 @@ function createDebugRunFileConfigs(
...
+
+          tdmlConfig.description =
+            tdmlConfig.description ||
+            (await vscode.commands.executeCommand(
+              'extension.dfdl-debug.getTDMLDescription',
+              tdmlConfig.path
+            ))
+
+          if (!tdmlConfig.name || !tdmlConfig.description) {
+            vscode.window.showErrorMessage(
```
michael-hoke, line 188: I'm not sure that we want this error message here. It makes sense to have one in the case where no test cases are found, but it doesn't make sense if the user presses Esc. Maybe we can meet in the middle and do a toast message in the bottom right, the same way we do for the Infoset being created?

michael-hoke:
Just as a sanity check, can you try using a launch config where tdmlConfig.path is set to a non-existent file. The getTDMLTestCases function should be able to handle that properly, but I want to be sure.

src/utils.ts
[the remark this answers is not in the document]
JeremyYao: Lets hear others' perspectives on this.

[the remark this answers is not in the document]
michael-hoke: I'm not asking you to make any changes in this PR. I was just noting the upcoming code changes here. Any conversation can be done on the upcoming PR.

src/adapter/activateDaffodilDebug.ts
JeremyYao: Addressed in https://github.com/apache/daffodil-vscode/pull/1336/commits/ea0b4013c405d3cdcf7991cbea3a051e2cd079b8

michael-hoke (approved):

shanedell:
@JeremyYao Please squash the commits and then I will approve

PR #1345, opened 2025-08-04
Disable popup for Schema Definition Errors

### LINKED ISSUE #58: SDE should not be a blocking dialog
It would be better if SDEs were reported on the terminal window, not in a blocking dialog. Also multiple SDEs are often reported at once, and users will want to see all existing.

michael-hoke:
Closes #58 

Removes the popup in cases where invalid XML is sent to the parser. An Output window is created that contains the error text, and it is also contained in the terminal window.

### DIFF
--- a/src/adapter/daffodilEvent.ts
+++ b/src/adapter/daffodilEvent.ts
@@ -32,8 +32,11 @@ export function handleDebugEvent(e: vscode.DebugSessionCustomEvent) {
       break
     // this allows for any error event to be caught in this case
     case e.event.startsWith('daffodil.error') ? e.event : '':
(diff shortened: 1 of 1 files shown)


lrbarber:
Works great. The error messages appear in the OUTPUT tab & it automatically switches to the OUTPUT tab.
Text includes a link to the file where the error occurs. The location (line/column) is included in the text, but, unfortunately, **not** the link (maybe a future enhancement effort?).
Also (and I think this is OK), if you do something stupid like click cancel in the dialog box that is asking you to select a file, a pop-up is generated:
image

CoverRyan (approved):

PR #1346, opened 2025-08-04
Remove dependency on Highlight Matching Tag extension

### LINKED ISSUE #1335: Extension will not activate when using VSCode with a Remote - SSH connection
When using VSCode over ssh, there are two ways that VSCode can install an extension - locally or remotely. This extension should be installed remotely because it touches the filesystem. However, one of the extensions that we depend on is UI-only, so it prefers to install locally. Since VSCode checks the installed extensions for dependencies on the side that it is installing an extension on, it checks remotely for the Highlight-Matching-Tag extension, which we have as a dependency. This causes an error popup to appear and the DFDL extension fails to active. The error popup suggests to reload the window, but that causes the same behavior.

I don't believe we actually depend on this extension - it was just a nice to have/extra extension to make writing DFDL schemas easier. Removing it as a dependency resolves this issue.

michael-hoke:
Closes #1335 

We had a dependency on the Highlight Matching Tag extension. We don't appear to actually be depending on this extension - it was just installed alongside ours as a quality of life extension for schemas. Removing it as a dependency allows the DFDL extension to install over an ssh connection, and the Highlight Matching Tag extension will appear in the list of recommended extensions if users wish to install it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,6 @@
     "webpack-cli": "6.0.1"
   },
   "extensionDependencies": [
-    "vincaslt.highlight-matching-tag",
     "wmanth.jar-viewer"
   ],
   "main": "./dist/ext/extension.js",
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -228,7 +228,6 @@ Upon opening, VSCode may prompt you to install recommended extensions. Go ahead 
 Daffodil-VSCode depends on the following extensions
 
 - [JAR Viewer - Visual Studio Marketplace](https://marketplace.visualstudio.com/items?itemName=wmanth.jar-viewer)
-- [Highlight Matching Tag - Visual Studio Marketplace](https://marketplace.visualstudio.com/items?itemName=vincaslt.highlight-matching-tag)
 
 ### Verifying Setup Can Build
 
--- a/src/tests/runTest.ts
+++ b/src/tests/runTest.ts
@@ -60,21 +60,11 @@ async function main() {
       resolveCliArgsFromVSCodeExecutablePath(vscodeExecutablePath)
 
     // Install required extensions
-    cp.spawnSync(
-      cli,
-      [
-        ...args,
-        '--install-extension',
-        'vincaslt.highlight-matching-tag',
-        '--install-extension',
-        'wmanth.jar-viewer',
-      ],
-      {
-        encoding: 'utf-8',
-        stdio: 'inherit',
-        shell: os.platform().toLowerCase() != 'darwin',
-      }
-    )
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -104,7 +104,6 @@
     "webpack-cli": "6.0.1"
   },
   "extensionDependencies": [
-    "vincaslt.highlight-matching-tag",
     "wmanth.jar-viewer"
```
shanedell, line 107: Is this `jar-viewer` also necessarily required as a dependency either or could it be a nice to have as well?

JeremyYao (changes requested):
You may want to consider removing mentions of the extension in other places such as 

- Development.md (ln 231)
- .vscode/extensions.json
- src/tests/runTest.ts (ln 68)

michael-hoke:
Good call on the md change.

extensions.json contains recommended extensions that will show up in the extensions pane when our extension is installed - we want it to show up there.

I believe that we do want it to remain in the tests. If we are recommending to users that they can use this extension, we should be sure that our extension works when this one is installed.

JeremyYao (approved):

shanedell:
@michael-hoke Also remove this extension from being installed in the `src/tests/runTest.ts`. This updated should be something like below:

```ts
    cp.spawnSync(
      cli,
      [
        ...args,
        '--install-extension',
        'vincaslt.highlight-matching-tag',
        '--install-extension',
        'wmanth.jar-viewer',
      ],
      {
        encoding: 'utf-8',
        stdio: 'inherit',
        shell: os.platform().toLowerCase() != 'darwin',
      }
    )
```

becomes:

```ts
    cp.spawnSync(
      cli,
      [
        ...args,
        '--install-extension',
        'wmanth.jar-viewer',
      ],
      {
        encoding: 'utf-8',
        stdio: 'inherit',
        shell: os.platform().toLowerCase() != 'darwin',
      }
    )
```

michael-hoke:
> @michael-hoke Also remove this extension from being installed in the `src/tests/runTest.ts`. This updated should be something like below:

@shanedell If we're recommending this extension be used with ours, shouldn't we leave it in the tests so that we're sure that it will work with our extension? I wouldn't want to end up getting any bugs with the integration between the two extensions.

michael-hoke:
@shanedell Okay, makes sense. Updated and squashed

shanedell (approved):
+1

I believe we should also check into the `jar-viewer`, meaning is it a required dependency or just recommended.

michael-hoke:
Agreed - thought we were using this one, but I can't see anything obvious. Created an issue for it here #1351

rthomas320 (approved):
Tested on Windows 10. Works as expected.
+1

PR #1354, opened 2025-08-08
Remove Description from being needed in the TDML config

### LINKED ISSUE #1030: TDML Execute functionality - should only require TDML test case name only to execute

michael-hoke:
Closes #1030 

Also removed Description from the launch config, even for generate. It doesn't really make sense to have a field where a user specifies text to go into a file that may be overwritten/deleted. We should expect the users to copy/append the generated TDML file, then they will need to go into that file to create a Zip. At that point, they are able to see and edit the descriptions. Right now, it has to be edited in the XML, however, we will eventually have support for doing that in the GUI.

Removed dead TDML append-related code in the backend and fixed a bug where execute wouldn't work from the TDML GUI.


JeremyYao (changes requested):
One thought I had as I was testing this is that the omission of the TDML test case description in launch.config is that there may be cases where being able to input the description may be useful. 

For instance, describing specifics about a test case like "This test case tests this schemas' ability to detect EOL at the end of the file". 

Another thing I've noticed is that the TDML viewer still displays the test case description

image

This hints at the description could still be used as a way of adding comments about a test case. If we're still displaying the description, then it may worth re-adding the ability to specify the test cases' description. 

Otherwise, code looks good and functionality works as expected.

michael-hoke:
The reasons you bring up are exactly why we are leaving it in the GUI - we want the user to be able to supply meaningful information about each test case.

As for adding it back into the launch config, there are two reasons that we shouldn't want to do that:
- Names are the only way that the TDML spec states are used to identify a test case. The reason that the description was in there originally was that we were using both.
- We will be supplying a way for the description to be edited in the GUI. Right now, this is slated for 1.5.0, but it's possible that we bring it in to 1.4.2 depending on timelines. At the moment, a user could switch back to the XML view to modify the description. Although this isn't ideal, we also don't have the main use case (zip) completed yet, so most users aren't going to be wanting to interact with TDML files yet.

JeremyYao:
> The reasons you bring up are exactly why we are leaving it in the GUI - we want the user to be able to supply meaningful information about each test case.
> 
> As for adding it back into the launch config, there are two reasons that we shouldn't want to do that:
> 
> * Names are the only way that the TDML spec states are used to identify a test case. The reason that the description was in there originally was that we were using both.
> * We will be supplying a way for the description to be edited in the GUI. Right now, this is slated for 1.5.0, but it's possible that we bring it in to 1.4.2 depending on timelines. At the moment, a user could switch back to the XML view to modify the description. Although this isn't ideal, we also don't have the main use case (zip) completed yet, so most users aren't going to be wanting to interact with TDML files yet.

Makes sense. Thanks for expanding upon why the TDML Description field was removed from the launch config web view. 

Approved the PR.

src/adapter/activateDaffodilDebug.ts
```diff
@@ -143,12 +144,17 @@ async function createDebugRunFileConfigs(
   tdmlAction: string | undefined,
   runLast = false
 ) {
-  let targetResource = resource
+  let targetResource: vscode.Uri | undefined = resource
```
stricklandrbls, line 147: Can `targetResource` have the type defined in the function parameters as `vscode.Uri | undefined` to be consistent with the `tdmlAction: string | undefined` parameter in this function?

michael-hoke: I like the idea of this, but this does seem to open up a can of worms that I would really prefer to avoid in this PR. This call to this function (and, therefore the argument) comes from a function set up with the vscode.commands.registerCommand() function. The VSCode API will automatically provide a Uri as an argument to registered functions. However, it is apparently possible for the Uri to be undefined, which we are not handling in any of these functions.

That being said, I believe that we are likely restricting the use of these commands to cases where the Uri is always going to be defined (file opened and active in window). So, I'd agree that we should be handling the possibility of an undefined Uri, just in case there are holes in how we are exposing the commands to the user, but I think this should be a larger fix than what you've suggested.

stricklandrbls (approved):
+1

PR #1356, opened 2025-08-11
TDML Generate is not creating/updating the TDML file in /tmp

### LINKED ISSUE #1350: TDML Generate is not creating/updating the TDML file in /tmp
The config passed to the backend doesn't contain a tdmlConfig. We need to be sure that the default tdmlConfig is always being passed into the backend.

If tdmlConfig is missing, set it to generate
Add validation for name/path if they are missing

michael-hoke:
Closes #1350 

Set default tdmlConfig to be missing/undefined instead of empty. The default handling of this was already correct, but it didn't want to "overwrite" the tdmlConfig if it was defined but empty.

Tested with and without #1354 changes. Should be able to be merged in either order, but should prefer to do this one second as it's much smaller.

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -162,10 +162,10 @@ async function createDebugRunFileConfigs(
         noDebug: noDebug,
       })
     } else {
-      var tdmlConfig = <TDMLConfig>{}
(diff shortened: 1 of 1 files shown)


rthomas320 (approved):
Tested on Windows 10. Works as described.
+1

PR #1359, opened 2025-08-13
Fix error handling for TDML inputs

### LINKED ISSUE #1349: TDML Execute doesn't work when fields are missing from launch config
Once the description is removed, there are three fields in the tdmlConfig (action, path, and name). Action will always be required. If path is missing, the user should be prompted for the tdml file. If name is missing, we should take the path from the config, extract the names from that TDML file and prompt the user for which name to use.

michael-hoke:
Closes #1349

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -175,17 +175,6 @@ async function createDebugRunFileConfigs(
 
         if (tdmlAction === 'execute') {
           tdmlConfig.path = targetResource.fsPath
-
-          tdmlConfig.name =
-            tdmlConfig.name ||
-            (await vscode.commands.executeCommand(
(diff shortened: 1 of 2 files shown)


src/adapter/activateDaffodilDebug.ts
```diff
@@ -175,17 +175,6 @@ async function createDebugRunFileConfigs(
...
-          tdmlConfig.name =
-            tdmlConfig.name ||
-            (await vscode.commands.executeCommand(
-              'extension.dfdl-debug.getTDMLName',
-              tdmlConfig.path
-            ))
-
-          if (!tdmlConfig.name) {
-            return
-          }
```
shanedell, line 188: Why was this code block removed?

rthomas320 (approved):
Tested on Windows 10. Works as expected.
+1

shanedell (approved):
+1

Issue #951, opened 2024-01-25
Ωedit server fails to start if PID file exists


stricklandrbls:
When starting up the data editor an Ωedit API call to start the server is invoked in the local `startServer` function. The `stopServer` function is invoked immediately which:
1. Computes the PID file string.
2. Checks if the PID file exists on the filesystem.
3. Calls the Ωedit API function `stopServerUsingPID(...)` which returns a boolean.
    - `True` will remove the checkpoint directory path and report the server successfully stopped.
    - `False` will report an error stating that the server cannot be stopped. 

#### Issue
The issue is that if the PID file exists but no server is running then the data editor initialization stops.

#### Workaround
The current workaround is to delete the PID file or directory containing the PID file (APP_DATA_PATH + `/omega_edit/`).

shanedell:
@stricklandrbls I am not sure if this is the same issue I am seeing but when I am running `yarn test` it seems that the omega-edit server is not being killed. Is this under this issue or separate?

stricklandrbls:
> @stricklandrbls I am not sure if this is the same issue I am seeing but when I am running `yarn test` it seems that the omega-edit server is not being killed. Is this under this issue or separate?

It's possible, there are currently a couple of items that can cause the omega edit server to fail to shut down. Most often than not it's because of the improper disposal of the data editor resources, which I am about to put a PR up for review. Much less common, I've only had this happen once, is in reference to this issue I created.

If it is this issue than a VSCode error notification box appears and states, `"Ωedit server on port <port> with PID <pid> failed to stop"`. Though, given that you're seeing this issue from running the test script then it may not appear.

scholarsmate:
Additional lifecycle tests were recently added to Ωedit in addition to work on the server side to get graceful and immediate shutdowns to work properly cross-platform using API calls rather than sending signals from the OS (see: https://github.com/ctc-oss/omega-edit/commit/54704eafb4cd4371ebf5ec9a4ce2cdef48757270#diff-240f09e4fc40139a1cdd8c9ab7bb29a2399e3a09c466ca3326ea7db83de657cf)

stricklandrbls:
@CoverRyan - Added the replication steps to the description.

PR #1364, opened 2025-08-19
Prevent Stale PID File

CoverRyan:
Closes(#951)
**To Test:**
1. Navigate to the APP_DATA_PATH location on the filesystem.
2. Delete all .pid files if any exist and ensure no extension or omega edit server is currently running.
3. Create an arbitrary .pid file with the omega edit server pidfile naming convention (i.e, serv-9000.pid)
4. Start the extension, either through a debug launch config or packaged extension execution
5. Open the Data Editor through the command palette


rthomas320 (approved):
Tested on Windows 10. Worked as expected.
+1

JeremyYao:
> @JeremyYao the server pid file is always created in the APP_DATA_DIR which for UNIX systems is _/home/<user>/.local/share/omega_edit/_ by default but given that it's not creating the pidfile in the default directory then the data editor should open fine.
> 
> I'll try and see what's going on.
> 
> You could also set the log output to debug in the package.json to see more verbose logging.

Good catch! Thank you! 

I'm not seeing any .pid file in the omega_edit folder

image

I'm not running into the viewport issue as depicted on the screenshot when I install from .vsix file.

stricklandrbls (approved):
+1 Tested on Ubuntu 24.04

@JeremyYao - I suspect that there must be some pathing issues with the _serv-9000.pid_ file that must be created prior to running the extension.

I was able to create a dummy .pid, start the extension, verify removal of the dummy .pid, and see the Data Editor execute as normal.

scholarsmate (dismissed):
`ps -a | grep ${pid} | grep -v grep | awk '{print $2}'` is incorrect, please fix.

CoverRyan:
> Currently getting this error dialog when running the extension: image
> 
> **NOTE:** This was ran on a Windows OS that utilized VSCode's ssh remote tunnel on to a Ubuntu 24.04 machine.

Added error handling. It seems ps returns 1 if process doesn't exist resulting in execSync throwing a command failed error.

PR #1372, opened 2025-08-22
1355 - return if user cancels during file prompts

### LINKED ISSUE #1355: A Daffodil parse command should exit before starting up the backend if the user cancels the prompts
If a user starts a daffodil parse, through the command palette or launch config, they may get a input dialog of some kind asking them to provide the dfdl schema or the data file. If they press escape or cancel out of the dialog, the backend is started, we send the launch config to the backend, and then the backend throws an exception and creates an error dialog. If the input is cancelled, we shouldn't even be starting the backend as we know that it's an invalid config.

See how we're handling this on the TDML side - if either of the fields are empty, we should not start debugging. They can be empty from a user cancelling the dialog or from the launch config.

Also consider whether we want to check whether the provided files exist before we start the debugger.

duboisfordwork:
## Problem Description
Closes #1355 
Quoted from [issue 1355](https://github.com/apache/daffodil-vscode/issues/1355)
"If a user starts a daffodil parse, through the command palette or launch config, they may get a input dialog of some kind asking them to provide the dfdl schema or the data file. If they press escape or cancel out of the dialog, the backend is started, we send the launch config to the backend, and then the backend throws an exception and creates an error dialog. If the input is cancelled, we shouldn't even be starting the backend as we know that it's an invalid config."

## Testing
1. Build this branch and start an extension debugging session. Open a workspace folder. 
2. Start a debugging session based on known good `launch.json` 
image

3. Observe the prompt for the DFDL Schema
image

4. Press escape.
5. Observe that the debugging session gracefully ends and the following error messages are shown
image

6. Repeat 1-3 above. Provide a valid DFDL Schema this time.
7. Observe the prompt for the data file 
image

8. Press Escape
9. . Observe that the debugging session gracefully ends and the following error messages are shown
image

10. Test the same behavior when executing a TDML file from the `launch.json` or via the "Execute TDML" button present went a `.tdml` file is selected.
11.  Observe that the debugging session gracefully ends with the appropriate error messages.

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -341,22 +341,36 @@ export function activateDaffodilDebug(
       'extension.dfdl-debug.getSchemaName',
       async (fileRequested = null) => {
         // Open native file explorer to allow user to select data file from anywhere on their machine
-        return await getFile(
+        const retVal = await getFile(
           fileRequested,
           'Select DFDL schema to debug',
           'Select DFDL schema to debug'
         )
+
+        if (!retVal)
+          vscode.window.showInformationMessage(
+            'Invalid DFDL schema path selected'
+          )
+
+        return retVal
       }
     ),
     vscode.commands.registerCommand(
       'extension.dfdl-debug.getDataName',
       async (fileRequested = null) => {
         // Open native file explorer to allow user to select data file from anywhere on their machine
-        return await getFile(
+        const retVal = await getFile(
           fileRequested,
           'Select input data file to debug',
           'Select input data file to debug'
         )
+
+        if (!retVal)
+          vscode.window.showInformationMessage(
+            'Invalid data file path selected'
+          )
+
+        return retVal
       }
     ),
     vscode.commands.registerCommand('extension.dfdl-debug.showLogs', () => {
--- a/src/daffodilDebugger/debugger.ts
+++ b/src/daffodilDebugger/debugger.ts
@@ -145,16 +145,30 @@ export async function getDebugger(
 
     // Get schema file before debugger starts to avoid timeout
     if (config.schema.path.includes('${command:AskForSchemaName}')) {
-      config.schema.path = await vscode.commands.executeCommand(
+      const schemaPath = await vscode.commands.executeCommand(
         'extension.dfdl-debug.getSchemaName'
       )
+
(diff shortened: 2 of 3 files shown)


JeremyYao:
I'm running into something odd on Windows 10. When I run the extension via a debugging session and I close out of the prompt for a DFDL file, it crashes the Powershell window that was open in the VSCode window. 

image

Is this something happening to others or just an oddidty on my VSCode instance?

src/adapter/activateDaffodilDebug.ts
```diff
@@ -341,22 +341,32 @@ export function activateDaffodilDebug(
...
         // Open native file explorer to allow user to select data file from anywhere on their machine
-        return await getFile(
+        const retVal = await getFile(
           fileRequested,
           'Select DFDL schema to debug',
           'Select DFDL schema to debug'
         )
+
+        if (!retVal)
+          vscode.window.showErrorMessage('Invalid DFDL schema Path selected')
```
michael-hoke, line 351: Use showInformationMessage instead of showErrorMessage - this matches the behavior in the same error path for TDML.

src/daffodilDebugger/debugger.ts
```diff
@@ -145,16 +145,32 @@ export async function getDebugger(
...
     // Get schema file before debugger starts to avoid timeout
     if (config.schema.path.includes('${command:AskForSchemaName}')) {
-      config.schema.path = await vscode.commands.executeCommand(
+      const schemaPath = await vscode.commands.executeCommand(
         'extension.dfdl-debug.getSchemaName'
       )
+
+      // Stop execution if schemaPath is null, undefined, or empty string
+      if (!schemaPath) {
+        vscode.window.showErrorMessage('Schema path not provided.')
```
michael-hoke, line 154: I think this should be removed - right now, if I press Esc on an input, I get two error messages, but I should only ever see one.

michael-hoke (changes requested):

rthomas320 (dismissed):
Tested on Windows 10. Successfully ran yarn test and yarn package. Ran the extension debugger, when canceling out of the file selection dialog the extension ends gracefully without causing an application error.
Will provide my +1 when the commit is squashed to a single commit and the activateDaffodilDebug.ts prettier formatting issues are resolved.

rthomas320 (approved):
Retested on Wndows 10. Worked as expected.
+1

Issue #1374, opened 2025-08-22
Create Github Template


michael-hoke:
Create a PR template that organizes the information better and provides a checkbox that must be checked for updating the wiki.

Do we want an issue template as well?

michael-hoke:
PR Template starting point (this goes in .github/pull_request_template.md)
``` 
Closes #(issue)

## Description

Please include a summary of the change and which issue is fixed. Also include relevant context or motivation.

## Wiki

- [ ] I have determined that no wiki updates are needed for these changes
- [ ] I have added documentation to the wiki for these changes

## Screenshots (if applicable)

Add screenshots or GIFs to help explain the change visually.
```

PR Workflow (for disallowing a merge if the checkbox isn't checked)
``` 
name: Enforce Wiki Checkbox

on:
  pull_request:
    types: [opened, edited, synchronize, reopened]

jobs:
  enforce-checkbox:
    runs-on: ubuntu-latest
    steps:
      - name: Check required confirmation checkbox
        uses: actions/github-script@v7
        with:
          script: |
            const prBody = context.payload.pull_request.body || "";
            const checkbox1Text = "- [x] I have determined that no wiki updates are needed for these changes";
            const checkbox2Text = "- [x] I have added documentation to the wiki for these changes";

            if (!prBody.includes(checkbox1Text) && !prBody.includes(checkbox2Text) {
              core.setFailed("❌ Required wiki checkbox not checked. Please check the box before merging.");
            } else {
              core.info("✅ Required wiki checkbox is checked.");
            }
```

Issue template (goes in a named file, maybe bug_report.md, under .github/ISSUE_TEMPLATE/ ). We could have multiple, depending on the issue type. We can also force the use of a template in a config.yml file, probably under .github, with the blank_issues_enabled directive set to false.
``` 
## Description

A clear and concise description of what the bug is.

## Steps to Reproduce

1. Go to '...'
2. Click on '...'
3. See error

## Expected behavior

What did you expect to happen?

## Actual behavior

What actually happened?

## Screenshots (optional)

## Environment

- OS: 
- Browser/VS Code version: 
- Extension version: 
```

naga-panchumarty:
[...]

PR #1375, opened 2025-08-22
Fix Visibility/Enablement for Extension Commands and Unify Conditions

### LINKED ISSUE #1344: Move Copy TDML File command out of being DFDL file specific
Currently, a DFDL file has to be active in order for the Copy TDML File command to appear. This doesn't really make sense as we shouldn't need a DFDL file to be active to copy down the generated TDML file.

Two things should happen as a result of this change:
- Copy TDML File should not show up in the command dropdown for DFDL files

Image

- The Copy TDML File command should be runnable from the command palette without a specific type of active file.

### LINKED ISSUE #1358: Commands in the Daffodil Explorer window can be run without the appropriate file in the active window
If I have a non-dfdl file as the active file in the window and hit 'Run File' in the Daffodil Explorer pane, it will treat the active window as the dfdl schema and will ask for the data file. We should either figure out whether the active file is of the appropriate type or remove commands from this pane when incorrect file types are active.

Note that all commands should be verified, including TDML files.

michael-hoke:
Closes #1358 
Closes #1344

This PR is also meant to be a trial run for a PR template - please provide any feedback in #1374 

## Description

- Dynamically evaluate a command's enablement conditions for displaying it in the Daffodil Explorer Pane
- Unify all commands from all sources to have (mostly) the same enablement/when conditions
- Fix keywords used in enablement/when conditions to be common and succinct

## Wiki
- [ ] I have determined that no wiki updates are needed for these changes

Documented a slightly modified version of the screenshot in the wiki: https://github.com/apache/daffodil-vscode/wiki/Apache-Daffodil%E2%84%A2-Extension-for-Visual-Studio-Code:-v1.4.2-%E2%80%90-In-Development/_compare/a7845a450e81b2b07ea028cc96eda737e21d375e...d93088265b1834cbbc4be29a40e21c5194b99415

## Screenshots (if applicable)

Here is a spreadsheet containing all 18 of our extension's commands (not including the 2 commandsView commands that were separately getting filtered out). Conditions for when they should appear are showing at the top, except for the one bolded note. I have unified all commands on all locations (command palette, editor/title, editor/title/run, commands view) to be the same (except for the one note). Please review these as well and make sure that these all make sense.

image

The keywords are variables/macros that appear in the context of the contributes/commands section in the package.json. The commands view keys off of a command's enablement field, and we have to supply the value to the parser so that it can evaluate the statements. These keys are internal values that VSCode is able to fill in when looking at something like the command palette, but it does not expose a way for us to read these values in our code.

### DIFF
--- a/build/package/LICENSE
+++ b/build/package/LICENSE
@@ -705,6 +705,30 @@ conditions of the following licenses.
 
     THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+- 'jsep' in extension/dist/ext/extension.js
+  This product bundles the 'jsep' from the above files.
+  These files are available under the MIT License
+    Copyright (c) 2013 Stephen Oney, https://ericsmekens.github.io/jsep/
+
+    Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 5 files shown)


michael-hoke:
Should we be able to run/debug a dfdl file from anywhere? We have code in to support `${askForSchemaName}`, but we have no way to get to the same code from non-launch config places. Because we have this capability, we could allow the user to run/debug a schema from anywhere. If editorLangId is dfdl, use the active resource (like we're doing now). If it isn't, prompt for it.

If this seems like a good idea, I'll make a new issue.

scholarsmate (approved):
+1. Wow, this is pretty neat!

shanedell (approved):
+1

PR #1378, opened 2025-08-26
Set tdmlConfig.action to 'none' when testing daffodil

michael-hoke:
The TDML file that was being created during the tests is no longer being created. The 'Invalid TDML Path Selected' notification should be showing up. We are running a test where we call getValidatedTDMLPath with a non-existent path. The correct result of this is the notification.


shanedell (approved):
+1

PR #1381, opened 2025-08-26
Added DFDL cross reference excel spreadsheet

rthomas320:
Added markdown files created from dfdl cross reference spreadsheets Added html files created from dfdl cross reference spreadsheets


rthomas320:
In github the excel spreadsheet shows as an empty file, however the actual excel file can be downloaded from the screen with the empty file view.
image

I used the excel saveas function to create an html view of the spreadsheet. When viewing the html from the repo on a local machine, it looks like this
image
But when viewing in github it is html source code.
image

I also used an online utility to convert each of the spreadsheets to markdown.  When viewed locally in a markdown capable viewer (not a web browser) the data looks fine, In a web browser the data is jumbled because there isn't a way to turn off line wrap. When view in gitlab the files a set viewing area that requires scrolling vertically and horizontally to see all the data. The column headers scroll off the screen when scrolling vertically, making the data hard to decipher.
image
image
Please provide your opinion about which versions of the spreadsheet we should leave in the repo for community feedback.  At some point we would like community feed back as to the accuracy of the spreadsheet.  Keep in mind that the actual excel file is stored in binary format, so any updates would be stored in github to the excel file would be stores as an entirely new file.

JeremyYao (approved):
I'd say we should include all 3 excel sheets for community review so that we get full-picture feedback from the community. 

When we're getting feedback from the community, it would be nice to include context as well as how each item in the tables were determined. The context should also be included as a markdown file too in the dfdl-xref folder to give others context.

scholarsmate (approved):
+1

PR #1384, opened 2025-08-28
Fixed TDML Saved File w/o TDML Extension (Non-Windows OSes)

### LINKED ISSUE #1383: TDML Extension Not Appending to Saved TDML File
While testing the extension w/ the PNG DFDL Schema repo I noticed that upon debug session completion and the _Copy TDML File_ button is selected the file chooser dialog clearly states that it will save a TDML file but does not append the `.tdml` if not provided for the filename.

Image

The saved TDML file above gets saved as _beach_ with XML content

Image

---

This occurred on an Ubuntu 24.04 system.

stricklandrbls:
Closes #1383 

## Description

Added shorthand conditional return when saving the generated TDML file in case the `.tdml` extension is not automatically appended when selected within the dialog.

## Wiki

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -103,6 +103,26 @@ function normalizePath(path: string): string {
   return path
 }
 
+/**
+ * Configures the file path string of an intended TDML save file if the
+ * extension is malformed.
+ * @param pathStr
+ * @returns file path string with a valid TDML file extension.
+ */
+function validateTDMLFilePath(pathStr: string): string {
+  // Create capture groups for the path
+  // Capture Group 2 will be all valid extensions potentially chained at the end of the path
+  // Capture Group 1 will be the rest of the filename (before the valid extensions)
+  const matches = pathStr.match(/(^.*?)((\.tdml|\.tdml\.xml)*)$/i)
+  if (matches) {
+    // We want to grab the first valid extension found from the previous regex's Capture Group 2
+    //    and append that to the filename with TDML extensions stripped (previous regex's Capture Group 1)
+    const extMatches = matches[2].match(/^\.tdml.xml|^\.tdml/i)
+    return extMatches ? matches[1] + extMatches : matches[1] + '.tdml'
+  }
+  return pathStr + '.tdml'
+}
+
 /** Method to show dialog to save TDML file
  * Details:
  *   Required so that the vscode api commands:
@@ -110,25 +130,19 @@ function normalizePath(path: string): string {
  *   can be sent a file instead of always opening up a prompt.
  */
 async function showTDMLSaveDialog(fileRequested, label, title) {
-  let file = ''
-
-  file = await vscode.window
-    .showSaveDialog({
-      saveLabel: label,
-      title: title,
-      filters: {
-        TDML: ['tdml'],
-      },
-    })
-    .then((fileUri) => {
-      if (fileUri) {
-        return normalizePath(fileUri.fsPath)
-      }
-
-      return ''
-    })
-
-  return file
+  let file = await vscode.window.showSaveDialog({
+    saveLabel: label,
+    title: title,
+    filters: {
+      TDML: ['tdml', 'tdml.xml'],
+    },
+    defaultUri: fileRequested,
+  })
+  if (!file) {
+    vscode.window.showErrorMessage('No output TDML filename provided')
+    return
+  }
+  return validateTDMLFilePath(normalizePath(file.fsPath))
 }
 
 // Function for setting up the commands for Run and Debug file


michael-hoke:
Please change `.includes` to `.endsWith`. The filename is user-defined and potentially could include .tdml as a part of the filename.

michael-hoke (changes requested):
Please change .includes to .endsWith. The filename is user-defined and potentially could include .tdml as a part of the filename.

michael-hoke:
> > Please change .includes to .endsWith. The filename is user-defined and potentially could include .tdml as a part of the filename.
> 
> Resolved.

Just checked the apache page for TDML and it lists .tdml.xml as a valid extension as well (even though we're not using that here)

Would the following work?

  return (file.endsWith('.tdml') || file.endsWith('.tdml.xml')) ? file : file + '.tdml'

michael-hoke:
A valid filename here could be test.tdml.gif.final.tdml - I really think we need to use endsWith instead of indexOf/split/includes.

We're using a saveFileDialog in this specific case, and I'm not sure what kind of convenience method we could use for a save - typically, those return the last chunk of text after the last dot (eg. they'll return .gz for a .tar.gz file), and we wouldn't want to force append .tdml to a .tdml.xml extension.

stricklandrbls:
@michael-hoke - Pushed up a commit that uses regex to match and if the match is invalid it appends a default `.tdml` extension.

src/adapter/activateDaffodilDebug.ts
```diff
@@ -103,6 +103,18 @@ function normalizePath(path: string): string {
...
 }
 
+/**
+ * Configures the file path string of an intended TDML save file if the
+ * extension is malformed.
+ * @param pathStr
+ * @returns file path string with a valid TDML file extension.
+ */
+function validateTDMLFilePath(pathStr: string): string {
+  return !pathStr.match(/.*tdml[.]xml|.*tdml$/gm)
```
scholarsmate, line 113: Probably ought to make sure the file name ends with `.tdml` not just `tdml` (e.g., foo.tdml vs footdml).

JeremyYao:
+1 on W10. Tested with generating a TDML file where I only entered `test` in the file dialog and it appended `.tdml` to the file name. Tried creating `t.tdml` and an extra `.tdml` was not appended

scholarsmate (approved):
+1

scholarsmate:
@stricklandrbls, this one is ready for merging.

stricklandrbls:
May consider closing this and the related issue as it seems to be a very specific case for both operating system and file browser.

rthomas320:
After creating a new launch.json file I am nolonger seeing the error below on Ubuntu 24.04.

I re-cloned the repo on Ubuntu 24.04, then pulled in the code for PR# 1384. I am to compile and run the code. I am still getting an error when I try to create a TDML file. The TDML file does not get written to the /tmp directory (as it does on Ubuntu 22.04) after running the debugger.
490169226-0b3efdf1-391f-4f96-b700-b04458f1e4cb

JeremyYao:
Ubuntu 24.04-- appended the `.tdml` extension
RHEL9 -- I get the following error when using the following config

image

```JSON
{
    "version": "0.2.0",
    "configurations": [
        {
            "request": "launch",
            "type": "dfdl",
            "name": "Wizard Config",
            "schema": {
                "path": "${command:AskForSchemaName}",
                "rootName": null,
                "rootNamespace": null
            },
            "data": "${command:AskForDataName}",
            "debugServer": 4711,
            "infosetFormat": "xml",
            "infosetOutput": {
                "type": "file",
                "path": "${workspaceFolder}/target/infoset.xml"
            },
            "tdmlConfig": {
                "action": "generate",
                "name": "undefined",
                "description": "undefined",
                "path": "undefined"
            },
            "trace": true,
            "stopOnEntry": true,
            "useExistingServer": false,
            "openDataEditor": false,
            "openInfosetView": false,
            "openInfosetDiffView": false,
            "daffodilDebugClasspath": [],
            "dataEditor": {
                "port": 6969,
                "logging": {
                    "level": "info"
                }
            },
            "dfdlDebugger": {
                "logging": {
                    "file": "${workspaceFolder}/daffodil-debugger.log",
                    "level": "INFO"
                }
            }
        }
    ]
}
```

However, using this launch.json config worked

```JSON
{
    "version": "0.2.0",
    "configurations": [
        {
            "request": "launch",
            "type": "dfdl",
            "name": "Wizard Config",
            "schema": {
                "path": "${command:AskForSchemaName}",
                "rootName": null,
                "rootNamespace": null
            },
            "data": "${command:AskForDataName}",
            "debugServer": 4711,
            "infosetFormat": "xml",
            "infosetOutput": {
                "type": "file",
                "path": "${workspaceFolder}/target/infoset.xml"
            },
            "tdmlConfig": {
                "action": "generate",
                "path": "/tmp/generatedTDML.tdml",
                "name": "Default Test Case"
            },
            "trace": true,
            "stopOnEntry": true,
            "useExistingServer": false,
            "openDataEditor": false,
            "openInfosetView": false,
            "openInfosetDiffView": false,
            "daffodilDebugClasspath": [],
            "dataEditor": {
                "port": 6969,
                "logging": {
                    "level": "info"
                }
            },
            "dfdlDebugger": {
                "logging": {
                    "file": "${workspaceFolder}/daffodil-debugger.log",
                    "level": "INFO"
                }
            }
        }
    ]
}
```

and naming the file `test_robert_extension` in the file dialog resulted in in `test_robert_extension.tdml` file being made. 

Also putting `test.tdml` in the file dialog resulted in `test.tdml` being made. No extra appended `.tdml` was observed when `.tdml` extension is specified in file name.

stricklandrbls:
@rthomas320 - I think there was an issue with how the old version of the `Create TDML File` was working:
```ts
  file = await vscode.window
    .showSaveDialog({
      saveLabel: label,
      title: title,
      filters: {
        TDML: ['tdml'],
      },
    })
    .then((fileUri) => {
      if (fileUri) {
        return normalizePath(fileUri.fsPath)
      }

      return ''
    })
```
Previously, this block was already inside an `async` function then calling `await` on another asynchronous function while also providing a `.then` clause _and_ a return value. I updated this to `await` then moved the clause out.

PR #1386, opened 2025-08-28
Normalize path for schema in launch config

### LINKED ISSUE #1379: Debugging on Windows skips breakpoints
When on Windows (win10 for sure), debugging a schema, whether using TDML Execute or `Daffodil Debug: Debug File`, the breakpoints are skipped when launching the debug session from the command palette or run/editor/title menu. The breakpoints are NOT skipped when launching a debug session with the same parameters from a launch config.

michael-hoke:
Closes #1379 

The path used by the debugger to determine where the source file lives is case-sensitive, even in case-insensitive Windows. Normalizing the drive letter in the path allows the debugger to find the source file for debugging.

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -77,17 +77,7 @@ async function getFile(fileRequested, label, title) {
         if (fileUri && fileUri[0]) {
           let path = fileUri[0].fsPath
 
-          if (
-            process.platform === 'win32' &&
-            path.length > 2 &&
-            path.charCodeAt(0) > 97 &&
-            path.charCodeAt(0) <= 122 &&
-            path.charAt(1) === ':'
-          ) {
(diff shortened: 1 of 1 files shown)


hdalsania (approved):
+1 It is working as expected on Windows platform.

scholarsmate (approved):
+1, but needs to pass CI. I've re-run the failed CI tests.

stricklandrbls (changes requested):
https://github.com/apache/daffodil-vscode/blob/ab70499df3471d7319e6bfb21c00dcf45aa10d4f/src/adapter/activateDaffodilDebug.ts#L127-L137

Should be replaced with the `normalizePath` addition given that it's the exact syntax.

CoverRyan (approved):
+1, Tested on Windows 10 and Ubuntu 24.04

stricklandrbls (approved):
+1

PR #1388, opened 2025-08-29
Make tdmlConfig.path always use default value for generate

### LINKED ISSUE #1376: Remove path from tdmlConfig during a generate
Currently, TDML Generate will key off of tdmlConfig.path to determine if the TDML file should be generated outside of the temp directory. This is currently a problem for Create/Append because they are hard-coded to use the temp directory. Path should not be removed from the launch config because it's needed for Execute, but it should be modified to have no effect during a Generate.

If we end up wanting to make this configurable, we will probably have to use a global setting (VSCode settings menu) to set the path.

michael-hoke:
Closes #1376 

## Description

@shanedell If you don't believe this is something we should merge in, hold off with merging until we can figure out at least an intermediate solution that we all agree on.

Hardcode path for TDML generate to always go into the /tmp directory. We should properly address this at some point post 1.4.2 release so that it will be configurable for Generate, Create, and Append.

Also puts a message in the Daffodil Output Channel containing the path to the generated file - could change that to a notification if it isn't loud enough.

## Wiki

- [ ] I have determined that no documentation updates are needed for these changes

Documented in wiki - diff here: https://github.com/apache/daffodil-vscode/wiki/Apache-Daffodil%E2%84%A2-Extension-for-Visual-Studio-Code:-v1.4.2-%E2%80%90-In-Development/_compare/8c5f447149f53adfe3b5eaeed90ea7a8b661269c...a7845a450e81b2b07ea028cc96eda737e21d375e

### DIFF
--- a/src/daffodilDebugger/debugger.ts
+++ b/src/daffodilDebugger/debugger.ts
@@ -24,6 +24,7 @@ import {
   getDefaultTDMLTestCaseName,
   getTmpTDMLFilePath,
 } from '../tdmlEditor/utilities/tdmlXmlUtils'
+import { outputChannel } from '../adapter/activateDaffodilDebug'
 
 // Function to get data file given a folder
 export async function getDataFileFromFolder(dataFolder: string) {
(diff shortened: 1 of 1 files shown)


shanedell:
@michael-hoke I don't like the way of doing it mostly in this PR currently as it is different from what you mentioned in that issue. This change really does nothing at the moment because its always defaulting to that temp file anyway since the launch configs aren't being loaded. If you update this PR so that:

- The launch `tdmlConfig.path` does not work for `generate`
  - This ensures the generated file is always in the same location.
- After the file is generated there is a notification giving the path to the generated file

michael-hoke:
@shanedell We could definitely add a notification, but we're already putting up 2 notifications during a normal run. Is adding a 3rd something that we want to do?

PR #1402, opened 2025-09-02
Stop Terminal Crashes

### LINKED ISSUE #1387: Terminal Crashes when Exiting File Prompts
Observed on Win 11 and rocky9

### Windows
1. Start an extension debugging session.
2. Open up a terminal in vs code.
3. Run the extension, when prompted for a schema press ESC
4. Observed that the terminal closes/crashes and has an exit code of one.

Image

### Rocky9
1. Repeat the steps above.
2. Observe that a terminal is open when the extension is ran and waiting for the user to select a schema
Image
3. Press ESC and observe that the terminal closes/crashes. Linux doesn't show the error message in the lower right like in windows for reasons unknown 

Image

This ticket should stop the terminal from crashing and normalize the error message if possible. You probably have to test the error message normalization first because once the fix is implemented, there shouldn't be an error generated.

duboisfordwork:
## Testing
Repeat the steps defined in #1387 . Observe that the terminals do not crash.

Closes #1387

### DIFF
--- a/src/daffodilDebugger/utils.ts
+++ b/src/daffodilDebugger/utils.ts
@@ -26,7 +26,7 @@ import { deactivate } from '../adapter/extension'
 import { getDaffodilVersion } from './daffodil'
 import { Artifact } from '../classes/artifact'
 import { DFDLDebugger } from '../classes/dfdlDebugger'
-import { osCheck, runScript } from '../utils'
+import { osCheck, runScript, terminalName } from '../utils'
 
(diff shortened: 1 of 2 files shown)


CoverRyan (approved):
+1, tested on Windows 10 and Ubuntu 24.04

scholarsmate (approved):
+1

stricklandrbls (approved):
+1

PR #1403, opened 2025-09-03
Switch prompts for path from TDML generate to TDML execute in launch …

### LINKED ISSUE #1347: TDML Path not appearing at correct times in Launch Wizard
When copying a launch config (not when creating a new one), the tdmlConfig's path member shows up when tdmlAction is set to 'generate', but not when it is set to 'execute'. This is backwards. Be sure to test all combinations for copy and create to make sure that nothing is missed.

Consider setting the default value to an empty string. The current default may be generatedTDML.tdml. I think we should expect the user to execute a test case from a tdml file that they copied to their workspace and not the generated file.

JeremyYao:
…config.

Add Default Test Case name when adding config with the blue Add Configuration... button

Closes #1347

### Description

Refer to #1347

### Wiki

- [ ] I have determined that no wiki updates are needed for these changes
   - yes, updates are needed
   - See https://github.com/apache/daffodil-vscode/pull/1403#issuecomment-3292625611 for more details

#### Example Files

[test_images.zip](https://github.com/user-attachments/files/22125186/test_images.zip)

#### Validating generate still works

0. Delete launch.json if it exists
1. Open the launch config wizard. 
2. Ensure no path field shows for generate

image

3. Save the config. You should have something that looks like 

image

4. Run `Wizard Config` and select a sample DFDL file and data file

6. Navigate to your OS's temp directory and verify generatedTDML.tdml was created and/or updated after the DFDL debugging session is over 

image

7. Create the TDML file 

image

I named the TDML file `1347.tdml`

#### Validate Execute

1. Go back to `Wizard Config`
2. Select TDML action to be execute
3. Type in the path to the TDML file previously created

image

4. Run the config

#### Execute with blank string in path

1. Create new config

image

2. Set execute to empty string path 

image

The following config item should be added 
```JSON
{
            "request": "launch",
            "type": "dfdl",
            "name": "Wizard Config 2",
            "schema": {
                "path": "${command:AskForSchemaName}",
                "rootName": null,
                "rootNamespace": null
            },
            "data": "${command:AskForDataName}",
            "debugServer": 4711,
            "infosetFormat": "xml",
            "infosetOutput": {
                "type": "file",
                "path": "${workspaceFolder}/target/infoset.xml"
            },
            "tdmlConfig": {
                "action": "execute",
                "path": "",
                "name": "Default Test Case"
            },
            "trace": true,
            "stopOnEntry": true,
            "useExistingServer": false,
            "openDataEditor": false,
            "openInfosetView": false,
            "openInfosetDiffView": false,
            "daffodilDebugClasspath": [],
            "dataEditor": {
                "port": 9000,
                "logging": {
                    "level": "info"
                }
            },
            "dfdlDebugger": {
                "logging": {
                    "file": "${workspaceFolder}/daffodil-debugger.log",
                    "level": "INFO"
                }
            }
        }
```

3. Run this config
4. Validate prompt for .tdml file pops up
5. Select a .tdml file that has a test case name of `Default Test Case` (example `1347.tdml` created in prior step)
6. Verify debugging session runs smoothly

#### Execute with ${AskForValidatedTDMLPath}

Follow steps above, but use this launch config as mentioned in https://github.com/apache/daffodil-vscode/pull/1403#issuecomment-3271741062 instead

### DIFF
--- a/src/daffodilDebugger/debugger.ts
+++ b/src/daffodilDebugger/debugger.ts
@@ -58,13 +58,18 @@ async function getTDMLConfig(
     config.data = ''
     config.schema.path = ''
 
-    config.tdmlConfig.path =
-      config.tdmlConfig.path ||
-      (await vscode.commands.executeCommand(
+    if (
+      config.tdmlConfig.path === '${command:AskForValidatedTDMLPath}' ||
+      !config.tdmlConfig.path
+    ) {
+      config.tdmlConfig.path = await vscode.commands.executeCommand(
         'extension.dfdl-debug.getValidatedTDMLPath'
-      ))
+      )
+    }
 
-    if (!config.tdmlConfig.path) return false
+    if (!config.tdmlConfig.path) {
+      return false
+    }
 
     config.tdmlConfig.name =
       config.tdmlConfig.name ||
--- a/package.json
+++ b/package.json
@@ -111,7 +111,6 @@
     "onCommand:extension.dfdl-debug.getSchemaName",
     "onCommand:extension.dfdl-debug.getDataName",
     "onCommand:extension.dfdl-debug.getTDMLName",
-    "onCommand:extension.dfdl-debug.getTDMLPath",
     "onCommand:extension.dfdl-debug.getValidatedTDMLPath",
     "onCommand:extension.dfdl-debug.getValidatedTDMLCopyPath"
   ],
@@ -576,7 +575,7 @@
                   },
                   "path": {
                     "type": "string",
-                    "default": "${command:AskForTDMLPath}",
+                    "default": "${command:AskForValidatedTDMLPath}",
                     "description": "TDML Case Path"
                   }
                 },
@@ -788,7 +787,8 @@
                 "path": "^\"\\${workspaceFolder}/target/infoset.xml\""
               },
               "tdmlConfig": {
-                "action": "generate"
+                "action": "generate",
+                "name": "Default Test Case"
               },
               "debugServer": 4711,
               "openDataEditor": false,
@@ -816,7 +816,6 @@
         "variables": {
           "AskForSchemaName": "extension.dfdl-debug.getSchemaName",
           "AskForDataName": "extension.dfdl-debug.getDataName",
-          "AskForTDMLPath": "extension.dfdl-debug.getTDMLPath",
           "AskForValidatedTDMLPath": "extension.dfdl-debug.getValidatedTDMLPath",
           "AskForValidatedTDMLCopyPath": "extension.dfdl-debug.getValidatedTDMLCopyPath"
         }
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -440,22 +440,6 @@ export function activateDaffodilDebug(
     )
   )
 
-  context.subscriptions.push(
-    vscode.commands.registerCommand(
-      'extension.dfdl-debug.getTDMLPath',
-      async (_) => {
-        return await vscode.window
-          .showInputBox({
-            prompt: 'TDML File: ',
-            value: '${workspaceFolder}/infoset.tdml',
-          })
-          .then((value) => {
-            return value
-          })
-      }
-    )
-  )
-
   // register a configuration provider for 'dfdl' debug type
   const provider = new DaffodilConfigurationProvider(context)
   context.subscriptions.push(
--- a/src/launchWizard/launchWizard.ts
+++ b/src/launchWizard/launchWizard.ts
@@ -495,8 +495,16 @@ class LaunchWizard {
       }
     })
 
-    let tdmlActionSelect = 'none'
-    let tdmlActions = ['none', 'generate', 'execute']
+    const TDML_EXECUTE_ACTION = 'execute'
+    const TDML_NONE_ACTION = 'none'
+    const TDML_GENERATE_ACTION = 'generate'
+
+    let tdmlActionSelect = TDML_NONE_ACTION
+    let tdmlActions = [
+      TDML_NONE_ACTION,
+      TDML_GENERATE_ACTION,
+      TDML_EXECUTE_ACTION,
+    ]
     let tdmlAction =
       'tdmlConfig' in defaultValues ? defaultValues.tdmlConfig['action'] : null
     let tdmlName =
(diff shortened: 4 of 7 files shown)


JeremyYao:
Please @ me when this gets merged in so I can update the wiki afterwards

src/daffodilDebugger/debugger.ts
```diff
@@ -78,7 +78,7 @@ async function getTDMLConfig(
   if (config?.tdmlConfig?.action === 'generate') {
     config.tdmlConfig.name =
       config.tdmlConfig.name || getDefaultTDMLTestCaseName()
-    config.tdmlConfig.path = config.tdmlConfig.path || getTmpTDMLFilePath()
+    config.tdmlConfig.path = getTmpTDMLFilePath()
```
CoverRyan, line 81: This change is the same as the change made in [1388](https://github.com/apache/daffodil-vscode/pull/1388). The issue this resolves, [1347](https://github.com/apache/daffodil-vscode/issues/1347), seems to overlap with [1376](https://github.com/apache/daffodil-vscode/issues/1376) and the changes in 1388 in general.

JeremyYao: Added `Closes #1376` to the commit message

michael-hoke:
If you need to make wiki changes, do them and have them reviewed as a part of this PR. The checkbox is also meant to indicate to reviewers that there are documentation changes needed that they need to review.

JeremyYao:
> Why does this PR touch the default value for tdmlConfig.path?

If the user can't specify a custom pathing, then it should be reflected in how the pathing for the generated TDML is set.

michael-hoke:
> > Why does this PR touch the default value for tdmlConfig.path?
> 
> If the user can't specify a custom pathing, then it should be reflected in how the pathing for the generated TDML is set.

I'm not sure I follow. Are you saying that since we're not allowing the users to change the path, we should default to an empty string and then hardcode it right before we start the debugging session?

JeremyYao:
> > > Why does this PR touch the default value for tdmlConfig.path?
> > 
> > 
> > If the user can't specify a custom pathing, then it should be reflected in how the pathing for the generated TDML is set.
> 
> I'm not sure I follow. Are you saying that since we're not allowing the users to change the path, we should default to an empty string and then hardcode it right before we start the debugging session?

The default TDML config values are the starting initialization values for when we're creating new launch configurations. So, what this means is that '' is the default value.

michael-hoke:
> > > > Why does this PR touch the default value for tdmlConfig.path?
> > > 
> > > 
> > > If the user can't specify a custom pathing, then it should be reflected in how the pathing for the generated TDML is set.
> > 
> > 
> > I'm not sure I follow. Are you saying that since we're not allowing the users to change the path, we should default to an empty string and then hardcode it right before we start the debugging session?
> 
> The default TDML config values are the starting initialization values for when we're creating new launch configurations. So, what this means is that '' is the default value.

I do not think that '' should be the default value. What if we used `${AskForTDMLPath}` as the default value and added a block in debugger.ts:getDebugger to handle this for the two cases? I think this should also let us separate our PRs again.

src/daffodilDebugger/debugger.ts
```diff
@@ -58,12 +58,18 @@ async function getTDMLConfig(
...
+          )
+        : config.tdmlConfig.path
 
-    if (!config.tdmlConfig.path) return false
+    if (!config.tdmlConfig.path) {
+      vscode.window.showErrorMessage(
+        `Invalid path TDML Execute path ${config.tdmlConfig.path} was specified!`
+      )
+      return false
+    }
```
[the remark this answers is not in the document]
JeremyYao: The check for undefined `tdmlConfig.path` is already defined via the 

```Typescript
    if (!config.tdmlConfig.path) {
      vscode.window.showErrorMessage(
        `Invalid path TDML Execute path ${config.tdmlConfig.path} was specified!`
      )
      return false
    }
``` 

block. The value `!undefined` evaluates to `true`. Also `!''`` returns true as well.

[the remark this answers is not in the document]
JeremyYao: > If the user sends in an undefined or empty tdmlConfig.path, they will no longer be prompted for a path. I don't think that behavior should be regressed. If we see an undefined/empty path, I think we should prompt the user for the path instead of letting them know they made a mistake with their launch config and forcing them to fix it and restart.

Re-added via https://github.com/apache/daffodil-vscode/pull/1403/commits/bcb1c9bf52cd1e6ced778f38e65310366b7a5978.

src/daffodilDebugger/debugger.ts
```diff
@@ -58,12 +58,16 @@ async function getTDMLConfig(
...
-      config.tdmlConfig.path ||
-      (await vscode.commands.executeCommand(
-        'extension.dfdl-debug.getValidatedTDMLPath'
-      ))
+      config.tdmlConfig.path == '${AskForValidatedTDMLPath}' ||
+      !config.tdmlConfig.path
+        ? await vscode.commands.executeCommand(
+            'extension.dfdl-debug.getValidatedTDMLPath'
+          )
+        : config.tdmlConfig.path
```
michael-hoke, line 66: This is a case where a ternary is harder to read than an if statement. The `config.tdmlConfig.path == '${AskForValidatedTDMLPath}'` is the main culprit here (and should be ===).

``` 
if (!config.tdmlConfig.path || config.tdmlConfig.path === '${AskForValidatedTDMLPath}')
{
  config.tdmlConfig.path = await vscode.commands.executeCommand('extension.dfdl-debug.getValidatedTDMLPath')
}
```

src/daffodilDebugger/debugger.ts
michael-hoke: Did this need to be added back in?

michael-hoke (changes requested):

shanedell:
@JeremyYao Change

```txt
- [x] I have added documentation to the wiki for these changes 
```

to

```txt
```

and I will approved and merge after I rerun the enforce doc CI, which should pass after this change.

JeremyYao:
> @JeremyYao Change
> 
> ```
> - [x] I have added documentation to the wiki for these changes 
> ```
> 
> to
> 
> ```
> - [x] I have added following documentation for these changes
> ```
> 
> and I will approved and merge after I rerun the enforce doc CI, which should pass after this change.

Done

michael-hoke:
Looks ready to merge - created #1418 to track related bug

michael-hoke (approved):

JeremyYao:
> Looks almost ready to merge - created #1418 to track related bug
> 
> Need to change references of ${AskForValidatedTDMLPath} to ${command:AskForValidatedTDMLPath}

Done in https://github.com/apache/daffodil-vscode/commit/fe97ae329938e99d463022daa381348d1bd74477

PR #1409, opened 2025-09-09
remove jar viewer dependency

### LINKED ISSUE #1351: Investigate use of jar-viewer plugin
We're have the wmanth.jar-viewer plugin listed as a dependency. Investigate whether we are actually dependent on it, or whether we can move it to the recommended extension list only.

I thought we were using it automatically when we read more complex schemas that needed to be loaded from jars, but the location of that code isn't immediately obvious to me.

CoverRyan:
Closes #1351 

Made jar viewer plugin optional/recommended

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -50,7 +50,6 @@
     - [General Workflow](#general-workflow)
     - [Opening the Repository in VSCode](#opening-the-repository-in-vscode)
       - [Recommended VSCode Extensions](#recommended-vscode-extensions)
-      - [Required VSCode Extensions](#required-vscode-extensions)
     - [Verifying Setup Can Build](#verifying-setup-can-build)
       - [Yarn Package](#yarn-package)
       - [Automated Testing Suite](#automated-testing-suite)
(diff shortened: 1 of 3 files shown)


DEVELOPMENT.md
```diff
@@ -224,10 +223,6 @@ Upon opening, VSCode may prompt you to install recommended extensions. Go ahead
```
michael-hoke, line 223: It won't let me attach this to the above line, but I don't think the 'Go ahead and accept installing the recommended extensions' applies here anymore - I see that this is the README for devs, but Prettier is a different "type" of recommended extension than the other two. Instead, could we add a very short description alongside each recommended extension for when these would be needed?

eg:
Prettier - Code formatter needed to pass CI formatting checks for PRs
Jar Viewer - DFDL schema files can be packed in jar files when passed to Daffodil. This extension allows the debugger to show these schema files while stepping through them.
Highlight Matching Tag - Convenience extension to better highlight opening and closing tags in DFDL schema files

michael-hoke (changes requested):

rthomas320 (approved):
ested on Windows 10. Successfully ran yarn test and yarn package. Successfully ran extension. Did not get message to install jar viewer dependency.
+1

scholarsmate (approved):
+1

PR #1448, opened 2025-10-01
Make Infoset Not Automatically Open

### LINKED ISSUE #1421: Infoset view opens when debug session is started from command palette but not from launch config
## Description

Starting a debugging session from the command palette causes the infoset to be opened after the debugging session is completed. However, starting a debugging session from the launch config causes the infoset to not be opened.

I don't think it really matters how exactly this ticket is solved - either always open the infoset after a debugging session is finished or don't.

## Steps to Reproduce

[ 
1. Create basic launch config, keeping as close to the default as possible

2. Start a debugging session. Note that when it finishes, the infoset doesn't open
   
3. Run a debugging session via the command palette, using the same schema/data files.
   
4. Note that the infoset automatically opens at the end of the debugging session
 
]

## Expected Behavior

I expect both ways of starting a debugging session to work consistently

## Actual Behavior

They are not consistent

There is an `openInfosetView` field that opens the infoset in a split editor. This is not related to this issue.

In activateDaffodilDebug.ts, in the createDebugRunFileConfigs function, there is a block at the end that looks like this:

``` 
vscode.debug.onDidTerminateDebugSession(async () => {
      if (!vscode.workspace.workspaceFolders) {
        return
      }

      vscode.workspace
        .openTextDocument(
          `${vscode.workspace.workspaceFolders[0].uri.fsPath}/${infosetFile}`
        )
        .then((doc) => {
          vscode.window.showTextDocument(doc)
        })
    })
```

This is probably why the infoset opens only when the debugging session is started from the command palette.

## Environment

[ 
- Operating System(s): Rocky 9 and Windows 10

- Browser/VS Code version: 1.104.0
  
- Extension version: 1.4.2 development

]

CoverRyan:
Closes #1421 

## Description

Made the infoset not automatically open for consistency. This also fixes the issue with the schema file in preview mode closing when the debug session ends.

## Wiki

1. Start a debugging session from the command palette.
2. Continue through the debug session until it ends.
3. Ensure the infoset does not automatically open and the following popup appears at the bottom and the infoset opens when the open button is clicked.
image

4. Go back and repeat steps 1-3 but instead start a debug session using a launch.json file. 

Additionally, test that a file in preview mode is not closed when the debug session ends when started from the command palette. To open a file in preview mode, single click the file in the file explorer. The label for the file tab should be in italics and say in preview mode when hovered over.

### DIFF
--- a/src/adapter/activateDaffodilDebug.ts
+++ b/src/adapter/activateDaffodilDebug.ts
@@ -205,20 +205,6 @@ async function createDebugRunFileConfigs(
 
       vscode.debug.startDebugging(undefined, config, { noDebug: noDebug })
     }
-
-    vscode.debug.onDidTerminateDebugSession(async () => {
-      if (!vscode.workspace.workspaceFolders) {
-        return
-      }
-
-      vscode.workspace
-        .openTextDocument(
-          `${vscode.workspace.workspaceFolders[0].uri.fsPath}/${infosetFile}`
-        )
-        .then((doc) => {
-          vscode.window.showTextDocument(doc)
-        })
-    })
   }
 }
 


JeremyYao:
> +1, tested on W10 and Ubuntu 24.04. Infoset doesn't automatically open

I realized this was all done with preview mode enabled. The setting to enable/disable preview mode is found in VS Code settings

image

JeremyYao:
Did a quick re-test with the preview mode unchecked. No infoset was automatically opened using the two methods on both W10 and Ubuntu 24.04

rthomas320 (approved):
Tested on Windows 10.  Yarn test and yarn package completed successfully. Works as expected.
 +1

PR #1453, opened 2025-10-06
Remove workflow_dispatch from main CI file as it doesn't work as expected.

shanedell:
## Description

Remove workflow_dispatch from main CI file as it doesn't work as expected.

### Justification

The reason for adding this before was to make it so we could manually trigger the CI workflow for a PR. However, even though this will run the CI for the changes, it doesn't update the status of the PR checks. So, even when these are triggered you would have to dig through the CI job to know its running for a PR and not something else. For this reason I think we should just remove it as I don't think its helpful and add a lot non-necessary complexity to the CI file.

If you take a look at https://github.com/apache/daffodil-vscode/actions/runs/18291401974. I manually triggered CI workflow on the branch `update/munit-1.2.0` for PR https://github.com/apache/daffodil-vscode/pull/1445. However, when looking at https://github.com/apache/daffodil-vscode/pull/1445/checks nothing shows up even though the CI succesfully ran on the branch and PR.

I also updated the concurrency group to be a little more detailed and cancel in progress. This was in effort to more closely match what is done at https://github.com/apache/daffodil/blob/main/.github/workflows/main.yml and  https://github.com/apache/daffodil-sbt/blob/main/.github/workflows/main.yml

## Wiki

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,20 +22,15 @@ on:
       - "update/**" # scala-steward branches
   pull_request:
     types: [opened, synchronize, reopened]
-  workflow_dispatch:
-    inputs:
-      pr:
-        description: 'PR number to verify (e.g., 1400)'
-        required: true
-        type: number
 
 permissions:
   contents: read
   pull-requests: read
 
+# Cancel CI runs in progress when a pull request is updated.
 concurrency:
-  group: ci-${{ github.event_name }}-${{ inputs.pr || github.ref }}
-  cancel-in-progress: false
+  group: ci-${{ github.head_ref || ((github.ref_name != 'main' && github.ref_name) || github.run_id) }}-${{ github.workflow }}
+  cancel-in-progress: true
 
 jobs:
   single-commit:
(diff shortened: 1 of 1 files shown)


stevedlawrence:
> @stevedlawrence Does the [labler](https://github.com/apache/daffodil/blob/main/.github/labeler.yml) and [pr-labeler](https://github.com/apache/daffodil/blob/main/.github/workflows/pr-labeler.yml) out in in Daffodil make it so you guys can approve the scala-steward PR's to run workflows? If so that is similar to what we were trying to do with the workflow_dispatch. So I can add that stuff in this PR if that is how it behaves. Or does it strictly only add the `dependencies` label to the PR?

No, all it does is add the `dependencies` label. As far as I know there isn't away to auto run scala-steward PR's, I think the reasoning is because forked code is considered untrusted and peopel can do pretty nasty things--I remember bots used to modify PRs to include bitcoin mining scripts so github disabled automatically running CI on PRs from untrusted code.

We just always manually approve the PRs to run. I don't think switching to the scala-steward instance affects that.

scholarsmate (approved):
+1, we're moving to using the scala-steward instance that both daffodil and daffodil-sbt use.

PR #1601, opened 2026-02-02
Reverting changes clears data highlighting

### LINKED ISSUE #1536: Highlight block size wrong after editor revert
## Description

When you perform a substitution in the editor with a larger block, the highlight block grows to match the size of the insertion, but if you then revert the substitution, the highlight block maintains the new larger size rather than returning to the original size.

## Steps to Reproduce

With a file open in the Data Editor, enter a search string that is present in the file.
Click "Start Search"
Enter a replacement string that is longer than the search string.
Click "Start Replacment"
Click "Replace the Current Match"
Click "Revert All Changes"
Click "Yes" on the pop-up asking if you want to revert all changes.

## Expected Behavior

When the substitution is reverted, the highlight block should also revert to the originally selected block.

## Actual Behavior

The highlighted block remains the size of the inserted block -- even though that change was reverted.

Successful search highlights the found string:
Image
Substitution replace found string and highlight expands to show bytes changed:
Image
"Revert changes" restores original value to bytes, but the highlight still marks all bytes affected by the substitution rather than reverting to the found search string:
Image

## Environment

- Operating System(s): Windows 11
- Browser/VS Code version: 1.105.0
- Extension version: 1.5.0 candidate
- Daffodil version: 4.0.0

stricklandrbls:
Closes #1536

## Description

Submitting an `undo` command through the Data Editor clears any existing search/replace result highlighting.

## Wiki

Perform the steps outlined in #1536

### DIFF
--- a/src/svelte/src/App.svelte
+++ b/src/svelte/src/App.svelte
@@ -64,6 +64,7 @@ limitations under the License.
   } from './components/DataDisplays/CustomByteDisplay/BinaryData'
   import { byte_count_divisible_offset } from './utilities/display'
   import Help from './components/layouts/Help.svelte'
+  import { viewportByteIndicators } from 'utilities/highlights'
 
   function requestEditedData() {
     if ($requestable) {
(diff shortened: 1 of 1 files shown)


shanedell (approved):
+1

PR #1648, opened 2026-03-27
Searching results highlighting appear as expected.

### LINKED ISSUE #1015: Byte highlighting doesn't always work when scrolling through the results
Most of the time the highlighting works, but from time to time, it doesn't.  I suspect it might be an order of events kind of bug.  It also affect the replace feature where often you'll see the bytes that will be replaced by the new sequence of bytes, but from time to time, the highlighting doesn't work.

Here you can see the searched for bytes (xml) at the found offset (272f), but those bytes are not being highlighted.

Screenshot 2024-04-29 at 10 43 24 AM

But if I seek back to the last match, then forward to this match again, the highlights appear.

Screenshot 2024-04-29 at 10 48 44 AM

stricklandrbls:
Closes #1015 

## Description

Reworked the byte highlighting assignments for selections to replace a value in the highlighting array rather than modifying in place.

## Wiki

See https://github.com/apache/daffodil-vscode/issues/1015#issuecomment-2752334098

### DIFF
--- a/src/svelte/src/utilities/highlights.ts
+++ b/src/svelte/src/utilities/highlights.ts
@@ -99,23 +99,23 @@ class ViewportByteIndications extends SimpleWritable<Uint8Array> {
     if (selectionData.active || selectionData.makingSelection()) {
       const offsetPartitions = [
         generateSelectionCategoryParition(0, start, (byte) => {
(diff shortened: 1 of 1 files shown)


CoverRyan (approved):
+1

shanedell (approved):
+1

PR #1661, opened 2026-04-10
Upgrade Omega Edit to v2.0.0-rc1

scholarsmate:

## What changed
- upgraded the Omega Edit integration to `v2.0.0-rc1`
- updated the data editor to the Omega Edit 2.x client APIs, including heartbeat and viewport subscription helpers
- adjusted server metrics and packaging to the new native runtime fields and `@omega-edit/server/out` layout
- refreshed the Omega Edit test startup call sites for the 2.x `startServer(..., { logConfigFile })` signature

## Why
Omega Edit `v2.0.0-rc1` changes the client/server package layout and several APIs. This PR aligns the VS Code extension with the migration guide and the new helper APIs so we can validate the upgrade in CI.

## Impact
- keeps the data editor building and packaging against the RC release
- updates the runtime metrics surfaced in the UI to match the new server fields
- ensures the packaged VSIX includes the server binaries from the new `out/` directory structure

## Validation
- `yarn install`
- `yarn compile`
- `yarn test:svelte`
- `yarn vite:pkg`
- `yarn package`

## Notes
- `node .\\out\\tests\\runTest.js` still fails on this Windows environment because the downloaded `Code.exe` rejects the `@vscode/test-electron` launch flags (`--no-sandbox`, `--extensionTestsPath`, etc.), so the VS Code integration suite could not be exercised locally here.

### DIFF
--- a/build/package/LICENSE
+++ b/build/package/LICENSE
@@ -1103,6 +1103,41 @@ conditions of the following licenses.
     TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
     SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+- '@protobuf-ts/runtime' in extension/dist/ext/extension.js
+- '@protobuf-ts/runtime' in node_modules/@omega-edit/client
+  This product bundles '@protobuf-ts/runtime' from the above files.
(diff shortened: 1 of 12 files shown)


scholarsmate:
OBE due to v2.0.0 being released.

PR #1735, opened 2026-06-26
Add build time launch file schema constraints from allowed tunable properties

### LINKED ISSUE #455: Add build-time launch file schema constraints from allowed tunable properties
#200 adds the ability to specify variables and tunables as maps of string-valued properties. The allowed set of tunable property names is fixed in [an XML configuration file](https://github.com/apache/daffodil/blob/163fa7336148b2e1a6d1f2e44967edaf1fea6616/daffodil-propgen/src/main/resources/org/apache/daffodil/xsd/dafext.xsd#L101) that is packaged in the daffodil-lib artifact.

It should be possible to create a build task that re-writes the package.json file to include schema restrictions on the tunable property names that are derived from that upstream configuration file. Then the built extension could better validate launch.json files that use tunables.

adedoyinNteligen:
Closes #455 

## Description

If a user adds a invalid tunable their launch.json either by wizard or manually dfdl extension does not fail gracefully. 
This PR is meant to prevent users from adding invalid tunables that would crash dfdl. From the wizard, users wont be allowed to save invalid tunables and are alerted if the tunable is invalid. In case a user creates the launch.json manually, when they try to run dfdl, they will get a error box with two options
1. Ignore Invalid tunables. This option will run dfdl and ignore the invalid tunables in the launch.json. It does NOT remove the tunable from the launch.json, just runs the config without the tunable attached

2. Cancel. This stops dfdl from running and the user goes back to correct their config.

The list of valid tunables is in constants/tunables.json . This will be updated according to [DFDL tunables ](https://daffodil.apache.org/tunables/) with each release. A future will be created in the future to give users a way to bypass or add new tunables to the allowed list in case the .json is not updated properly. (The current recommendation is a switch in the launch.json which defaults to true, and the user can set it to false and tunables wont be validated anymore. This way users can remove the guiderails at their own risk ) 
https://github.com/apache/daffodil-vscode/issues/1736
The enhancement ticket is listed above

## Wiki

- [] I have determined that no documentation updates are needed for these changes  
- [x] I have added the following documentation for these changes

Constants/tunables.json is a list of valid tunables. This file will be updated as necessary with each version of DFDL

[Add review instructions including screenshots or GIFs to help explain the change visually.]

### Confirmation Testing

1. Add a invalid tunable to the launch.json manually

2. Open the wizard and scroll to where tunables are listed. You should see a invalid warning. 

3. Try to save. This should fail

4. Edit the tunable to be valid. Save should now work

5. The launch json should be updated to the valid tunable.

6. Add a invalid tunable to launch.json manually

7. Try to run dfdl extension. You should fail with two options

8. Ignore valid tunable should allow you to proceed with running. Cancel should fail

### Regression Testing

[OPTIONAL: Describe any regression testing performed to ensure existing functionality was not broken.  
List relevant test suites, manual checks, or validation steps.]


adedoyinNteligen:
Add screenshots
Create enhancement ticket
Add docs on tunables.json to wiki

lrbarber (changes requested):
Some of the Daffodil tunables seem to be missing from the list used for validating entries -- mostly from the bottom of the list provided at https://daffodil.apache.org/tunables/

Also (minor) when the tunable name is invalid, the error message changes the vertical spacing so that the value field is lower than then name field.

Image

adedoyinNteligen:
The validation of values in the wizard is not very strong, to avoid blocking from someone from submitting something that may be omitted in documentation accidentally. this can always be improved upon, but runs the risk of a valid value being omitted.

lrbarber:
Unfortunately, the new version doesn't allow spacing for the warning message and it gets hidden by the next entry.
image

lrbarber (approved):
Errors are nicely flagged with red text and a note explaining the error!

CoverRyan (approved):
+1

PR #1738, opened 2026-06-26
DFDL-1555: insteading of erroring yet, change to warning

### LINKED ISSUE #1555: Variable not found error (in launch.json) should not be catastrophic
## Description

If a variable definition has been added to the launch.json file, but that variable does not exist in the schema the debugger will not start.

## Steps to Reproduce

1. Add a viarable defintion to the launch.json file:

Image

2. Launch debugger with launch.json file
3. See pop-up error   

## Expected Behavior

This should probably just generate a warning. Definitely does not need to be an error that prevents debugger from running.

## Actual Behavior
Catastrophic error that aborts debugger.
Image

## Environment
- Found on Windows 11 (but probably NOT OS dependent)
- Extension version 1.5.0 RC 1

naga-panchumarty:
Closes #1555 

## Description

when a variable defined in launch.json does not exist in the DFDL schema, the debugger previously threw an unhandled ExternalVariableException and aborted the session entirely. This change catches that exception and changes it to a warning, allowing the debugger to start normally with the unrecognized variable simply skipped.

## Wiki

## Testing steps

1.  Run git clean -fdx
2. Start the extension 
3. Add a viarable defintion to the launch.json file, it can be anything you want, see example below:
image

4. Then run a debug session, you should see a warning banner on the bottom right, and debugger should continue to work as normal
{5336952D-4BF1-4125-B645-A1826D28D5F7}

### Confirmation Testing
Manually tested by adding "variables": { "foo": "bar" } to a launch.json where foo does not exist in the target schema. Confirmed the debugger launches successfully and the warning message appears. Confirmed that valid variables (ones that do exist in the schema) are still applied correctly.

### DIFF
--- a/debugger/src/main/scala-2/org.apache.daffodil.debugger.dap/Support.scala
+++ b/debugger/src/main/scala-2/org.apache.daffodil.debugger.dap/Support.scala
@@ -27,6 +27,8 @@ import java.nio.file.Path
 import org.apache.daffodil.sapi._
 import org.apache.daffodil.sapi.io.InputSourceDataInputStream
 import org.apache.daffodil.sapi.infoset.{JsonInfosetOutputter, XMLTextInfosetOutputter}
+import cats.effect.IO
+import cats.syntax.all._
 
 object Support {
   /* Daffodil DataProcessor wrapper methods */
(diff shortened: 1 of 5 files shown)


scholarsmate (approved):
+1

lrbarber (approved):
Verified that parsing works as expected with an unknown variable setting provided in launch.json and provides a warning pop-up in the lower right corner of the screen.

CoverRyan (approved):
+1

PR #1755, opened 2026-07-21
Add Extension Settings Documentation

### LINKED ISSUE #1526: New Launch configuration not defaulting as expected
## Description

When creating a new launch.json file, it is pulling in some values from VSCode workspace settings rather than inserting the default values we're used to.

## Steps to Reproduce

Use CTRL + ',' , then click "Workspace", then "Extensions", then "Daffodil" 
Change the Data file selection or debugServer setting.
Remove your launch.json file.
Use command palette to create a new launch.json file.
See that the workspace value has been inserted into the launch.json settings.

## Expected Behavior

New launch configurations data file selection defaults to ${command:AskForDataName}

## Actual Behavior

Default value was taken from VSCode workspace settingds.

The biggest problem with this is that if the data file doesn't default to ${command:AskForDataName}, the user is unlikely to know what to enter to return to this functionality.

We should either force launch.json defaults to use specific know values like ${command:AskForDataName}, or add a button to the launch configuration wizard to set the data file selection to it -- so the user can get that functionality without having to know the exact text to enter on their own.
## Environment

[ 
All versions.
OS independent.
]

CoverRyan:
Closes #1526 

## Description

Added documentation for extension settings.

## Wiki

Review the changes made to Wiki.md.


hdalsania (approved):
+1

stricklandrbls (approved):
+1

PR #1773, opened 2026-08-14
Update daffodil support for daffodil version 4.1.0 and 4.2.0

### LINKED ISSUE #1639: Update to Apache 4.1.0 for Scala Debugger 3
Update the Scala 3 debugger code to leverage Apache Daffodil 4.1.0 by default instead of 4.0.0.

hdalsania:
Closes #1639 

## Description

Added support for daffodil 4.1.0 and daffodil version 4.2.0.
Added dropdown to the launch configuration wizard. This version of daffodil vscode extension supports daffodil versions 3.9.0, 3.10.0, 3.11.0, 4.0.0, 4.1.0, 4.2.0

## Wiki

- [] I have determined that no documentation updates are needed for these changes  
- [x] I have added the following documentation for these changes
README.md

When testing this PR you will want to try running multiple versions of the debugger. To do this you will want to update the dfdlDebugger settings to be something like below:

```
dfdlDebugger": {
    "version": "3.11.0",
    "timeout": "10s",
    "logging": {
        "file": "${workspaceFolder}/leftOver/daffodil-debugger.log",
        "level": "INFO"
    }
}
```
or select a version from the launch configuration wizard 

image

When testing use this same dfdlDebugger config but update version to be 4.2.0, 4.1.0, 4.0.0, 3.11.0, 3.10.0, and 3.9.0 at least once for each. This will make it so you would have tested all 3 versions of the debugger (scala 3, scala 2.13 and scala 2.12). When testing a different version you should see a popup in the bottom right of your screen saying the Daffodil JARs are being downloaded/configured.

### Regression Testing
Check core functionality for every version of daffodil this extension supports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,13 @@
 
 This file outlines the release notes and known issues for the "daffodil-vscode" extension.
 
+## 1.6.0
+  ### Debugger:
+  - Updated Scala 3 debugger compatibility for Apache Daffodil 4.x runtimes.
+    - Added support for Daffodil 4.0.0, 4.1.0, and 4.2.0 infoset runtime API shapes.
+    - Added runtime compatibility regression coverage for infoset walker API detection on Scala 3.
+    - Aligned Scala 3 target to 3.3.7 based on Daffodil release notes guidance.
+
 ## 1.5.0
   ### Debugger:
   - Added support for multiple versions of the Daffodil parser (i.e., Daffodil 3.10.0, 3.11.0, and 4.0.0). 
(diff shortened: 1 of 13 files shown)


lrbarber (commented):
Initial test seemed to fail, but I was unknowingly running the wrong configuration "wizard config copy" rather than "wizard config" (the entire name was not visible in the tab.

lrbarber (approved):
Selection works as expected.
+1

scholarsmate (approved):
+1, LGTM.

CoverRyan (approved):
+1

